PR #28, opened 2018-07-30
CronJob Scheduler

Ilanak:


presidio-scanner/cmd/presidio-scanner/scanner-factory.go
```diff
@@ -8,7 +8,7 @@ import (
 
 func createScanner(scanRequest *message_types.ScanRequest) scanner.Scanner {
 	if scanRequest.GetKind() == "s3" || scanRequest.GetKind() == "azure" {
```
torosent: this needs to azureblob
according to 
https://github.com/presid-io/presidio-genproto/blob/master/src/databinder.proto#L33

Issue #52, opened 2018-08-12
Performance: Create template caching mechanism


torosent:
Create an in-memory template caching mechanism to improve performance and reduce I/O to disk and k8s secrets.

PR #85, opened 2018-12-26
Create redis cache for templates

torosent:


pkg/presidio/templates/templates.go
```diff
@@ -28,18 +35,41 @@ func createKey(project string, action string, id string) string {
 // GetTemplate from key store
 func (templates *Templates) GetTemplate(project string, action string, id string) (string, error) {
 	key := createKey(project, action, id)
+	if templates.cacheStore != nil {
+		res, err := templates.cacheStore.Get(key)
```
Ilanak: if the key is not in the cache, but the cache is configured, it should be added to the cache..

torosent:
#52

PR #86, opened 2019-01-03
Fix abort error to return the error message in the response body

Ilanak:


pkg/server/server.go
```diff
@@ -91,3 +91,12 @@ func WriteResponse(
 ) {
 	c.JSON(statusCode, responseBody)
 }
+
+//AbortWithError aborts the request and returns the error in the response body
+func AbortWithError(c *gin.Context,
+	statusCode int,
+	err error) {
+	c.Error(err)
+	c.JSON(http.StatusBadRequest, err.Error())
```
torosent: Hardcoded :)

PR #88, opened 2019-01-07
Json anonymizer

Ilanak:
* add structured json analyzer and anonymizer.


README.MD
```diff
@@ -126,6 +126,17 @@ You can also create reusable templates
     $ http -f POST <api-service-address>/api/v1/projects/<my-project>/anonymize-image detectionType='OCR' analyzeTemplateId='<my-analyze-template-name>' anonymizeImageTemplateId='<my-anonymize-image-template-name>' imageType='image/png' file@~/test-ocr.png > test-output.png
     ```
 
+***Sample 6*** 
+1. Create an anonymize json template, this will define the json schema to be analyzed and anonymized.
+   ```sh
+    echo -n '{"jsonSchema":"{\"Name\": \"<PERSON>\", "Address": "<LOCATION>}"}' | http <api-service-address>/api/v1/templates/<my-project>/anonymize-json/<my-anonymize-json-template-name>
```
eladiw: just a thought. will it be better if the api will be just 'anonymize' / 'analyze' and you can pass 'regular' or json text?
and the method will detect if its a json or not?

pkg/jsonCrawler/json-crawler.go
```diff
@@ -0,0 +1,189 @@
...
+		analyzeItem:       analyze,
+		anonymizeItem:     anonymize,
+		ctx:               ctx,
+		analyzeTemplate:   analyzeTemplate,
+		anonymizeTemplate: anonymizeTemplate,
+	}
+}
+
+// ScanJSON scan the json in DFS to get to all the nodes
+func (jsonCrawler *JSONCrawler) ScanJSON(schemaMap map[string]interface{}, valuesMap map[string]interface{}) error {
```
eladiw: DFS is a good idea. 👍 
However, you implemented it using recursion and not using a stack. The only caveat with this approach is that json with many level might get you stack overflow...
so, either make sure the code is robust for these cases or try to avoid recursion.
I also believe this will help run time.

pkg/jsonCrawler/json-crawler.go
```diff
@@ -0,0 +1,189 @@
...
+
+func (jsonCrawler *JSONCrawler) scanArray(schemaArray []interface{}, valuesArray []interface{}) error {
+	err := checkIfEmptyArray(schemaArray, valuesArray)
+	if err != nil {
+		return err
+	}
+
+	i := 0
+	for j := range valuesArray {
+		if len(schemaArray) > 1 {
```
eladiw: can you add a comment about this snippet. not sure I get it...

pkg/jsonCrawler/json-crawler.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	if valType == "map" {
+		return jsonCrawler.ScanJSON(val.(map[string]interface{}), newVal.(map[string]interface{}))
+	}
+
+	return jsonCrawler.scanArray(val.([]interface{}), newVal.([]interface{}))
+}
+
+func (jsonCrawler *JSONCrawler) analyzeAndAnonymizeJSON(val string, field string) (string, error) {
+	match, err := regexp.MatchString("<[A-Z]+(_*[A-Z]*)*>", field)
```
eladiw: move the hard coded regex to the top of the file and give it a meaningful name/comment regarding what it finds.
e.g.
const namesRegex="...."

pkg/jsonCrawler/json-crawler.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	if match {
+		for key := range types.FieldTypesEnum_value {
+			fieldName := field[1 : len(field)-1]
```
eladiw: what happens if field == "" , meaning len(field)=0?

pkg/jsonCrawler/json-crawler_test.go
```diff
@@ -0,0 +1,206 @@
...
+		}
+	]}`)
+
+	crawler := New(context.Background(), analyzeFuncMock, anonymizeFuncMock, &types.AnalyzeTemplate{}, &types.AnonymizeTemplate{})
+
+	// Act
+	crawler.ScanJSON(jsonSchema, jsonToAnonymize)
+
+	// Assert
+	expextedResult, _ := unmarshalJSON(`{"patients":[
```
eladiw: typo.
'expected'

pkg/jsonCrawler/json-crawler_test.go
```diff
@@ -0,0 +1,206 @@
...
```
eladiw: typo...

README.MD
Ilanak: you're right, but we need to make other changes in the api - we will do it separately

PR #98, opened 2019-02-12
Presidio support for language code in template

itye-msft:
**Note to reviewer:** This PR is dependent on [presidio proto change merge](https://github.com/Microsoft/presidio-genproto/pull/13)
Added a unit test
updated docker file and linting errors

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -52,7 +52,7 @@
   name = "github.com/korovkin/limiter"
 
 [[constraint]]
-  branch = "move-lang-code"
+  branch = "development"
   name = "github.com/Microsoft/presidio-genproto"
 
 [[constraint]]
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -52,7 +52,7 @@
   name = "github.com/korovkin/limiter"
 
 [[constraint]]
-  branch = "master"
+  branch = "move-lang-code"
   name = "github.com/Microsoft/presidio-genproto"
 
 [[constraint]]
--- a/presidio-analyzer/.dockerignore
+++ b/presidio-analyzer/.dockerignore
@@ -5,7 +5,6 @@
 !model*
 !setup*
 !requirements.txt
-!tests/*
 # Make an exception for private SDKs if they exist
 !privates/*
 # Make an exception for the license file
--- a/presidio-analyzer/.dockerignore
+++ b/presidio-analyzer/.dockerignore
@@ -5,7 +5,6 @@
 !model*
 !setup*
 !requirements.txt
-!requirements-dev.txt
 !tests/*
 # Make an exception for private SDKs if they exist
 !privates/*
--- a/presidio-analyzer/.dockerignore
+++ b/presidio-analyzer/.dockerignore
@@ -5,6 +5,8 @@
 !model*
 !setup*
 !requirements.txt
+!requirements-dev.txt
+!tests/*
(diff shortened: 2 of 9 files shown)


presidio-analyzer/.dockerignore
```diff
@@ -5,6 +5,8 @@
 !model*
 !setup*
 !requirements.txt
+!requirements-dev.txt
```
omri374: The test fails because it can't find requirements-dev.txt 
We should have this as a part of the dockerignore, it is required for running tests

Gopkg.toml
```diff
@@ -40,14 +45,14 @@
...
-  version = "6.13.2"
+  version = "6.15.1"
 
 [[constraint]]
   branch = "master"
   name = "github.com/korovkin/limiter"
 
 [[constraint]]
-  branch = "master"
+  branch = "move-lang-code"
```
eladiw: you cant merge with this change

omri374: This is in continuation of the discussion around testing with new protocols. We won't merge it this way

omri374: changed to "development"

presidio-api/cmd/presidio-api/api/analyze/analyze_test.go
```diff
@@ -74,3 +74,17 @@ func TestAnalyzeWithNoTemplate(t *testing.T) {
...
+	api := setupMockServices()
+
+	project := "tests"
+	analyzeAPIRequest := &types.AnalyzeApiRequest{
+		Text:              "My number is (555) 253-0000 and email johnsnow@foo.com",
+		AnalyzeTemplateId: "test",
+		AnalyzeTemplate:   &types.AnalyzeTemplate{},
+	}
+	Analyze(context.Background(), api, analyzeAPIRequest, project)
+	assert.Equal(t, "langtest", analyzeAPIRequest.AnalyzeTemplate.LanguageCode)
```
lawrencegripper: What test cases would cause a failure here? From reading the `Analyze` method the majority of this test is asserting that the `mock` returns the language code expected.

itye-msft: @lawrencegripper , since we copy manually files from the proto-gen project,
Failure would occur if the template was not generated as expected. We are testing to verify the existence of the LanguageCode. 
Since no code change were made, but only proto template change were made, we are testing that the change exists.

presidio-analyzer/tests/test_analyzer_engine.py
```diff
@@ -139,3 +140,36 @@ def test_remove_analyzer(self):
                                      language='en')
 
         assert len(res) == 0
+
+    def test_Apply_with_language_returns_correct_response(self):
+        analyze_engine = AnalyzerEngine(MockRecognizerRegistry())
+
+        request = AnalyzeRequest()
+        request.analyzeTemplate.languageCode = 'en'
```
itye-msft: I don't understand how do you test the language?
You seem to define it, but how do you test the logic?

omri374: This is just setting up the request. Then, we call Apply(request,None) which is the function that's being used by the service. There's an additional test that shows it breaks if you don't define any language.

omri374: I changed the test to just test the AnalyzerEngine.Apply and not the inside recognizers.

PR #104, opened 2019-03-17
New functionality: Support custom pattern recognizers

eladiw:
_Issue 583._
This PR introduce a new functionality of support custom pattern based (regex) recognizers within Presidio.
Presidio user can use this functionality to define new regex expressions upload them into Presidio using a newly exposed REST command and use these regexex for text analysis.

_**This PR includes:**_
1) Changes to the API microservice, exposing management commands such as get/get all/update/delete and insert of new recognizers
2) A new microservice called recognizers_store which abstracts the underlying storage (redis) and is responsible for the custom recognizers. each microservice who needs custom recognizers capabilities talks to this micro service (currently these microservices are the API and the Analyzer)
3) Changes in the analyzer microservice to support custom recognizers
4) Helm chart updates to incorporate the new micro-service
5) Makefile changes to incorporate the new micro-service
6) Unit & E2E tests

_**Notes:**_
1. Currently works with the already existing redis. next PR will add a new, internal, redis
2. pb2 files are autogenerated, no need to review


presidio-recognizers-store/cmd/presidio-recognizers-store/main.go
```diff
@@ -0,0 +1,348 @@
...
+
+func getExistingRecognizers() ([]types.PatternRecognizer, error) {
+	recognizersArr := make([]types.PatternRecognizer, 0)
+	existingItems, err := recognizersStore.Get(recognizersKey)
+	if err == nil && existingItems != "" {
+		log.Info("Found existing recognizers...")
+		var recognizers []types.PatternRecognizer
+		err = json.Unmarshal([]byte(existingItems), &recognizers)
+		if err != nil {
+			return nil, err
```
lawrencegripper: If the key contains information that cannot be unmarshalled the system will remain in a failed state until there is some user interaction to clear down the key. 

Maybe one option would be to clear down the key data if it's in a bad state before returning the error.

presidio-recognizers-store/cmd/presidio-recognizers-store/main.go
```diff
@@ -0,0 +1,348 @@
...
+		if found == false {
+			return errors.New("custom pattern recognizer with name '" +
+				newRecognizer.Name + "' was not found")
+		}
+	}
+
+	recognizersBytes, _ := json.Marshal(existingRecognizersArr)
+	currentTimeStr := strconv.FormatInt(time.Now().Unix(), 10)
+
+	return setValue(string(recognizersBytes), currentTimeStr)
```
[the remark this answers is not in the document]
lawrencegripper: An option to resolve this would be to move to using the [Redis list](https://redis.io/topics/data-types#lists) datatype

charts/presidio/templates/recognizers-store-deployment.yaml
```diff
@@ -0,0 +1,50 @@
...
+kind: Deployment
+metadata:
+  name: {{ $fullname }}
+  labels:
+    app: {{ $fullname }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+spec:
+  replicas: 1
```
lawrencegripper: A single replica won't provide DR/HA within the cluster. 

Setting replicas to 2 will ensure mutliple instance exist in the cluster at any time and when combinded with [PodAntiAffinity](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#never-co-located-in-the-same-node) ensure that both pods don't run on a single node. 

With that in place a single node failuring or becoming stuck in the cluster won't cause a service outage.

charts/presidio/templates/recognizers-store-deployment.yaml
```diff
@@ -0,0 +1,50 @@
...
+    spec:
+      containers:
+      - name: {{ .Chart.Name }}
+        image: "{{ .Values.registry }}/{{ .Values.recognizersstore.name }}:{{ default .Chart.AppVersion .Values.tag }}"
+        imagePullPolicy: {{ default "IfNotPresent" .Values.recognizersstore.imagePullPolicy }}
+        ports:
+        - containerPort: {{ .Values.recognizersstore.service.internalPort }}
+        resources:
+          requests:
+            memory: "1500Mi"
```
lawrencegripper: This seems like a pretty large memory allocation for a go service or am I missunderstanding.

presidio-api/cmd/presidio-api/main.go
```diff
@@ -124,6 +125,25 @@ func setupHTTPServer(port int, loglevel string) {
 			projects.POST("/schedule-streams-job", scheduleStreamJob)
 		}
 
+		// recognizers group
+		// /api/v1/analyzer/recognizers
+		recognizers := v1.Group("/analyzer/recognizers")
+		{
+			// Get an existing recognizer
+			recognizers.GET(":id", getRecognizer)
```
lawrencegripper: The REST -> GRPC mapping could be handled by a [GRPC-Gateway](https://github.com/grpc-ecosystem/grpc-gateway), however, I've never used this and would say it's out of scope of this PR. 

It would be good to track this as an issue and maybe spike to see if this can be done.

charts/presidio/values.yaml
```diff
@@ -10,6 +10,11 @@ redis:
     #db: 0
     #ssl: false
 
+# Update url with the internal redis url, phase 2
+redis_internal:
+    nameOverride: persistent-redis
+    url: redis-master.presidio-system.svc.cluster.local:6379
```
lawrencegripper: Is this Redis deployed by the Chart or expected to already exist in the cluster? Probably me being silly but I can't find any reference to it in the Chart information.

presidio-recognizers-store/cmd/presidio-recognizers-store/main.go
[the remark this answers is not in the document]
lawrencegripper: Or look at using a more transnational store like PostgresSQL, K8s CRDs etc

presidio-recognizers-store/cmd/presidio-recognizers-store/main.go
```diff
@@ -0,0 +1,348 @@
...
+	"github.com/Microsoft/presidio/pkg/platform"
+	"github.com/Microsoft/presidio/pkg/presidio/services"
+	"github.com/Microsoft/presidio/pkg/rpc"
+)
+
+type server struct{}
+
+var (
+	settings         *platform.Settings
+	recognizersStore cache.Cache
```
lawrencegripper: My view is the Redis is best used as a cache not a persistent storage. While it's easy to use maybe something more focused on persistent data with query support etc would be worth considering. Cosmos can be easily accessed from Golang using the Mongo API, for example. See: https://github.com/lawrencegripper/ion/blob/master/internal/app/handler/dataplane/documentstorage/mongodb/mongodb.go

presidio-analyzer/analyzer/recognizer_registry/recognizers_store_api.py
```diff
@@ -0,0 +1,70 @@
...
+        logging.info("Latest timestamp found in store is: %d", lst_update)
+        return lst_update
+
+    def get_all_recognizers(self):
+        """
+        Returns a list of CustomRecognizer which were created from the
+        recognizers stored in the underlying store
+        """
+        req = recognizers_store_pb2.RecognizersGetAllRequest()
+        raw_recognizers = self.rs_stub.ApplyGetAll(req).recognizers
```
omri374: is exception handling needed here?

presidio-recognizers-store/cmd/presidio-recognizers-store/main.go
omri374: Would that work on-premises as well?

lawrencegripper: In the ION example we used the MongoAPI so this would work with a standard Mongo Server on premise and Cosmos when on Azure. When using more complex Mongo features the Cosmos support is limited but for this usecase I wouldn't see that being a problem.

presidio-api/cmd/presidio-api/api/recognizers/recognizers.go
```diff
@@ -0,0 +1,62 @@
...
+
+	store "github.com/Microsoft/presidio/presidio-api/cmd/presidio-api/api"
+)
+
+//InsertRecognizer Inserts a new custom pattern recognizer via the Recognizer
+// store service
+func InsertRecognizer(ctx context.Context,
+	api *store.API,
+	request *types.RecognizerInsertOrUpdateRequest) (
+	*types.RecognizersStoreResponse, error) {
```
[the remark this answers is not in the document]
eladiw: 👍

charts/presidio/templates/recognizers-store-deployment.yaml
eladiw: Correct. typo

charts/presidio/values.yaml
eladiw: This redis is **CURRENTLY**  deployed explicitly before Presidio is deployed.
in [this PR](https://github.com/Microsoft/presidio/pull/106), I am installing Redis internally with Presidio.

presidio-analyzer/analyzer/predefined_recognizers/custom_recognizer.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def __init__(self, name, patterns, entity, language, black_list,
+                 context=[]):
+        patterns_list = []
+        for pat in patterns:
+            patterns_list.extend([Pattern(pat.name, pat.regex, pat.score)])
+        super().__init__(name=name, supported_entity=entity,
+                         patterns=patterns_list,
+                         context=context, supported_language=language,
+                         black_list=black_list)
```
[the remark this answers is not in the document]
eladiw: 👍

charts/presidio/templates/recognizers-store-deployment.yaml
eladiw: This is a default value, same for the others micro services.
I opened a new task to keep track and consider how we change this

presidio-api/cmd/presidio-api/main.go
eladiw: thanks, added as a new task

presidio-recognizers-store/cmd/presidio-recognizers-store/main.go
eladiw: Redis is not the first choice for persistent storage. however it does it well and can be used as an on premise solution as well (containerized).
It can support high scale and fast query time.

Also, the plan is to use this Redis as a cache in other places, so it simplifies the solution to have one such component (while satisfying both performance / requirements)

presidio-analyzer/analyzer/recognizer_registry/recognizer_registry.py
```diff
@@ -1,78 +1,57 @@
+import time
 import logging
 
-from analyzer import PatternRecognizer
+from analyzer.recognizer_registry.recognizers_store_api \
+    import RecognizerStoreApi  # noqa: F401
```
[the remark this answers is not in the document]
eladiw: 👍

presidio-analyzer/analyzer/recognizer_registry/recognizers_store_api.py
eladiw: 👍

presidio-analyzer/analyzer/recognizer_registry/recognizer_registry.py
```diff
@@ -96,8 +75,67 @@ def get_recognizers(self, entities=None, language=None):
             else:
                 to_return.extend(subset)
 
+        logging.info("Found %d predefined recognizers", len(to_return))
+        custom = self.get_custom_recognizers()
+        logging.info("Found %d (total) custom recognizers", len(custom))
+        for entity in entities:
```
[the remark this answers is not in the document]
eladiw: 👍

presidio-recognizers-store/cmd/presidio-recognizers-store/main.go
eladiw: A valid point. however, this can happen for example due to a change in the types.PatternRecognizer class which is not backward compatible or any other issue.
Deleting the recognizer (from the storage) as a self healing solution can be done, however this can lead to lose of configuration which is irreversible.

I think that in the current suggested implementation, if the unmarshaling fails you get an error logged and the process continues without the custom recognizers, so the system remains fully operational, until some user manually fix it, or a new version is deployed (if thats the case)

@omri374  WDYT?

pkg/presidio/presidio.go
```diff
@@ -31,6 +32,13 @@ type ServicesAPI interface {
 	ApplyScan(ctx context.Context, scanJobRequest *types.ScannerCronJobRequest) (*types.ScannerCronJobResponse, error)
 	InitDatasink(ctx context.Context, datasinkTemplate *types.DatasinkTemplate) (*types.DatasinkResponse, error)
 	CloseDatasink(ctx context.Context, datasinkTemplate *types.CompletionMessage) (*types.DatasinkResponse, error)
+
+	InsertRecognizer(ctx context.Context, rec *types.PatternRecognizer) (*types.RecognizersStoreResponse, error)
+	UpdateRecognizer(ctx context.Context, rec *types.PatternRecognizer) (*types.RecognizersStoreResponse, error)
+	DeleteRecognizer(ctx context.Context, name string) (*types.RecognizersStoreResponse, error)
+	GetRecognizer(ctx context.Context, name string) (*types.RecognizersGetResponse, error)
+	GetAllRecognizers(ctx context.Context) (*types.RecognizersGetResponse, error)
+	GetUpdateTimeStamp(ctx context.Context) (*types.RecognizerTimestampResponse, error)
```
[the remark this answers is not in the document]
eladiw: Agree, changing to a hash based method.

presidio-recognizers-store/cmd/presidio-recognizers-store/main.go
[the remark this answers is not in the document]
eladiw: I've elaborated why I think Redis is a good option [here](https://github.com/Microsoft/presidio/pull/104#discussion_r268660677).
Since in this implementation I chose to keep all the recognizers under a single key 'custom_recognizers' it makes it very easy to retrieve with a single query (although insert/update/delete are harder) and it helps with the timestamp creation (which I will change to etag as discussed above). 

Moving to Redis list DS will not completely remove the possibility of race, although it will decrease the probability.
Another assumption we can leverage, is that writing recognizers is _very very rare_ and we can compromise on the time it takes for a write operation to finish.
Hence, I suggest adding a mutex in insert/update/delete operations, this will completely remove the race issue and at the same time wont affect the 'get' performance we strive to keep as fast as possible.

presidio-recognizers-store/cmd/presidio-recognizers-store/main.go
omri374: If there's information that cannot be unmarshalled, we'd lose it anyway, wouldn't we? Services would not be able to parse it and return it. However I agree that the system can still remain functional (although it's not ideal). @eladiw let's discuss this offline and decide what would be the best way.

presidio-recognizers-store/cmd/presidio-recognizers-store/main.go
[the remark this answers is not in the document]
lawrencegripper: Mutex would work as long as there is only ever one instance of the store service running. 

I think either using `Transactions` or `List` operations (Less sure on this one) in redis would allow you to avoid the race condition while sticking with this approach ~~BUT I've never tried these out.~~

https://redis.io/topics/transactions
https://stackoverflow.com/questions/40999446/how-to-create-redis-transaction-in-go-using-go-redis-redis-package

I tried this out and it works nicely and uses the transaction removes the race for multi deployments. 

In the code below the first go routine writes every 1 second to the key and the second reads then pauses for 5 seconds before attempting a while (simulating the read then write in your code). The transaction fails when twice as the key is changes between the read and write then succeeds on the last try as go routine 1 has stopped updating the key. 

```
15:48 $ go build . && ./go-redis-list 
#PONG <nil>
2019/03/26 15:49:10 #1ROUTINE: Key set outside transation to 2
2019/03/26 15:49:11 #1ROUTINE: Key set outside transation to 3
2019/03/26 15:49:12 #1ROUTINE: Key set outside transation to 4
2019/03/26 15:49:13 #1ROUTINE: Key set outside transation to 5
2019/03/26 15:49:14 #1ROUTINE: Key set outside transation to 6
2019/03/26 15:49:15 #2ROUTINE: Failed to write redis: transaction failed 
2019/03/26 15:49:15 #1ROUTINE: Key set outside transation to 7
2019/03/26 15:49:16 #1ROUTINE: Key set outside transation to 8
2019/03/26 15:49:17 #1ROUTINE: Key set outside transation to 9
2019/03/26 15:49:18 #1ROUTINE: Key set outside transation to 10
2019/03/26 15:49:20 #2ROUTINE: Failed to write redis: transaction failed 
2019/03/26 15:49:25 #2ROUTINE: Transaction updated key to: get RECOGKEY: get RECOGKEY: 10-after

```

```
package main

import (
	"fmt"
	"log"
	"sync"
	"time"

	"github.com/go-redis/redis"
)

func main() {
	client := redis.NewClient(&redis.Options{
		Addr:     "localhost:6379",
		Password: "", // no password set
		DB:       0,  // use default DB
	})

	pong, err := client.Ping().Result()
	fmt.Println(pong, err)
	// Output: PONG <nil>

	key := "RECOGKEY"

	client.Set(key, "1", time.Duration(time.Second*360))

	wg := sync.WaitGroup{}
	wg.Add(2)

	go func() {
		i := 1
		for {
			i++

			if i > 10 {
				break
			}

			client.Set(key, i, time.Duration(time.Second*360))
			log.Printf("#1ROUTINE: Key set outside transation to %v", i)

			time.Sleep(1 * time.Second)
		}
		wg.Done()
	}()

	go func() {
		for {
			err := client.Watch(func(tx *redis.Tx) error {
				n := tx.Get(key).String()

				_, err = tx.Pipelined(func(pipe redis.Pipeliner) error {
					time.Sleep(time.Second * 5)
					pipe.Set(key, n+"-after", 0)
					return nil
				})
				return err
			}, key)

			if err != nil {
				log.Printf("#2ROUTINE: Failed to write %+v \n", err)
				continue
			}

			value := client.Get(key)
			log.Printf("#2ROUTINE: Transaction updated key to: %+v", value)

			break
		}
		wg.Done()
	}()

	wg.Wait()
}
```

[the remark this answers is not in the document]
lawrencegripper: (Not saying you have to do this just was interested to see if it was possible to do it with Redis as I'd not used their transations before)

presidio-recognizers-store/cmd/presidio-recognizers-store/main.go
eladiw: Discussed. Agreed that 'auto recovery' is not feasible and should not be a priority at this point.
Also, deleting these items may pose bigger risk than just ignoring.

Decided to leave as is

presidio-recognizers-store/cmd/presidio-recognizers-store/main.go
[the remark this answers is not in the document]
eladiw: The caveat with Redis transactions is that you can't get a value, process it and then set it.
As the Get command will be 'Queued', meaning the value won't be available until Exec is called.

There probably other solutions (such as lua scripts).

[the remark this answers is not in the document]
eladiw: To sum this conversation:
We expect very few changes to the config and even less possibilities for such a race condition. 
The mutex is a simple yet effective solution for 1 replica, if someone needs to scale this microservice, it is still possible, the race can then still happen but with even lower possibility as this means two commands arrived at the same time to two different replicas.

Marking as resolved

presidio-analyzer/analyzer/recognizer_registry/recognizers_store_api.py
```diff
@@ -0,0 +1,79 @@
...
+        try:
+            raw_recognizers = self.rs_stub.ApplyGetAll(req).recognizers
+
+        except grpc.RpcError:
+            logging.info("Failed getting recognizers from the remote store. \
+            Returning an empty list")
+            return raw_recognizers
+
+        custom_recognizers = []
+        for new_recognizer in raw_recognizers:
```
omri374: Can we use the from_dict method to instantiate a recognizer?

presidio-analyzer/analyzer/recognizer_registry/recognizer_registry.py
```diff
@@ -1,68 +1,56 @@
...
+               a store that uses a persistent storage
+        :param recognizers: An optional list of recognizers that will be
+               available in addition to the predefined recognizers and the
+               custom recognizers
+        """
+        if recognizers:
+            self.recognizers = recognizers
+        else:
+            self.recognizers = []
+        self.loaded_hash = None
```
omri374: Non-urgent - please add some doc on what is loaded_hash and loaded_timestamp

presidio-analyzer/analyzer/recognizer_registry/recognizers_store_api.py
eladiw: I will add it in a later phase

eladiw:
@omri374 @lawrencegripper the changes based on your comments were incorporated

lawrencegripper:
@omri374 Just taking a quick look now should be done in next 10/15mins

presidio-api/cmd/presidio-api/api/recognizers/recognizers.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	return "Recognizer was updated successfully", nil
+}
+
+// DeleteRecognizer deletes an existing recognizer via the Recognizer
+// store service
+func DeleteRecognizer(ctx context.Context,
+	api *store.API,
+	request *types.RecognizerDeleteRequest,
+) (string, error) {
```
lawrencegripper: Nice, imo looks much more readable now.

presidio-analyzer/analyzer/recognizer_registry/recognizer_registry.py
```diff
@@ -30,7 +30,11 @@ def __init__(self, recognizer_store_api=RecognizerStoreApi(),
             self.recognizers = recognizers
         else:
             self.recognizers = []
+        # loaded_hash is used to store the currently loaded hash of in memory
+        # recognizers, to avoid fetching the recognizers from t
```
omri374: Sentence is cut in the middle. A suggestion:
loaded_hash is the hash value of the recognizers in the recognizer store. It is used to avoid fetching recognizers when there hasn't been a change to the recognizers state.

PR #106, opened 2019-03-20
Internal redis

eladiw:
Installing Redis as an internal presidio component.
All redis clients are now referencing this instance


eladiw:
Closing for some more work as discussed offline.

PR #107, opened 2019-03-21
All fields

itye-msft:
In this PR I have added a simple flag in protobuf template called `allFields` which allows to analyze all fields if it is set to True.


presidio-analyzer/analyzer/analyzer_engine.py
```diff
@@ -117,6 +118,24 @@ def __convert_fields_to_entities(self, fields):
...
+        """
+        Returns a list of fields. If the "allFields" flag if True, all fields will be returned. Otherwise,
+        the template fields will be returned.
+
+        :param analyzeTemplate: The template to process.
+        :return: a list of fields
+        """
+        if analyzeTemplate.allFields:
+            fields = []
+            for item in common_pb2.FieldTypesEnum.keys():
```
omri374: This is not going to work for custom recognizers. the allFields param should skip the filtering of recognizers by field

presidio-analyzer/analyzer/analyzer_engine.py
```diff
@@ -64,22 +65,25 @@ def Apply(self, request, context):
...
 
     def get_language_from_request(self, request):
-        language = request.analyzeTemplate.languageCode
+        language = request.analyzeTemplate.language
         if language is None or language == "":
             language = DEFAULT_LANGUAGE
         return language
 
-    def analyze(self, text, entities, language):
+    def analyze(self, text, entities, language, all_fields=False):
```
eladiw: default 'False' is not needed as the only place that calls this method is lie 57, which passes a value.
you probably need to update some tests after your remove the default value

presidio-analyzer/analyzer/recognizer_registry/recognizer_registry.py
```diff
@@ -63,38 +63,40 @@ def remove_recognizer(self, name):
...
-
-            if len(subset) == 0:
-                logging.warning(
-                    "Entity " + entity +
-                    " doesn't have the corresponding recognizer in language :"
-                    + language)
-            else:
-                to_return.extend(subset)
+        if all_fields:
+            to_return = [rec for rec in self.recognizers if
```
eladiw: you can merge  this:

> [rec for rec in self.recognizers 
>  and language == rec.supported_language]

and then just filter for additional entities... (less duplicate code)

presidio-analyzer/analyzer/recognizer_registry/recognizer_registry.py
```diff
@@ -63,38 +63,40 @@ def remove_recognizer(self, name):
...
+                         language == rec.supported_language]
+        else:
+            for entity in entities:
+                subset = [rec for rec in self.recognizers if
+                          entity in rec.supported_entities
+                          and language == rec.supported_language]
+
+                if len(subset) == 0:
+                    logging.warning(
```
eladiw: this line is going to fail linting once the other lint pr is merged .
go for:
"text text %s text", "text"
style

PR #111, opened 2019-04-03
Entities list for all fields true

omri374:
Bug fix for all_fields=True.
Currently: Calling presidio's API with all_fields=True and a Spacy entity (person, location) does not return an answer. SpacyRecognizer expects entities in order to filter the ones needs to be evaluated. 
New logic gathers entities from all recognizers, in case of all_fields=True and sends it to recognizers, in order not to change recognizer internal logic.

### DIFF
--- a/presidio-analyzer/tests/test_analyzer_engine.py
+++ b/presidio-analyzer/tests/test_analyzer_engine.py
@@ -253,3 +253,15 @@ class TestAnalyzerEngine(TestCase):
         assert "PERSON" in returned_entities
         assert "LOCATION" in returned_entities
         assert "DOMAIN_NAME" in returned_entities
+
+    def test_when_allFields_is_true_and_entities_not_empty_exception(self):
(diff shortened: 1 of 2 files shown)


presidio-analyzer/tests/test_analyzer_engine.py
```diff
@@ -234,5 +234,22 @@ def test_when_allFields_is_true_return_all_fields(self):
         request.text = " Credit card: 4095-2609-9393-4932,  my phone is 425 8829090 " \
             "Domain: microsoft.com"
         response = analyze_engine.Apply(request, None)
+        returned_entities = [field.field.name for field in response.analyzeResults]
```
eladiw: Can you add one test when you specify all fields and specify ONE entity type and verify you get an error

omri374: sure

PR #120, opened 2019-04-21
updated docs and samples based on new changes in dev

itye-msft:


docs/field_types.md
```diff
@@ -232,4 +232,12 @@
     </tr>
   </tbody>
 </table>
+<hr/>
+<h2>All fields</h2>
+In case you need to load all available fields in your Presidio deployment, you need to set the `allFields` flag to `true`.
```
omri374:
```suggestion
In case you need Presidio to return all possible PII entities, in your Presidio deployment, you need to set the `allFields` flag to `true`.
```

itye-msft: done

docs/install.md
```diff
@@ -7,7 +7,7 @@ You can install Presidio as a service in [Kubernetes](https://kubernetes.io/) or
 ```sh
 # Build the images
 
-$ export DOCKER_REGISTRY=presidio
+$ export DOCKER_REGISTRY=mcr.microsoft.com
```
eladiw: I think this is wrong. as 'presidio' is not really an ACR.
meaning this export line is overriding the default ACR so that all images will be pulled locally.
so this change is wrong in this context.

itye-msft: @eladiw Done

PR #121, opened 2019-04-24
Add public path to makefile

itye-msft:
task 875
On push release, also push to mcr, to make the image public


itye-msft:
> Hi,
> For completeness, add in the values.yaml (chart) file that the default registry is the mcr registry
> and update the image names. they should work out of the box with the public images.
> 
> here is an example of a public pull command
> docker pull mcr.microsoft.com/presidio/presidio-analyzer

@eladiw  DONE

PR #122, opened 2019-04-28
Build deps - adding build number to python&golang dependency containers

balteravishay:
This commit includes the two new Azure pipelines which are based on the checked in yaml files:
Presidio-Python-Deps-CI
Presidio-Golang-Deps-CI

they both only add the registry name as variable, which is added as a docker registry to the project services.


Makefile
```diff
@@ -44,13 +45,18 @@ docker-build: docker-build-base
...
+	docker build $(DOCKER_BUILD_FLAGS) --build-arg REGISTRY=$(DOCKER_REGISTRY) --build-arg VERSION=$(VERSION) --build-arg DEPS_LABEL=$(PRESIDIO_DEPS_LABEL) -t $(DOCKER_REGISTRY)/$*:$(PRESIDIO_LABEL) -f $*/Dockerfile .
 
 # You must be logged into DOCKER_REGISTRY before you can push.
 .PHONY: docker-push-deps
 docker-push-deps: 
+	docker push $(DOCKER_REGISTRY)/$(PYTHON_DEPS):$(PRESIDIO_DEPS_LABEL)
+	docker push $(DOCKER_REGISTRY)/$(GOLANG_DEPS):$(PRESIDIO_DEPS_LABEL)
+	docker image tag $(DOCKER_REGISTRY)/$(PYTHON_DEPS):$(PRESIDIO_DEPS_LABEL) $(DOCKER_REGISTRY)/$(PYTHON_DEPS):latest
+	docker image tag $(DOCKER_REGISTRY)/$(GOLANG_DEPS):$(PRESIDIO_DEPS_LABEL) $(DOCKER_REGISTRY)/$(GOLANG_DEPS):latest 
 	docker push $(DOCKER_REGISTRY)/$(PYTHON_DEPS):latest
```
itye-msft: if $(PRESIDIO_DEPS_LABEL) is `latest` than this line will actually be called twice (with line 53)

docs/development.md
```diff
@@ -64,9 +60,19 @@
 
 ## Development notes
 
+- Set version numbers:
```
itye-msft: Is it mandatory or optional?

docs/development.md
```diff
@@ -64,9 +60,19 @@
...
+    $ export PRESIDIO_DEPS_LABEL=[version number]
+    ```
+- Set registry name:
+    ```sh
+    $ export DOCKER_REGISTRY=[registry login server]
+    ```
 - Build the bins with `make build`
-- Build the the Docker image with `make docker-build`
-- Push the Docker images with `make docker-push`
+- Build the basecontainers with `make docker-build-deps DOCKER_REGISTRY=${DOCKER_REGISTRY} PRESIDIO_DEPS_LABEL=${PRESIDIO_DEPS_LABEL}`
```
itye-msft: Is registry mandatory? are there any defaults to mention?

docs/development.md
```diff
@@ -64,9 +60,19 @@
 
 ## Development notes
 
+- Set version numbers:
+    ```sh
+    $ export PRESIDIO_LABEL=[version number]
```
itye-msft: by saying `version number` do you mean `label`? I mean in docker terminology

docs/development.md
itye-msft:
```suggestion
- Build the base containers with `make docker-build-deps DOCKER_REGISTRY=${DOCKER_REGISTRY} PRESIDIO_DEPS_LABEL=${PRESIDIO_DEPS_LABEL}`
```

Makefile
balteravishay: correct, fixed with ifneq to "latest"

docs/development.md
balteravishay: for dev, it should be, otherwise, you'll end up with 'latest'
thoughts?

docs/development.md
balteravishay: should be label, you're right

docs/development.md
balteravishay: the default is ours, i think it should be removed from code. 
thoughts?

pipelines/presidio-golang-deps-CI.yaml
```diff
@@ -0,0 +1,27 @@
+trigger: # Build triggers
```
eladiw: looks good. however, now we have in the base code the ci definition.... 
which is the first time we save it in the source itself. for example presidio's main ci is not saved in this manner...

docs/development.md
balteravishay: fixed, added "optional" information

docs/development.md
balteravishay: fixed, added "optional" information

PR #126, opened 2019-04-29
Bug fixes - based on bugbash

eladiw:
Bugs 906, 908, 909


presidio-analyzer/analyzer/entity_recognizer.py
```diff
@@ -198,12 +208,13 @@ def __add_n_words(index,
                             take the successing words
         """
         i = index
-        # collect at most n words
+        # collect at most n words (in lower case)
         remaining = n_words
         while 0 <= i < len(lemmas) and remaining > 0:
-            if lemmas[i] in lemmatized_filtered_keywords:
+            lower_lemma = lemmas[i].lower()
```
itye-msft: I am not sure `toLower` is the best practice.
[https://github.com/explosion/spaCy/issues/2420](https://github.com/explosion/spaCy/issues/2420)
[https://spacy.io/api/token#attributes](https://spacy.io/api/token#attributes) (search `lower`)

omri374: Take a look at this: https://github.com/explosion/spaCy/issues/1901
When we don't use a tagger (which is disabled in our case), spacy does a lookup table based lemmatization. When using a tagger, the lemmas are always lowercase. This should be a better solution.

presidio-analyzer/analyzer/entity_recognizer.py
```diff
@@ -157,21 +157,31 @@ def __calculate_context_similarity(self,
...
-                self.logger.info("Found context keyword '%s'", context_keyword)
+        for predefined_context_word in context_list:
+            # result == true only if any of the predefined context words
+            # is found exactly or as a substring in any of the collected
+            # context words
+            result = \
+              next((True for keyword in lemmatized_keywords
+                    if predefined_context_word in keyword), False)
+            if result:
+                self.logger.info("Found context keyword '%s'",
```
omri374:
```suggestion
                self.logger.debug("Found context keyword '%s'",
```

presidio-analyzer/analyzer/entity_recognizer.py
```diff
@@ -198,12 +208,19 @@ def __add_n_words(index,
                             take the successing words
         """
         i = index
-        # collect at most n words
+        # The entity itself is no intrest to us... skip it
+        if is_backward:
```
omri374:
```suggestion
        i = i-1 if is_backward else i+1
```

presidio-analyzer/analyzer/entity_recognizer.py
eladiw: Good points, however @itye-msft, I am interested in lemmas and not tokens, so this solution doesnt apply.
@omri374 not sure why, but disabling the tagger didn't work

presidio-analyzer/analyzer/entity_recognizer.py
eladiw: 👍

presidio-analyzer/analyzer/entity_recognizer.py
eladiw: 👍

presidio-analyzer/analyzer/entity_recognizer.py
```diff
@@ -156,22 +156,32 @@ def __calculate_context_similarity(self,
         """Context similarity is 1 if there's exact match between a keyword in
            context_text and any keyword in context_list
 
-        :param context_text a string of the prefix and suffix of the found
-               match
+        :param context_text words before and after the matched enitity within
```
omri374:
```suggestion
        :param context_text words before and after the matched entity within
```

PR #130, opened 2019-04-30
Change deployment script to mcr

itye-msft:


deployment/deploy-presidio.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-REGISTRY=${1:-presidio.azurecr.io}
+REGISTRY=${1:-mcr.microsoft.com/presidio/}
```
balteravishay: all pushes will be done to mcr? will this be controllable?
what permission is required to commit to mcr?

eladiw: should be:
mcr.microsoft.com

PR #131, opened 2019-04-30
Bug 911 fix

eladiw:
When the persistent recognizers store is empty there should be no error log line when requesting for the hash value


presidio-analyzer/analyzer/recognizer_registry/recognizer_registry.py
```diff
@@ -125,7 +125,7 @@ def get_custom_recognizers(self):
 
         latest_hash = self.store_api.get_latest_hash()
         # is update time is not set, no custom recognizers in storage, skip
-        if latest_hash != "":
+        if latest_hash is not None:
```
omri374:
```suggestion
        if latest_hash:
```

presidio-recognizers-store/cmd/presidio-recognizers-store/main.go
```diff
@@ -298,10 +298,16 @@ func applyGetHash() (*types.RecognizerHashResponse, error) {
...
+	if err != nil {
 		errMsg := "Failed to find the latest hash"
 		log.Error(errMsg)
-		return &types.RecognizerHashResponse{}, errors.New(errMsg)
+		return nil, errors.New(errMsg)
+	}
+
+	// no error, however hash was not found, this means the redis is still
+	// empty.
+	if hash == "" {
```
balteravishay: shouldn't this be 
if not hash {

eladiw: Short answer is 'no'.
long answer is that you are correct. redis returns redis.Nil in the error object if item is not found.
however in our implementation ([Line 48-49](https://github.com/Microsoft/presidio/blob/development/pkg/cache/redis/redis.go#L48)) we return "" in such case (our redis wrapper)
so thats why my check is correct

PR #133, opened 2019-04-30
Dev docs

balteravishay:


docs/install.md
```diff
@@ -50,13 +50,24 @@ $ docker run --rm --name presidio-api --network mynetwork -d -p 8080:8080 -e WEB
 
 4. Verify that Redis and Traefik/NGINX are installed correctly
 
-5. Deploy from `/charts/presidio`
+5. Verify that Kubernetes has access to the [Docker Registry](https://docs.microsoft.com/en-us/azure/container-registry/container-registry-auth-aks).
 
+    - If using a Kubernetes secert, make sure it is registered under 'presidio' namespace
```
omri374: I think we should elaborate on this, or add a link to documentation. Is this a Presidio requirement or a user requirement?

omri374:
Do you think we should add documentation on how to install presidio based on images in MCR?

docs/install.md
```diff
@@ -50,13 +50,24 @@ $ docker run --rm --name presidio-api --network mynetwork -d -p 8080:8080 -e WEB
...
+    - If using a public Docker Registry:
     ```sh
     # Based on the DOCKER_REGISTRY and PRESIDIO_LABEL from the previous steps
     $ helm install --name presidio-demo --set registry=${DOCKER_REGISTRY},tag=${PRESIDIO_LABEL} . --namespace presidio
     ```
 
+    - If using a private Docker Registry:
+    ```sh
+    # Based on the DOCKER_REGISTRY and PRESIDIO_LABEL from the previous steps
+    $ helm install --name presidio-demo --set registry=${DOCKER_REGISTRY},tag=${PRESIDIO_LABEL},privateRegistry=${SECRET_NAME} . --namespace presidio
```
eladiw: @itye-msft we should align this file with the README.MD file.
The README.MD file is referring to the scripts your wrote in the deployment dir.

docs/install.md
balteravishay: its linked now, and moved to development as well.

omri374: I personally don't think that the installation page should direct to the readme. Maybe the other way around. the installation should have different installation options, and the readme can direct the user to a specific option or to the installation page in general. WDYT?

balteravishay: true, but I wanted to mention the quick installation option, detailed in the readme.
would you rather remove that part from the readme and move it to the installation page?

docs/development.md
```diff
@@ -96,3 +97,15 @@
...
     ```
+
+
+## Running in kubernetes
+
+    1. If deploying from a private registry, verify that Kubernetes has access to the [Docker Registry](https://docs.microsoft.com/en-us/azure/container-registry/container-registry-auth-aks).
+
+    2. If using a Kubernetes secert to manage the registry authentication, make sure it is registered under 'presidio' namespace
+
+    3. Edit charts/presidio/values.yaml to:
```
eladiw: you can link to the values.yaml file ... which is nicer.

docs/development.md
```diff
@@ -96,3 +97,15 @@
     ```sh
     wrk -t2 -c2 -d30s -s post.lua http://<api-service-address>/api/v1/projects/<my-project>/analyze
     ```
+
+
+## Running in kubernetes
+
+    1. If deploying from a private registry, verify that Kubernetes has access to the [Docker Registry](https://docs.microsoft.com/en-us/azure/container-registry/container-registry-auth-aks).
```
eladiw: mention that this link is an example relevant just for AKS not any k8s

docs/install.md
eladiw: what about this @balteravishay  @itye-msft

docs/development.md
balteravishay: The link shows also k8s native ways to do it, using secrets. it's not aks specific

docs/development.md
balteravishay: done

PR #159, opened 2019-05-15
Fix context bug where indices were off

eladiw:


presidio-analyzer/tests/test_context_support.py
```diff
@@ -90,3 +91,23 @@ def test_text_with_context_improves_score(self):
             assert(len(results_without_context) == len(results_with_context))
             for i in range(len(results_with_context)):
                 assert(results_without_context[i].score < results_with_context[i].score)
+
+    def test_text_with_context_improves_score_custom_recognizer(self):
```
omri374: Is there a way to test the specific index problem? The test seems rather generic

eladiw: I elaborated on the special case of the test + made a unit test just for this specific part

eladiw:
I also added context list to the integration test of the custom recognizer, which failed prior to this fix

PR #174, opened 2019-05-30
Release Candidate 0.3

eladiw:


Gopkg.toml
```diff
@@ -52,7 +52,7 @@
   name = "github.com/korovkin/limiter"
 
 [[constraint]]
-  branch = "master"
+  branch = "development"
```
balteravishay: is this right?

Makefile
```diff
@@ -11,6 +11,7 @@ GOLANG_BASE	= presidio-golang-base
 GIT_TAG   = $(shell git describe --tags --always 2>/dev/null)
 VERSION   ?= ${GIT_TAG}
 PRESIDIO_LABEL := $(if $(PRESIDIO_LABEL),$(PRESIDIO_LABEL),$(VERSION))
+PRESIDIO_DEPS_LABEL := $(if $(PRESIDIO_DEPS_LABEL),$(PRESIDIO_DEPS_LABEL),'latest')
```
omri374: Sorry if I missed that previously. Isn't this going to be problematic if I didn't specify a value for PRESIDIO_DEPS_LABEL? I might push the latest. Am I correct?

docs/development.md
```diff
@@ -88,11 +123,56 @@
...
+### Developing only for Presidio Analyzer under Windows environment
+Run locally the core services Presidio needs to operate:
+```
+docker run --rm --name test-redis --network testnetwork -d -p 6379:6379 redis
+docker run --rm --name test-presidio-anonymizer --network testnetwork -d -p 3001:3001 -e GRPC_PORT=3001 mcr.microsoft.com/presidio-anonymizer:latest
+docker run --rm --name test-presidio-recognizers-store --network testnetwork -d -p 3004:3004 -e GRPC_PORT=3004 -e REDIS_URL=test-redis:6379 mcr.microsoft.com/presidio-recognizers-store:latest
+```
+Naviagate to `<Presidio folder>\presidio-analyzer\`
+
+Install the python packages if didn't do so yet:
```
omri374: re2 is missing in this pipeline

Gopkg.toml
eladiw: No, fixed

Makefile
balteravishay: it means using 'latest' as the default for pulls mostly. 
the makefile allows pushing as well, but you have to run the specific command to do it.

docs/development.md
omri374: Actually let's leave it like this, as re2 is really hard to build in windows

Makefile
balteravishay: also, it corresponds to the previous way builds were done and was added to support the previous CIs

PR #175, opened 2019-05-30
remove push deps with default tag

balteravishay:


Makefile
```diff
@@ -48,11 +48,6 @@ docker-build: $(addsuffix -dimage,$(IMAGES))
 	docker build $(DOCKER_BUILD_FLAGS) --build-arg REGISTRY=$(DOCKER_REGISTRY) --build-arg VERSION=$(VERSION) --build-arg PRESIDIO_DEPS_LABEL=$(PRESIDIO_DEPS_LABEL) -t $(DOCKER_REGISTRY)/$*:$(PRESIDIO_LABEL) -f $*/Dockerfile .
 
 # You must be logged into DOCKER_REGISTRY before you can push.
-.PHONY: docker-push-deps
```
omri374: We're assuming that a developer isn't pushing a non-latest image from anywhere but the CI.

PR #185, opened 2019-06-11
Dumping interpretability information into logs

eladiw:


presidio-analyzer/analyzer/analysis_explanation.py
```diff
@@ -0,0 +1,38 @@
+class AnalysisExplanation:
+    """ AnalysisExplanation is a class that holds forensic information that
```
omri374: Not sure 'forensic' is the right word here, Dexter :)

presidio-analyzer/analyzer/analysis_explanation.py
```diff
@@ -0,0 +1,38 @@
...
+    def __init__(self, recognizer, pattern_name,
+                 pattern, original_score, free_text=None):
+        """
+        """
+        self.recognizer = recognizer
+        self.pattern_name = pattern_name
+        self.pattern = pattern
+        self.original_score = original_score
+        self.score = original_score
+        self.free_text = free_text
```
omri374: Consider renaming. maybe textual_explanation or human_readable_explanation?

presidio-analyzer/analyzer/analysis_explanation.py
```diff
@@ -0,0 +1,38 @@
...
+        """
+        self.score = score
+        self.score_context_improvement = self.score - self.original_score
+
+    def set_supportive_context_word(self, word):
+        """ Sets the context word which helped increase the score
+        """
+        self.supportive_context_word = word
+
+    def append_free_text_line(self, text):
```
omri374: add docstring

presidio-analyzer/analyzer/analyzer_engine.py
```diff
@@ -1,47 +1,54 @@
...
-                 nlp_engine=SpacyNlpEngine()):
+                 nlp_engine=SpacyNlpEngine(),
+                 ml_tracer=AppTracer()):
         # load nlp module
         self.nlp_engine = nlp_engine
         # prepare registry
         self.registry = registry
         # load all recognizers
         registry.load_predefined_recognizers()
+        self.ml_tracer = ml_tracer
```
omri374: consider renaming. We don't only trace ml models. We do it for the entire system.

pkg/presidio/presidio.go
```diff
@@ -19,7 +19,7 @@ type ServicesAPI interface {
 	SetupDatasinkService()
 	SetupRecognizerStoreService()
 	SetupCache() cache.Cache
-	AnalyzeItem(ctx context.Context, text string, template *types.AnalyzeTemplate) ([]*types.AnalyzeResult, error)
+	AnalyzeItem(ctx context.Context, text string, template *types.AnalyzeTemplate) (*types.AnalyzeResponse, error)
```
omri374: This has to go with the presidio-genproto change. Correct? if we want to do it in two phases, this should be reverted.

presidio-analyzer/analyzer/analyzer_engine.py
```diff
@@ -1,47 +1,54 @@
...
     def Apply(self, request, context):
-        logging.info("Starting Apply")
+        logger.info("Starting Analyzer's Apply")
+
         entities = AnalyzerEngine.__convert_fields_to_entities(
             request.analyzeTemplate.fields)
         language = AnalyzerEngine.get_language_from_request(request)
-        results = self.analyze(request.text, entities, language,
+
+        generated_request_id = str(uuid.uuid4())
```
omri374: If we generate it here, it should be called differently. "correlation_id" or something like that. It's not a request Id because it doesn't come with the request

presidio-analyzer/analyzer/analyzer_engine.py
```diff
@@ -108,6 +118,10 @@ def analyze(self, text, entities, language, all_fields):
         # run the nlp pipeline over the given text, store the results in
         # a NlpArtifacts instance
         nlp_artifacts = self.nlp_engine.process_text(text, language)
+
+        self.ml_tracer.trace(request_id,
+                             "nlp artifacts:" + repr(nlp_artifacts))
```
omri374: Better to use the \_\_repr\_\_ function inside nlp_artifacts. This is prone to errors.

omri374:
```suggestion
                             "nlp artifacts: {}".format(repr(nlp_artifacts)))
```

presidio-analyzer/analyzer/analyzer_engine.py
```diff
@@ -120,7 +134,10 @@ def analyze(self, text, entities, language, all_fields):
             if current_results:
                 results.extend(current_results)
 
-        return AnalyzerEngine.__remove_duplicates(results)
+        results = AnalyzerEngine.__remove_duplicates(results)
+        self.ml_tracer.trace(request_id, repr(results))
```
omri374: If previous call used a title ("nlp artifacts: X"), please use the same scheme

presidio-analyzer/analyzer/app_tracer.py
```diff
@@ -0,0 +1,29 @@
+import logging
+
+
+class AppTracer:
+    """This class handles app traces"""
```
omri374: This is also affecting the logger, isn't it? Maybe we should combine the two and add a 'trace' call to the logger (We can potentially have two instances of this, but it's subject to making sure that the logger isn't a singleton or something like that)

presidio-analyzer/analyzer/app_tracer.py
```diff
@@ -0,0 +1,29 @@
...
+            logger.addHandler(ch)
+            logger.setLevel(logging.INFO)
+            logger.propagate = False
+
+        self.logger = logger
+        self.enable_interpretability = enable_interpretability
+
+    def trace(self, request_id, trace_data):
+        """
+        Writes interpretability trace
```
omri374: Here too. Assume a new developer doesn't know what we mean by 'interpretability'

presidio-analyzer/analyzer/entity_recognizer.py
```diff
@@ -38,9 +38,13 @@ def __init__(self, supported_entities, name=None, supported_language="en",
...
         loglevel = os.environ.get("LOG_LEVEL", "INFO")
         self.logger = logging.getLogger(__name__)
+        logging.basicConfig(
+            format='%(asctime)s:%(levelname)s:[%(req_id)s]%(message)s',
+            level=loglevel)
+
         self.logger.setLevel(loglevel)
         self.load()
-        logging.info("Loaded recognizer: %s", self.name)
+        logging.info("Loaded recognizer: %s", self.name, extra={'req_id': ''})
```
omri374: Is this still required?

presidio-analyzer/analyzer/predefined_recognizers/credit_card_recognizer.py
```diff
@@ -33,6 +33,10 @@ def __init__(self):
     def validate_result(self, pattern_text, pattern_result):
         sanitized_value = CreditCardRecognizer.__sanitize_value(pattern_text)
         res = CreditCardRecognizer.__luhn_checksum(sanitized_value)
+
+        # Send interpretability traces
+        pattern_result.append_analysis_explenation_text(
+            "Credit card checksum result: {}".format(str(res)))
```
omri374: Why not add a field for checksum/verification logic? This is quite generic. The textual representation should be used only for specific use cases

presidio-analyzer/analyzer/recognizer_result.py
```diff
@@ -1,15 +1,29 @@
+from . import AnalysisExplanation
+
+
 class RecognizerResult:
 
-    def __init__(self, entity_type, start, end, score):
+    def __init__(self, entity_type, start, end, score,
+                 analysis_explanation: AnalysisExplanation = None):
```
omri374: Consider adding a field for textual explanations that will be used onlly if the analysis_explanation is None (only one of the two could be given). This would make life easier for recognizer developers that don't know how to use the AnalysisExplanation class, but want to provide some textual information about how their recognizer decided on an output. 
Something like:

`

     def __init__(self, entity_type, start, end, score, analysis_explanation: AnalysisExplanation=None, textual_result_explanation: str=None):

        if analysis_explantion and textual_result_explanation:
            raise ValueError("Expecting either analysis_explanation or textual_result_explanation, not both")
    
        if not analysis_explanation:
            analysis_explanation = AnalysisExplanation(free_text=textual_result_explanation)
`

presidio-analyzer/analyzer/entity_recognizer.py
```diff
@@ -132,28 +136,34 @@ def enhance_using_context(self, text, raw_results,
...
                     self.MIN_SCORE_WITH_CONTEXT_SIMILARITY)
                 result.score = min(
                     result.score,
                     EntityRecognizer.MAX_SCORE)
+
+                # Update the explainability object with context information
+                # helped improving the score
+                result.analysis_explanation.set_supportive_context_word(
+                    supportive_context_word)
+                result.analysis_explanation.set_improved_score(result.score)
```
omri374: Here we should add the call for .set_pattern_verification_output (checksum passed yes or no)

presidio-analyzer/analyzer/analyzer_engine.py
```diff
@@ -108,6 +118,10 @@ def analyze(self, text, entities, language, all_fields):
         # run the nlp pipeline over the given text, store the results in
         # a NlpArtifacts instance
         nlp_artifacts = self.nlp_engine.process_text(text, language)
+
+        self.ml_tracer.trace(request_id,
```
omri374: This outputs a list of AnalysisExplanation. Consider putting this trace inside the loop and have one trace per recognizer. It might make reading easier.

presidio-analyzer/analyzer/nlp_engine/spacy_nlp_engine.py
```diff
@@ -18,7 +18,7 @@ def __init__(self):
         self.logger = logging.getLogger(__name__)
         self.logger.setLevel(loglevel)
 
-        self.logger.info("Loading NLP model...")
+        logging.info("Loading NLP model...", extra={'req_id': ''})
```
omri374: Is this still required?

presidio-analyzer/presidio-venv/bin/activate
```diff
@@ -0,0 +1,76 @@
+# This file must be used with "source bin/activate" *from bash*
+# you cannot run it directly
```
omri374: This file should not be committed

presidio-analyzer/presidio-venv/bin/activate.csh
```diff
@@ -0,0 +1,37 @@
+# This file must be used with "source bin/activate.csh" *from csh*.
```
omri374: This file should not be committed

presidio-analyzer/presidio-venv/bin/activate.fish
```diff
@@ -0,0 +1,75 @@
+# This file must be used with ". bin/activate.fish" *from fish* (http://fishshell.org)
```
omri374: This file should not be committed

presidio-analyzer/presidio-venv/bin/easy_install
```diff
@@ -0,0 +1,10 @@
+#!/Users/ityer/go/src/github.com/Microsoft/presidio/presidio-analyzer/presidio-venv/bin/python3.7
```
omri374: This file should not be committed

presidio-analyzer/presidio-venv/bin/easy_install-3.7
```diff
@@ -0,0 +1,10 @@
+#!/Users/ityer/go/src/github.com/Microsoft/presidio/presidio-analyzer/presidio-venv/bin/python3.7
```
omri374: This file should not be committed

presidio-analyzer/presidio-venv/bin/pip
```diff
@@ -0,0 +1,10 @@
+#!/Users/ityer/go/src/github.com/Microsoft/presidio/presidio-analyzer/presidio-venv/bin/python3.7
```
omri374: This file should not be committed

presidio-analyzer/presidio-venv/bin/pip
```diff
@@ -0,0 +1,10 @@
...
+# -*- coding: utf-8 -*-
+import re
+import sys
+
+from pip._internal import main
+
```
omri374: This file should not be committed

presidio-analyzer/presidio-venv/bin/pip3.7
```diff
@@ -0,0 +1,10 @@
+#!/Users/ityer/go/src/github.com/Microsoft/presidio/presidio-analyzer/presidio-venv/bin/python3.7
```
omri374: This file should not be committed

presidio-analyzer/presidio-venv/bin/python
```diff
@@ -0,0 +1 @@
+python3.7
```
omri374: This file should not be committed

presidio-analyzer/presidio-venv/bin/python3
```diff
@@ -0,0 +1 @@
+python3.7
```
omri374: This file should not be committed

presidio-analyzer/presidio-venv/bin/python3.7
```diff
@@ -0,0 +1 @@
+/Library/Frameworks/Python.framework/Versions/3.7/bin/python3.7
```
omri374: This file should not be committed

presidio-analyzer/presidio-venv/pyvenv.cfg
```diff
@@ -0,0 +1,3 @@
+home = /Library/Frameworks/Python.framework/Versions/3.7/bin
```
omri374: This file should not be committed

presidio-analyzer/tests/test_recognizer_registry.py
```diff
@@ -49,7 +49,7 @@ def add_custom_pattern_recognizer(self, new_recognizer,
         self.latest_hash = m.digest()
 
     def remove_recognizer(self, name):
-        logging.info("removing recognizer " + name)
+        logging.info("removing recognizer %s", name)
```
omri374:
```suggestion
        logging.info("removing recognizer {}".format(name))
```

pkg/server/server.go
```diff
@@ -93,6 +93,17 @@ func WriteResponse(
...
 }
 
+//WriteResponseWithRequestID writes a response and adds a request id header
+func WriteResponseWithRequestID(
+	c *gin.Context,
+	statusCode int,
+	requestID string,
+	responseBody interface{},
+) {
+	c.Header("X-Request-Id", requestID)
```
balteravishay: x-request-id is usually populated by client, and not by server. does x-correlation-id fit better? 
https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#cite_note-stackoverflow1-39 
https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#cite_note-stackoverflow1-40
https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#cite_note-stackoverflow1-41 
https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#cite_note-stackoverflow1-42

presidio-analyzer/analyzer/analysis_explanation.py
balteravishay: description ?

presidio-analyzer/analyzer/analysis_explanation.py
eladiw: 🤣

presidio-analyzer/presidio-venv/bin/python
eladiw: The entire presidio-venv should be deleted

pkg/server/server.go
```diff
@@ -10,7 +10,7 @@ import (
 	"time"
 
 	"github.com/gin-contrib/cors"
-	"github.com/gin-contrib/zap"
+	ginzap "github.com/gin-contrib/zap"
```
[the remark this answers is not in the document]
itye-msft: I don't recall changing this. perhaps it's an auto-changed by IDE etc.

presidio-analyzer/tests/test_recognizer_registry.py
itye-msft: This logger has it's own internal string formatting, and I'm in favour of using it

pkg/server/server.go
itye-msft: yes I agree

pkg/presidio/presidio.go
itye-msft: To my understanding of Elad's code, this is a response of the API, and therefore, is not changing the genproto

presidio-analyzer/analyzer/analyzer_engine.py
itye-msft: This is a temporal fix. It should be recieved from the API. not generated in the analyzer. 
we wanted to minimize the scope of this change, and later on to change it so the ID is generated by the API.

presidio-analyzer/analyzer/app_tracer.py
itye-msft: I am not sure I understand.
Is it a semantic request? I can leave out the word "interpretability"

presidio-analyzer/analyzer/entity_recognizer.py
itye-msft: I was not aware of this line, good catch. I'll remove it

presidio-analyzer/analyzer/predefined_recognizers/credit_card_recognizer.py
itye-msft: Do you mean that  all recognizers will have checksum/verification logic?
I also thought the point was to make an example that utilizes the solution

presidio-analyzer/analyzer/entity_recognizer.py
itye-msft: But you wanted the result of the checksum, not only if it passed. Am I wrong?

presidio-analyzer/analyzer/nlp_engine/spacy_nlp_engine.py
itye-msft: I have no idea what it is, removed.

presidio-analyzer/analyzer/recognizer_result.py
itye-msft: Actually, this is not a good solution, because you can create a new instance of `AnalysisExplanation` in `RecognizerResult`: you don't have in hand the parameters needed to instantiate it. I think it's a good enough at this point to accept the `AnalysisExplanation` as is, and have more verbose doc string of how to add simple text

pkg/presidio/presidio.go
omri374: I'm not sure. Check out https://github.com/microsoft/presidio-genproto/pull/22/files

presidio-analyzer/analyzer/analysis_explanation.py
```diff
@@ -0,0 +1,40 @@
...
+        """
+        self.recognizer = recognizer
+        self.pattern_name = pattern_name
+        self.pattern = pattern
+        self.original_score = original_score
+        self.score = original_score
+        self.textual_explanation = textual_explanation
+        self.score_context_improvement = 0
+        self.supportive_context_word = ''
+
```
omri374:
```suggestion
self.validation_result = validation_result
```
We are providing a generic explanation object for pattern recognizers. Validation (checksum) is done automatically, whether the user implemented the validation logic or not. That's why the validation result should also be part of the explanation output.

presidio-analyzer/analyzer/analyzer_engine.py
omri374: Regardless of this being temporary or not, it's confusing to call it "generated request id". Even if we plan to refactor it at a later phase, it should still be named properly, which is a correlation_id or input_id or something in that nature.

presidio-analyzer/analyzer/analyzer_engine.py
```diff
@@ -108,6 +118,10 @@ def analyze(self, text, entities, language, all_fields):
         # run the nlp pipeline over the given text, store the results in
         # a NlpArtifacts instance
         nlp_artifacts = self.nlp_engine.process_text(text, language)
+
+        self.app_tracer.trace(request_id, "nlp artifacts:"
+                              + repr(nlp_artifacts))
```
omri374: `repr` is usually used for debugging and is usually a short version of the object's values. If we want to store the entire object into the log, we should provide a method for nlp_artifacts to serialize itself, such as `to_dict` or `to_json`.

presidio-analyzer/analyzer/analyzer_engine.py
```diff
@@ -120,7 +134,10 @@ def analyze(self, text, entities, language, all_fields):
             if current_results:
                 results.extend(current_results)
 
-        return AnalyzerEngine.__remove_duplicates(results)
+        results = AnalyzerEngine.__remove_duplicates(results)
+        self.app_tracer.trace(request_id, repr(results))
```
omri374: Same here. it's better to avoid `repr` if we want to make sure the entire object is serialized

presidio-analyzer/analyzer/app_tracer.py
omri374: The entire documentation of this class is very limited. We want a newcomer to understand why we have both Logger and AppTracer, and how the trace method should be used. It's ok to have the word "interpretability", but it should be much more detailed. For example:
```
class AppTracer:
    """
    This class provides the ability to log/trace the system's decisions, 
    such as which modules were used for detection, 
    which logic was utilized, what results were given and potentially why. 
    This can be useful for analyzing the detection accuracy of the system.
    """
```

```
def trace(self, request_id, trace_data):
       """
       Writes a value associated with a decision for a specific request into the trace, 
       for further inspection if needed.
       """
```

presidio-analyzer/analyzer/entity_recognizer.py
omri374: Yes. In many cases it's the same, but providing the full checksum result is preferred.

presidio-analyzer/analyzer/entity_recognizer.py
```diff
@@ -38,6 +38,10 @@ def __init__(self, supported_entities, name=None, supported_language="en",
 
         loglevel = os.environ.get("LOG_LEVEL", "INFO")
```
omri374: This should be removed

presidio-analyzer/analyzer/entity_recognizer.py
```diff
@@ -38,6 +38,10 @@ def __init__(self, supported_entities, name=None, supported_language="en",
...
         self.logger = logging.getLogger(__name__)
+        logging.basicConfig(
```
omri374: This should be removed

presidio-analyzer/analyzer/logger.py
```diff
@@ -0,0 +1,74 @@
+import logging
```
omri374: I don't see why we need this class. First, not all modules use it. Second, Python best practices is to define logger configuration once (e.g. on \_\_main\_\_) and the use `logger = logging.getLogger(__name__)` on each module. Right now we have this class + multiple logger configurations + calls to the logging module directly. Let's try to standardize everything.

presidio-analyzer/analyzer/pattern_recognizer.py
```diff
@@ -99,6 +100,16 @@ def validate_result(self, pattern_text, pattern_result):
...
 
+    @staticmethod
+    def build_regex_explanation(
+            recognizer_name,
+            pattern_name,
+            pattern,
+            original_score):
+        explanation = AnalysisExplanation(recognizer_name, pattern_name,
+                                          pattern, original_score)
+        return explanation
```
omri374: checksum (validation) output should be added here too

presidio-analyzer/analyzer/predefined_recognizers/credit_card_recognizer.py
omri374: The generic analysis_explanation should contain a value for checksum/validation too. 
We can provide other examples if necessary, but the validation is a part of the generic process and could be automatically traced too.

presidio-analyzer/analyzer/predefined_recognizers/spacy_recognizer.py
```diff
@@ -15,6 +15,12 @@ def load(self):
         # preprocessed nlp artifacts
         pass
 
+    @staticmethod
+    def build_spacy_explanation(recognizer_name, original_score):
+        explanation = AnalysisExplanation(recognizer_name, '',
```
omri374: Here we could provide an example of how to add a textual explanation

presidio-analyzer/analyzer/recognizer_result.py
omri374: I'm not sure what would be best. In other languages we would do overloading. In python the closest solution is to add more parameters. Let's take it offline and decide.

pkg/server/server.go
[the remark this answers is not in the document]
omri374: Could you please verify this isn't a bug?

.gitignore
```diff
@@ -89,6 +89,7 @@ venv/
 ENV/
 env.bak/
 venv.bak/
+*venv/
```
omri374: Files under this folder are still committed. Please remove them

presidio-analyzer/analyzer/nlp_engine/spacy_nlp_engine.py
```diff
@@ -18,7 +18,7 @@ def __init__(self):
         self.logger = logging.getLogger(__name__)
         self.logger.setLevel(loglevel)
```
omri374: Please remove (together with the loglevel param)

presidio-analyzer/analyzer/logger.py
itye-msft: I totally agree, but standardising **everything** should be in a different pr. It's a different concern, which must be taken care of. I will open a new task for it

pkg/presidio/presidio.go
itye-msft: correct

presidio-analyzer/analyzer/logger.py
omri374: This file is new and was introduced in this PR. If we do or don't standardize everything, I don't believe we need it.

presidio-analyzer/analyzer/app_tracer.py
```diff
@@ -0,0 +1,29 @@
+import logging
+
+
+class AppTracer:
+    """This class handles app traces"""
+    def __init__(self, enable_interpretability=True):
+
```
balteravishay: enable_interpretability controls trace output. i would remove interpretability from the name.

presidio-analyzer/analyzer/app_tracer.py
```diff
@@ -0,0 +1,29 @@
...
+        self.logger = logger
+        self.enable_interpretability = enable_interpretability
+
+    def trace(self, request_id, trace_data):
+        """
+        Writes interpretability trace
+        :param request_id: A unique ID, to correlate across calls.
+        :param trace_data: A string to write.
+        :return:
+        """
```
balteravishay: is there room for additional argument such as "trace_type"? currently trace is just a dump of all kinds of "logs" (result, artefacts, etc) and it may be hard to go through them if not typed in some way.

presidio-analyzer/analyzer/logger.py
balteravishay: agree. If this can be resolved at the __main__ level it would be better.

presidio-analyzer/analyzer/app_tracer.py
itye-msft: This class accepts only string as input.
Are you suggesting other types?

presidio-analyzer/analyzer/logger.py
itye-msft: The practice on the "wrapper" class is to give you the ability to change the logging mechanism instantly across the the app, without changing the app.
 
It looks like there are many practices about how to use the "logging" module, and using the name of the module `logger = logging.getLogger(__name__)` is one of them.

The wrapper class initialises the correct way. It can be extended to provide the `getLogger` functionality easily, and let other class do it - although in Presidio, no code is doing it anyway. And I don't see any reason to do it in this "explainability" PR. It deserves a separate PR.

The current suggested change was required because it was not possible to use different loggers, due to wrong instantiation of the logger in the analyzer_engine code. 
I **Had** to introduce a the minimal code change required to have 2 different loggers. Wrapping the logging module was the only way to do it, without radical code changes.

The change is at point that:
1. The logging and tracing works, and the PR has minimal code changes for a single iteration.
2. There are many places in the python files which still use logging module, and it needs to be changed, and aligned to the wrapper logger. This tho, is not critical to the PR, and because of the nature of it - that it includes so many files, I suggest to make the changes in a dedicated PR, and not in the Tracing PR, because it crosses the gentle line of separation of concerns.

presidio-analyzer/analyzer/analyzer_engine.py
```diff
@@ -28,23 +28,30 @@ def __init__(self, registry=None, nlp_engine=None):
...
     def Apply(self, request, context):
-        logging.info("Starting Apply")
+        logger.info("Starting Analyzer's Apply")
+
         entities = AnalyzerEngine.__convert_fields_to_entities(
             request.analyzeTemplate.fields)
         language = AnalyzerEngine.get_language_from_request(request)
-        results = self.analyze(request.text, entities, language,
+
+        correlation_id = str(uuid.uuid4())
```
omri374:
```suggestion
       #correlation is used to group all traces related to on request
        correlation_id = str(uuid.uuid4())
```

Or something like that

presidio-analyzer/analyzer/analyzer_engine.py
```diff
@@ -123,7 +136,10 @@ def analyze(self, text, entities, language, all_fields):
             if current_results:
                 results.extend(current_results)
 
-        return AnalyzerEngine.__remove_duplicates(results)
+        results = AnalyzerEngine.__remove_duplicates(results)
+        self.app_tracer.trace(correlation_id, repr(results))
```
omri374: Better not to use repr as we don't always know how it is implemented

presidio-analyzer/analyzer/nlp_engine/nlp_artifacts.py
```diff
@@ -33,3 +33,9 @@ def set_keywords(nlp_engine, lemmas, language):
         keywords = \
             [item for sublist in keywords for item in sublist]
         return keywords
+
+    def __repr__(self):
+        return str(self.__dict__)
+
+    def to_json(self):
+        return self.__repr__()
```
omri374: Why go through two functions? 
```suggestion
        return str(self.__dict__)
```

presidio-analyzer/analyzer/predefined_recognizers/spacy_recognizer.py
omri374: Please add argument names to function call, and some text about this being identified by Spacy as the interpretation

presidio-analyzer/tests/test_analyzer_engine.py
```diff
@@ -9,15 +9,15 @@
...
+    RecognizerResult, RecognizerRegistry, AnalysisExplanation
 from analyzer.predefined_recognizers import CreditCardRecognizer, \
     UsPhoneRecognizer, DomainRecognizer, UsItinRecognizer, \
     UsLicenseRecognizer, UsBankRecognizer, UsPassportRecognizer
 from analyzer.recognizer_registry.recognizers_store_api \
     import RecognizerStoreApi  # noqa: F401
 from analyzer.nlp_engine import SpacyNlpEngine, NlpArtifacts
 from analyzer.predefined_recognizers import IpRecognizer, UsSsnRecognizer
-from tests.mocks import MockNlpEngine
+from tests.mocks import MockNlpEngine, app_tracer_mock
```
omri374: Why not import the class itself?
add AppTracerMock to the mock's \_\_init\_\_.py file

```suggestion
from tests.mocks import MockNlpEngine, AppTracerMock
```

presidio-analyzer/tests/test_analyzer_engine.py
```diff
@@ -85,19 +85,23 @@ def load_recognizers(self, path):
...
 
+
 class TestAnalyzerEngine(TestCase):
 
     def __init__(self, *args, **kwargs):
         super(TestAnalyzerEngine, self).__init__(*args, **kwargs)
         self.loaded_registry = MockRecognizerRegistry(RecognizerStoreApiMock())
         mock_nlp_artifacts = NlpArtifacts([], [], [], [], None, "en")
-        self.loaded_analyzer_engine = AnalyzerEngine(self.loaded_registry, MockNlpEngine(stopwords=[], punct_words=[], nlp_artifacts=mock_nlp_artifacts))
+        self.app_tracer = app_tracer_mock.AppTracerMock(enable_interpretability=True)
```
omri374:
```suggestion
        self.app_tracer = AppTracerMock(enable_interpretability=True)
```

presidio-analyzer/tests/test_analyzer_engine.py
```diff
@@ -282,4 +289,12 @@ def test_when_allFields_is_true_and_entities_not_empty_exception(self):
         new_field.minScore = '0.5'
         with pytest.raises(ValueError):
             analyze_engine.Apply(request, None)
-    
\ No newline at end of file
+
+    def test_when_analyze_then_mltracer_has_value(self):
```
omri374:
```suggestion
    def test_when_analyze_then_apptracer_has_value(self):
```

presidio-analyzer/tests/test_analyzer_engine.py
```diff
@@ -282,4 +289,12 @@ def test_when_allFields_is_true_and_entities_not_empty_exception(self):
...
+        text = " Credit card: 4095-2609-9393-4932,  my phone is 425 8829090"
+        language = "en"
+        entities = ["CREDIT_CARD", "PHONE_NUMBER"]
+        analyzer_engine_with_spacy = AnalyzerEngine(self.loaded_registry, app_tracer=self.app_tracer)
+        results = analyzer_engine_with_spacy.analyze(self.unit_test_guid, text, entities, language, all_fields=False)
+        assert len(results) == 2
+        assert self.app_tracer.get_last_trace() is not None
```
omri374: Better to test that it has all the different parts traced- NLP artifacts, AnalysisExplanations

omri374:
@itye-msft I've done another review. Looks really good. All the comments are minor. Please review and let's approve it.

@balteravishay Please review too once these are addressed

docs/interpretability_logs.md
```diff
@@ -0,0 +1,57 @@
...
+        recognizer=recognizer_name,
+        original_score=original_score,
+        textual_explanation=SPACY_DEFAULT_EXPLANATION.format(entity))
+    return explanation
+```
+
+The `textual_explanation` field in `AnalysisExplanation` class allows you to add your own custom text into the final trace which will be written.
+
+## Enabling/Disabling Traces
+Interpretability traces are aenabled by default. Disable App Tracing by setting the `enabled` constructor parameter to `False`.
```
balteravishay:
```suggestion
Interpretability traces are enabled by default. Disable App Tracing by setting the `enabled` constructor parameter to `False`.
```

PR #188, opened 2019-06-26
load spacy nlp engine once

balteravishay:


presidio-analyzer/analyzer/analyzer_engine.py
```diff
@@ -17,9 +14,14 @@
 
 class AnalyzerEngine(analyze_pb2_grpc.AnalyzeServiceServicer):
 
-    def __init__(self, registry=RecognizerRegistry(),
-                 nlp_engine=SpacyNlpEngine()):
+    def __init__(self, registry,
```
omri374: Not sure what the right answer is. In my opinion we should set up the default objects if nothing is passed.
```
def __init__(self, registry=None, nlp_engine=None):
    if not registry:
    if not nlp_engine:
        nlp_engine = SpacyNlpEngine()
```

This might make life easier for someone who's trying to initialize this class but isn't sure what the required input is.

balteravishay: that point was raised during code review with itye, and i agree with him that the right DI pattern is not to instantiate the injected component within the type. this way analyzer_engine does not know of recognizerregistry and spacynlpengine concrete types, which is the right approach.
perhaps a static helper/factory method is what we want?

omri374: WDYT about this option?

```
   def __init__(self, registry=None, nlp_engine=None):
        if not registry:        
            from analyzer import RecognizerRegistry 
        if not nlp_engine:
            from analyzer.nlp_engine import SpacyNlpEngine 
            nlp_engine = SpacyNlpEngine()
```

presidio-analyzer/analyzer/analyzer_engine.py
```diff
@@ -17,9 +14,16 @@
...
```
omri374:
```suggestion
    def __init__(self, registry=None, nlp_engine=None)
```
Otherwise the check `if not nlp_engine` only makes sense if someone explicitly passes None for both arguments

presidio-analyzer/analyzer/analyzer_engine.py
```diff
@@ -16,18 +16,18 @@ class AnalyzerEngine(analyze_pb2_grpc.AnalyzeServiceServicer):
...
+            from analyzer import RecognizerRegistry
             registry = RecognizerRegistry()
         # load nlp module
 
         self.nlp_engine = nlp_engine
         # prepare registry
         self.registry = registry
         # load all recognizers
-        registry.load_predefined_recognizers()
+        registry.load_predefined_58()
```
omri374: typo?

presidio-analyzer/tests/test_analyzer_engine.py
```diff
@@ -193,8 +193,8 @@ def test_removed_pattern_recognizer_doesnt_work(self):
 
         # Make sure the analyzer doesn't get this entity
         recognizers_store_api_mock = RecognizerStoreApiMock()
-        analyze_engine = AnalyzerEngine(MockRecognizerRegistry(
-            recognizers_store_api_mock))
+        analyze_engine = AnalyzerEngine(registry= MockRecognizerRegistry(
+            recognizers_store_api_mock), nlp_engine=SpacyNlpEngine())
```
omri374: I don't think we need actual Spacy for these tests. We can use a mock instead. https://github.com/microsoft/presidio/blob/74ea983cc50ff76d7fc734683ca897ff01ceb9ba/presidio-analyzer/tests/mocks/nlp_engine_mock.py

balteravishay:
Yes, already fixed. Thanks

charts/presidio/templates/analyzer-deployment.yaml
```diff
@@ -26,10 +26,10 @@ spec:
         - containerPort: {{ .Values.analyzer.service.internalPort }}
         resources:
           requests:
-            memory: "2000Mi"
+            memory: "1500Mi"
             cpu: "1500m"
           limits:
-            memory: "5000Mi"
+            memory: "3000Mi"
```
torosent: I think you also need to specify in the docs min requirements

balteravishay: added a note about cluster planning in install.md (https://github.com/microsoft/presidio/blob/spacy_load_twice/docs/install.md) and a note about adding more recognisers which may add to consumed mem\cpu in custom_fields.md (https://github.com/microsoft/presidio/blob/spacy_load_twice/docs/custom_fields.md). do you think it need more?

PR #189, opened 2019-07-09
Fix empty results

balteravishay:


presidio-api/cmd/presidio-api/api/analyze/analyze_test.go
```diff
@@ -104,3 +116,19 @@ func TestAllFields(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, 2, len(results))
 }
+
+func TestEmptyResponseIsOK(t *testing.T) {
```
itye-msft: When it comes to unit test, naming is crucial for quick understanding of what the test is about, I would consider renaming it.
Here is a naming convention to consider:
Test<FunctionName>When<Condition>Then<ExpectedResult>
```suggestion
func TestAnalyzeWhenNoEntitiesFoundThenExpectEmptyResponse(t *testing.T) {
```

PR #190, opened 2019-07-16
added threshold support for low scores

omri374:
This PR introduces a score threshold for each request. It allows users to filter low confidence results.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -107,11 +107,11 @@
 
 [[projects]]
   branch = "analyzer_threshold_support"
-  digest = "1:2b546a67a59237d4672b038d4d5429aebb8dc8b804869cf2fc7232137405ab95"
+  digest = "1:c79ff0c1cb49b28ad509306045dd8d5e46072354c0b4dbaff75434f3e630ce8d"
   name = "github.com/Microsoft/presidio-genproto"
   packages = ["golang"]
   pruneopts = "UT"
-  revision = "53ccfdef39bcf1bcbc1e444f6b2daa0aa91d502d"
+  revision = "0c1ca1066110a665c708a66d598f14b6080ab1f6"
 
 [[projects]]
   digest = "1:b449dbaada891dc97f016bc3519bb1af0e1a4296828f7a560a0ba47ea75f20bb"
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -106,12 +106,12 @@
   version = "v1.4.0"
 
 [[projects]]
-  branch = "analyzer_threshold_support"
+  branch = "development"
   digest = "1:c79ff0c1cb49b28ad509306045dd8d5e46072354c0b4dbaff75434f3e630ce8d"
   name = "github.com/Microsoft/presidio-genproto"
   packages = ["golang"]
   pruneopts = "UT"
-  revision = "0c1ca1066110a665c708a66d598f14b6080ab1f6"
+  revision = "db15bf6b1589b71e88d3fab12188c42d5364789c"
 
 [[projects]]
   digest = "1:2ec153af6a806c3d63d4299f2549bcb29d75d9703097341be309a46db3481488"
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -6,8 +6,8 @@
   name = "cloud.google.com/go"
   packages = ["civil"]
   pruneopts = "UT"
-  revision = "457ea5c15ccf3b87db582c450e80101989da35f7"
-  version = "v0.40.0"
+  revision = "cf81fad90a1a1de334c4fc27e23eb9a4224b627a"
+  version = "v0.41.0"
 
 [[projects]]
   digest = "1:6b1426cad7057b717351eacf5b6fe70f053f11aac1ce254bbf2fd72c031719eb"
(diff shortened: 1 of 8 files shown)


presidio-analyzer/analyzer/analyzer_engine.py
```diff
@@ -87,14 +87,24 @@ def __remove_duplicates(results):
 
         return filtered_results
 
+    @staticmethod
+    def __remove_low_scores(results, score_threshold):
+        new_results = []
+        for result in results:
+            if result.score > score_threshold:
```
eladiw: >=?

presidio-analyzer/tests/test_analyzer_engine.py
```diff
@@ -304,10 +333,50 @@ def test_when_analyze_then_apptracer_has_value(self):
...
+
+        # This analyzer engine is different from the global one, as this one
+        # also loads SpaCy so it can detect the phone number entity
+
+        analyzer_engine = AnalyzerEngine(
+            registry=self.loaded_registry, nlp_engine=MockNlpEngine())
+        results = analyzer_engine.analyze(self.unit_test_guid, text,
+                                          entities, language,
+                                          all_fields=False,
+                                          score_threshold=0.6)
```
eladiw: 0.6 -> 0.5 and change the threshold statement to >=
I think it makes more sense, also the name of this test says: "...is_half..."

omri374: I changed the test name. It has to be more than 0.5 because of the internal confidence of the pattern

presidio-analyzer/analyzer/analyzer_engine.py
omri374: Fixed

Gopkg.lock
```diff
@@ -73,12 +73,12 @@
   version = "v23.2.0"
 
 [[projects]]
-  digest = "1:b15d5bdadce5d98f1e06353508a4029fccfeb68761957b3a2a8cb38ebb8caca4"
+  digest = "1:9d3a690fcd29a7efbc783d7a8c9c173c2c67d33c0d2b6ac4be10c78d4ae46df2"
   name = "github.com/Azure/azure-storage-blob-go"
   packages = ["azblob"]
   pruneopts = "UT"
-  revision = "678206e7e6e55abf0265a6440135e92005176ebf"
-  version = "v0.6.0"
```
eladiw: These changes are not correct.
please update this file with:
dep ensure --update github.com/Microsoft/presidio-genproto

presidio-analyzer/analyzer/analyzer_engine.py
```diff
@@ -104,6 +118,8 @@ def analyze(self, correlation_id, text, entities, language, all_fields):
         :param language: the language of the text
         :param all_fields: a Flag to return all fields
         of the requested language
+        :param score_threshold: A minimumm value for which
```
eladiw: minimum -> typo

presidio-analyzer/analyzer/analyzer_engine.py
```diff
@@ -144,6 +160,7 @@ def analyze(self, correlation_id, text, entities, language, all_fields):
                 results.extend(current_results)
 
         results = AnalyzerEngine.__remove_duplicates(results)
+        results = AnalyzerEngine.__remove_low_scores(results, score_threshold)
```
eladiw: consider: adding something like:
total results: X, dropped due to low score: Y

maybe it will be useful when debugging in the future

omri374: I will put this line before the app tracer line. This will log all the results prior to removal. Would that be sufficient?

presidio-analyzer/analyzer/analyzer_engine.py
```diff
@@ -87,14 +91,24 @@ def __remove_duplicates(results):
 
         return filtered_results
 
+    @staticmethod
+    def __remove_low_scores(results, score_threshold):
```
balteravishay: re-iterates the scores which is un-optimal. consider inject a predicate to the analyze method

omri374: You mean to filter in advance? We'd like the entire list of results (prior to removal) to be traced

presidio-analyzer/analyzer/analyzer_engine.py
omri374: fixed

Gopkg.lock
omri374: Fixed (by you...)

presidio-analyzer/analyzer/analyzer_engine.py
```diff
@@ -87,14 +100,24 @@ def __remove_duplicates(results):
...
     @classmethod
     def get_language_from_request(cls, request):
         language = request.analyzeTemplate.language
         if language is None or language == "":
             language = DEFAULT_LANGUAGE
         return language
 
-    def analyze(self, correlation_id, text, entities, language, all_fields):
+    def analyze(self, correlation_id, text, entities, language, all_fields,
+                score_threshold=0.1):
```
itye-msft: It's nicer to have it as a const, not hard coded
```suggestion
                score_threshold=MIN_THRESHOLD):
```

omri374: I'm not sure that's the right thing. When working interactively with the code (i.e. in a jupyter notebook or similar), you will not have visibility to the const value. It is customary to have default numeric values for functions. In addition, when importing the function alone (without the module), the code would break for not importing the const.

eladiw: I prefer the const option. how about having the const in this file, so if using this module alone, you still have it?

omri374: Having the const in the file won't work if you import the function alone. Do you have any reference of places where default function arguments are const?

omri374: Fixed

PR #200, opened 2019-07-28
code + documentation for a standalone Presidio Analyzer installation using Wheel

omri374:
Code and docs for generating a `wheel` file for a standalone analyzer installation.
**For internal purposes only. Will not be merged for now.**


presidio-analyzer/analyzer/nlp_engine/spacy_nlp_engine.py
```diff
@@ -14,6 +15,7 @@ class SpacyNlpEngine(NlpEngine):
 
     def __init__(self):
         logger.info("Loading NLP model...")
+        download('en_core_web_lg')
```
eladiw: does this means that if the file already exists it will be replaced? or we need to delete it from the repo?

presidio-analyzer/VERSION
```diff
@@ -0,0 +1 @@
+0.0.4
```
eladiw: 0.0.1?

presidio-analyzer/analyzer/nlp_engine/spacy_nlp_engine.py
omri374: it only downloads in case the model isn't downloaded already. Otherwise it does nothing

presidio-analyzer/VERSION
omri374: Wanted it to be aligned with the next version of presidio

omri374: Anyway this is just for internal purposes. We're not going to merge this for now.

presidio-analyzer/analyzer/nlp_engine/spacy_nlp_engine.py
itye-msft: This is important to document, as a comment

omri374: @itye-msft added.

presidio-analyzer/VERSION
omri374: @eladiw changed to 0.0.1

PR #205, opened 2019-08-31
fix: Development documentation

fernandoBRS:
'sudo' command added to re2 installation on python section to avoid permission denied for file creation on /usr/local/include/re2


omri374:
Hi Fernando, thanks for the PR! I'm not sure that adding `sudo` in this documentation is the right way to go. How about running this script as administrator?

Issue #192, opened 2019-07-17
Current make file throws errors if container registry is undefined (first build only)


CodeRunRepeat:

omri374:
The make file works even without a defined container registry. During the build process, and error will be written but the build would succeed and the error could be ignored.

PR #226, opened 2019-09-19
Fix Current make file throws errors if container registry is undefined

balteravishay:
Fixing by suppressing the errors in makefile, as they are sometime expected (running locally without registry for instance).

Error text is still output to std out, but not as error.

See example:

docker pull presidio/presidio-golang-deps:'latest'       ||:
Error response from daemon: pull access denied for presidio/presidio-golang-deps, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
docker pull presidio/presidio-python-deps:'latest'       ||:
Error response from daemon: pull access denied for presidio/presidio-python-deps, repository does not exist or may require 'docker login': denied: requested access to the resource is denied

Closes: #192


Makefile
```diff
@@ -28,8 +28,8 @@ $(BINS): vendor
 
 .PHONY: docker-build-deps
 docker-build-deps:
-	-docker pull $(DOCKER_REGISTRY)/$(GOLANG_DEPS):$(PRESIDIO_DEPS_LABEL)
-	-docker pull $(DOCKER_REGISTRY)/$(PYTHON_DEPS):$(PRESIDIO_DEPS_LABEL)
+	-docker pull $(DOCKER_REGISTRY)/$(GOLANG_DEPS):$(PRESIDIO_DEPS_LABEL) ||:
```
eladiw: This looks good the build scenario

Makefile
```diff
@@ -81,7 +81,7 @@ docker-push: $(addsuffix -push,$(IMAGES))
 docker-push-latest-dev: $(addsuffix -push-latest-dev,$(IMAGES))
 
 %-push-latest-dev:
-	docker pull $(DOCKER_REGISTRY)/$*:$(PRESIDIO_LABEL)
+	docker pull $(DOCKER_REGISTRY)/$*:$(PRESIDIO_LABEL) ||:
```
eladiw: but this doesnt, It is not possible for a push scenario without a registry defined

Makefile
```diff
@@ -90,7 +90,7 @@ docker-push-latest-branch: $(addsuffix -push-latest-branch,$(IMAGES))
 
 
 %-push-latest-branch:
-	docker pull $(DOCKER_REGISTRY)/$*:$(PRESIDIO_LABEL)
+	docker pull $(DOCKER_REGISTRY)/$*:$(PRESIDIO_LABEL) ||:
```
eladiw: same

PR #230, opened 2019-10-15
Addsgnric

lordlinus:
Adding SG NRIC/FIN recognizer based on info from https://en.wikipedia.org/wiki/National_Registration_Identity_Card


omri374:
Hi @lordlinus 
Thanks for your contribution!

A few initial comments:
1. Your PR contains multiple files that aren't related to the proposed change. Could you please remove those? You should only have change in 6 files.
2. Please try to avoid using very weak regex like `'(\b[0-9]{7}\b)'` as they have a negative factor on the overall outcome (even if the confidence is very low)
3. Could you share the source of these regexes?

lordlinus:
apologies @omri374 , i will cleanup and resubmit a new pull.

Issue #239, opened 2019-11-14
Custom recognizers won't work


apresmoi:
Hello.

For some reason custom recognizers don't work on my current presidio instance.

I'm following the documentation:

I post into `api/v1/analyzer/recognizers/spaceship`
`
{
	"value": {
		"entity": "SPACESHIP",
		"language": "en", 
		"patterns": [
			{
				"name": "spaceship regex",
				"regex": "\\W*(spaceship)\\W*",
				"score": 0.9
			}
		]
	}
}
`

And then I try to run the analyzer posting into `/api/v1/projects/testing/analyze` in two ways:
`
{
	"text":"They went inside the spaceship", 
	"analyzeTemplate": {
		"allFields": true
	}
}
`
Response:
`
"No results"
`

`
{
	"text":"They went inside the spaceship", 
	"analyzeTemplate": {
		"fields": [
			{ "name": "SPACESHIP" }
                ]
	}
}
`
Response
`
"rpc error: code = Unknown desc = Exception calling application: No matching recognizers were found to serve the request."
`

Any thoughts ?

Regards.

omri374:
Hello @apresmoi, could you please provide some more details on your deployment? Custom recognizers by API require a Kubernetes deployment with Redis (deployed together with Presidio)

cc @eladiw

omri374:
Closing as more info from the issue opener was not received.

navalev:
I assume the services were deployed using the build.sh script. The services in the script were not configured properly and the analyzer did not communicate with the recognizers store.
PR #260 is fixing the build script and all tests cases manined in this issue pass.

kevin-yauris:
> I assume the services were deployed using the build.sh script. The services in the script were not configured properly and the analyzer did not communicate with the recognizers store.
> PR #260 is fixing the build script and all tests cases manined in this issue pass.

hi @navalev and @omri374, is this mean custom recognizers by API can be done with deployment using docker on local? 
I tried to run `echo -n '{"value": {"entity": "ROCKET","language": "en", "patterns": [{"name": "rocket-regex","regex": "\\W*(rocket)\\W*","score": 1}]}}' | http -f POST http://0.0.0.0:8080/api/v1/analyzer/recognizers/rocket`
but the response is 
```
HTTP/1.1 500 Internal Server Error
Content-Length: 0
```

I also try to get all recognizers enable using 
`http http://0.0.0.0:8080/api/v1/analyzer/recognizers` responded with 
```
[...]

PR #240, opened 2019-11-22
Clarified local docker build process and added build.sh script

stevehaigh:
Minor doc clarifications for docker process

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,17 @@
+# Build the images
+
+export DOCKER_REGISTRY=presidio
+export PRESIDIO_LABEL=latest
+make DOCKER_REGISTRY=${DOCKER_REGISTRY} PRESIDIO_LABEL=${PRESIDIO_LABEL} docker-build-deps
+make DOCKER_REGISTRY=${DOCKER_REGISTRY} PRESIDIO_LABEL=${PRESIDIO_LABEL} docker-build
+
+# Run the containers
+
+docker network create mynetwork
+docker run --rm --name redis --network mynetwork -d -p 6379:6379 redis
+docker run --rm --name presidio-analyzer --network mynetwork -d -p 3000:3000 -e GRPC_PORT=3000 ${DOCKER_REGISTRY}/presidio-analyzer:${PRESIDIO_LABEL}
+docker run --rm --name presidio-anonymizer --network mynetwork -d -p 3001:3001 -e GRPC_PORT=3001 ${DOCKER_REGISTRY}/presidio-anonymizer:${PRESIDIO_LABEL}
+docker run --rm --name presidio-recognizers-store --network mynetwork -d -p 3004:3004 -e GRPC_PORT=3004 -e REDIS_URL=redis:6379 ${DOCKER_REGISTRY}/presidio-recognizers-store:${PRESIDIO_LABEL}
+
+sleep 30 # Wait for the analyzer model to load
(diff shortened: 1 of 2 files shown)


docs/install.md
```diff
@@ -24,11 +34,19 @@ sleep 30 # Wait for the analyzer model to load
...
 
-**NOTE: Building the deps images currently takes some time** (~70 minutes, depending on the build machine). We are working on improving the build time through improving the build and providing pre-built dependencies.
+Once the build is complete you can verify the local deployment by running `docker ps`. You should see 4 Presidio containers and one Redis container running with the following names:
+
+```sh
+presidio-api
+presidio-recognizers-store
+presidio-anonymizer
+presidio-analyzer
+redis
```
omri374: Related to #239 
@eladiw Do we have a functional test which checks the recognizer store in a Docker installation?

docs/install.md
```diff
@@ -1,9 +1,19 @@
...
 
 ## The easy way with Docker
 
+You will need to have Docker installed and running, and [make](https://www.gnu.org/software/make/) make installed on your system.
+
+Sync this repo use `make` to build and deploy locally.
+
+For convenience the script `build.sh` at the root of this repo will run the `make` commands for you. If you use the script remember to make it executable by running `chmod +x build.sh` after syncing the code.
+
+**NOTE: Building the deps images currently takes some time** (~70 minutes, depending on the build machine). We are working on improving the build time through improving the build and providing pre-built dependencies.
```
omri374: Should we add an additional note about the errors which get printed out when running this without a container registry?

build.sh
```diff
@@ -0,0 +1,17 @@
+# Build the images
```
omri374: I think it would be good to have a comment saying that this build is for a local Docker build and not for other types of deployments. WDYT?

omri374:
Thanks @stevehaigh! That's really useful.

build.sh
itye-msft: Although this was written in a doc file, it would also be a good idea to have it written in the file as well.

docs/install.md
```diff
@@ -1,9 +1,22 @@
 # Install guide
 
-You can install Presidio as a service in [Kubernetes](https://kubernetes.io/) or use it as a framework
+You can install Presidio locally using [Docker](), as a service in [Kubernetes](https://kubernetes.io/) or use it as a framework in a python application.
```
omri374:
```suggestion
You can install Presidio locally using [Docker](https://www.docker.com/), as a service in [Kubernetes](https://kubernetes.io/) or use it as a framework in a python application.
```

docs/install.md
```diff
@@ -1,9 +1,22 @@
...
 
 ## The easy way with Docker
 
+You will need to have Docker installed and running, and [make](https://www.gnu.org/software/make/) make installed on your system.
```
omri374:
```suggestion
You will need to have Docker installed and running, and [make](https://www.gnu.org/software/make/) installed on your system.
```

docs/install.md
```diff
@@ -1,9 +1,22 @@
...
+
+Sync this repo use `make` to build and deploy locally.
```
omri374: Is this necessary given the new build.sh file?

stevehaigh: Maybe not, I'm happy to remove it you are.

stevehaigh: updated now.

PR #260, opened 2020-01-15
Update build.sh with configuration per service

navalev:
Update build.sh script with relevant configuration per container service.
This PR will resolve issue #239

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -16,9 +16,10 @@ make DOCKER_REGISTRY=${DOCKER_REGISTRY} PRESIDIO_LABEL=${PRESIDIO_LABEL} docker-
 
 docker network create mynetwork
 docker run --rm --name redis --network mynetwork -d -p 6379:6379 redis
-docker run --rm --name presidio-analyzer --network mynetwork -d -p 3000:3000 -e GRPC_PORT=3000 ${DOCKER_REGISTRY}/presidio-analyzer:${PRESIDIO_LABEL}
-docker run --rm --name presidio-anonymizer --network mynetwork -d -p 3001:3001 -e GRPC_PORT=3001 ${DOCKER_REGISTRY}/presidio-anonymizer:${PRESIDIO_LABEL}
-docker run --rm --name presidio-recognizers-store --network mynetwork -d -p 3004:3004 -e GRPC_PORT=3004 -e REDIS_URL=redis:6379 ${DOCKER_REGISTRY}/presidio-recognizers-store:${PRESIDIO_LABEL}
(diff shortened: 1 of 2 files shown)


omri374:
Was the issue the anonymizer-image or something else?

navalev:
> Was the issue the anonymizer-image or something else?
It was the analyzer. The recognozer-store was not configured. This was fixed by setting a value to RECOGNIZERS_STORE_SVC_ADDRESS enviroment var. 
I added anonymizer-image container, but can remove it if it is reduntent for a sample build

omri374:
Yes, the analyzer image just adds complexity here IMHO.

navalev:
> Yes, the analyzer image just adds complexity here IMHO.

removed

build.sh
```diff
@@ -16,9 +16,9 @@ make DOCKER_REGISTRY=${DOCKER_REGISTRY} PRESIDIO_LABEL=${PRESIDIO_LABEL} docker-
 
 docker network create mynetwork
 docker run --rm --name redis --network mynetwork -d -p 6379:6379 redis
-docker run --rm --name presidio-analyzer --network mynetwork -d -p 3000:3000 -e GRPC_PORT=3000 ${DOCKER_REGISTRY}/presidio-analyzer:${PRESIDIO_LABEL}
```
limorl: Shouldn't you use the environment variables instead of actual ports and addresses? (you can use default values if not defined

navalev: I this scenario the build script does not set any environemnt variables other than the docker registry and label. it is used here https://github.com/microsoft/presidio/blob/master/docs/deploy.md 
The ports are the same as in the helm chart values yaml. 
I think adding env variables will add redundent complexity to the sample deployment.

limorl: I was asking since the development.md specifies environment variables. So wither we use them across all solution or we don't. But we need to be consistent. I'm revisiting the development.md now, and I saw the env variables hence my question.If we decide not to use env variables, I will remove that section from the development.md

navalev: in the build script the ports are set as env variables in the docker container (-e GRPC_PORT=3000 for example), so it is the same as setting it when running the processes directly, not as containers. 
Therefor, you should not remove this from the dev md (as it will be used during dev).

limorl: I actually prefer to avoid duplication between the script and the development.md. I would like to remove the command section from the .md and instead add a note to run build.sh.
Everyone can override the ports in their local build as they see fit. 

I'll approve your change and once merged, will submit a change for the development.md for your review

limorl:
LGTM

PR #266, opened 2020-03-02
Multi stage pipeline

balteravishay:
This PR closes issue [1747](https://dev.azure.com/csedevil/Presidio-internal/_workitems/edit/1747)

The following changes are suggested to the CI pipeline:

* Multi stage pipeline using new YAML capabilities in ADO. stages are:
1. Security scan
2. Environment setting - replace triggers with git diff to allow better filtering on deps build.
3. build python deps + service
4. build golang deps + base + services in parallel
5. Run integration tests
6. Publish artifacts

* Merge CI for deps and non-deps builds using code that performs git diff instead of using triggers (this PR makes deps-CI obsolete)

* Improve golang base container structure to use caching better by first copying Gopkg.lock Gopkg.toml, running dep ensure and only then copying the rest of code.

* Publish analzer as pypi wheel package and use wheel when building the container. Currently this PR publishes to Azure Artifacts. To allow devs to keep working locally using docker without having to publish the wheel, there are two dockerfiles for Analyzer, one which is used in CI is using the wheel, and one which can be used by devs is the same as before this PR and built using the local environment.

* Refactor Makefile 
1. Allow separate container build (used in CI to allow parallelism)
2. pull base image before build
3. allow switching between python container build types (local-dev and wheel)

* Move Version file to the root level so it can be used when building both Analyzer and Evaluator

* Cred Scan stage

Note the following build to validate the PR:

* Full build with deps - https://dev.azure.com/csedevil/Presidio/_build/results?buildId=8162&view=results


docs/build_release.md
```diff
@@ -1,46 +1,34 @@
...
+The project currently supports [Azure Pipelines](https://azure.microsoft.com/en-us/services/devops/pipelines/) using YAML pipelines which can be easily imported to any Azure Pipelines instance.
 
 ## Presidio Azure Devops Pipelines
 
-Azure Pipelines [YAML Schema](https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema) allows for code reuse using [templates](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops).<br/>
+Azure Pipelines [templates](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops) allows for code reuse using [YAML Schema](https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema).
 
-***[Presidio Build and Push template](../pipelines/templates/presidio-build-template.yaml)*** - build, test and push presidio images
+***[Presidio CI Pipeline](../pipelines/CI-presidio.yaml)*** - is the pipeline which is used to buil, test and  publish persidio services images.
```
omri374: typo: build instead of buil

docs/build_release.md
```diff
@@ -1,46 +1,34 @@
...
 
-- Push and publish steps only taken when build is not triggered by a PR.
-- Parameters:
-    - registry_parameter - the name of container registry service connection
-    - registry_name_parameter - the full name of the registry login
-    - deps_label_parameter - the base image (deps) label to be used for container build
+***[Presidio Build and Push template](../pipelines/templates/build-test-publish.yaml)*** - is the stages-template which contains the build, test and push logic of presidio. There are four stages for the build:
```
omri374: The difference between push and deploy is a bit confusing (maybe just to me)

docs/build_release.md
```diff
@@ -1,46 +1,34 @@
...
-* ***REGISTRY_NAME*** - the name of the registry
-* ***DEPENDENCY_DEFAULT*** - override dependency label in a Presidio-CI build.
+### Variables used by the pipeline
 
+* ***PYPI_INDEX_URL*** - pypi artifact used to push the wheel.
+* ***REGISTRY*** - service connection to docker registry.
+* ***REGISTRY_NAME*** - full name of container registry (i.e. presidio.azurecr.io).
+* ***DEPENDENCY_DEFAULT*** - presidio release version.
 
 ### import a pipeline to Azure Devops
```
omri374: Import

pipelines/templates/build-golang-template.yaml
```diff
@@ -0,0 +1,54 @@
+# This template builds a golang container and pushes it to a container registry, using presidio Makefile.
+# It is used in CI to build the base container as well as all golang based services.
+# parameters:
+#   registry_service_connection: service connection to docker registry.
```
omri374: is this param relevant when not using Azure?

presidio-analyzer/Dockerfile.local
```diff
@@ -0,0 +1,30 @@
+ARG REGISTRY=presidio.azurecr.io
```
omri374: Worth adding some documentation on the difference between the docker files (in the files themselves)

presidio-analyzer/setup.py
```diff
@@ -15,18 +15,18 @@
...
-        'grpcio==1.21.1',
-        'protobuf==3.8.0',
-        'tldextract==2.2.1',
+        'cython==0.29.10', 
+        'spacy==2.2.3', 
+        'regex==2019.6.8', 
+        'grpcio==1.21.1', 
+        'protobuf==3.8.0', 
+        'tldextract==2.2.1', 
+        'regex==2019.4.14',
```
omri374: regex appears twice

Dockerfile.golang.base
```diff
@@ -3,8 +3,10 @@ ARG PRESIDIO_DEPS_LABEL=latest
...
 FROM ${REGISTRY}/presidio-golang-deps:${PRESIDIO_DEPS_LABEL}
 
+
+COPY Gopkg.lock Gopkg.toml $GOPATH/src/github.com/Microsoft/presidio/
 WORKDIR $GOPATH/src/github.com/Microsoft/presidio
-ADD . $GOPATH/src/github.com/Microsoft/presidio
+RUN dep ensure -vendor-only
 
-RUN dep ensure
+ADD . $GOPATH/src/github.com/Microsoft/presidio
```
eladiw: If I understand correctly.
you divided it just so the layers will be easier for caching?
line 11, is now doing what the old line 7 was doing. right?

so functionality wise there is no difference. right?

Makefile
```diff
@@ -3,11 +3,14 @@ DOCKER_BUILD_FLAGS :=
...
 BINS        = presidio-anonymizer presidio-ocr presidio-anonymizer-image presidio-api presidio-scheduler presidio-datasink presidio-collector presidio-recognizers-store presidio-tester
-IMAGES      = presidio-anonymizer presidio-ocr presidio-anonymizer-image presidio-api presidio-scheduler presidio-datasink presidio-collector presidio-analyzer presidio-recognizers-store presidio-tester
+GOLANG_IMAGES      = presidio-anonymizer presidio-ocr presidio-anonymizer-image presidio-api presidio-scheduler presidio-datasink presidio-collector presidio-recognizers-store presidio-tester
+PYTHON_IMAGES      = presidio-analyzer
+IMAGES      = $(GOLANG_IMAGES)  $(PYTHON_IMAGES)
 GOLANG_DEPS	= presidio-golang-deps
 PYTHON_DEPS	= presidio-python-deps
 GOLANG_BASE	= presidio-golang-base
-
+PIP_EXTRA_INDEX_URL =
```
eladiw: can you document the meaning of those

Makefile
```diff
@@ -3,11 +3,14 @@ DOCKER_BUILD_FLAGS :=
 LDFLAGS            :=
 
 BINS        = presidio-anonymizer presidio-ocr presidio-anonymizer-image presidio-api presidio-scheduler presidio-datasink presidio-collector presidio-recognizers-store presidio-tester
-IMAGES      = presidio-anonymizer presidio-ocr presidio-anonymizer-image presidio-api presidio-scheduler presidio-datasink presidio-collector presidio-analyzer presidio-recognizers-store presidio-tester
+GOLANG_IMAGES      = presidio-anonymizer presidio-ocr presidio-anonymizer-image presidio-api presidio-scheduler presidio-datasink presidio-collector presidio-recognizers-store presidio-tester
```
eladiw: minor. align all as was before '=' on same column

Makefile
```diff
@@ -35,60 +38,79 @@ docker-build-deps:
 
 .PHONY: docker-build-base
 docker-build-base:
-	docker build --build-arg REGISTRY=$(DOCKER_REGISTRY) --build-arg PRESIDIO_DEPS_LABEL=$(PRESIDIO_DEPS_LABEL) -t $(DOCKER_REGISTRY)/$(GOLANG_BASE) -f Dockerfile.golang.base .
+	-docker pull $(DOCKER_REGISTRY)/$(GOLANG_BASE):$(PRESIDIO_BRANCH_LABEL) || echo "\nCould not pull shared base Go image from registry, building locally. If you planned to build locally, the previous error message could be ignored\n"
```
eladiw:
```suggestion
	-docker pull $(DOCKER_REGISTRY)/$(GOLANG_BASE):$(PRESIDIO_BRANCH_LABEL) || echo "\nCould not pull shared base Go image from registry, building locally. If you planned to build locally, the previous error message can be ignored\n"
```

Makefile
```diff
@@ -35,60 +38,79 @@ docker-build-deps:
...
+	docker build --build-arg REGISTRY=$(DOCKER_REGISTRY) --cache-from=$(DOCKER_REGISTRY)/$(GOLANG_BASE):$(PRESIDIO_BRANCH_LABEL) --build-arg PRESIDIO_DEPS_LABEL=$(PRESIDIO_DEPS_LABEL) -t $(DOCKER_REGISTRY)/$(GOLANG_BASE):$(PRESIDIO_LABEL) -f Dockerfile.golang.base .
```
eladiw: 👍

Makefile
```diff
@@ -35,60 +38,79 @@ docker-build-deps:
...
-docker-build: docker-build-base
-docker-build: $(addsuffix -dimage,$(IMAGES))
+docker-build: docker-build-golang
+docker-build: docker-build-python
+
+.PHONY: docker-build-golang
+docker-build-golang: docker-build-golang-base
+docker-build-golang: $(addsuffix -dimage,$(GOLANG_IMAGES))
+
+.PHONY: docker-build-golang-single
```
eladiw: what is this build command. i mean why do you need it if you have the one on line 51

Makefile
```diff
@@ -35,60 +38,79 @@ docker-build-deps:
...
+.PHONY: docker-build-python
+docker-build-python: $(addsuffix -dpypiimage,$(PYTHON_IMAGES))
+
 
 %-dimage:
-	docker build $(DOCKER_BUILD_FLAGS) --build-arg REGISTRY=$(DOCKER_REGISTRY) --build-arg VERSION=$(VERSION) --build-arg PRESIDIO_DEPS_LABEL=$(PRESIDIO_DEPS_LABEL) -t $(DOCKER_REGISTRY)/$*:$(PRESIDIO_LABEL) -f $*/Dockerfile .
+	docker build $(DOCKER_BUILD_FLAGS) --build-arg REGISTRY=$(DOCKER_REGISTRY) --build-arg VERSION=$(VERSION) --build-arg PRESIDIO_DEPS_LABEL=$(PRESIDIO_DEPS_LABEL) --build-arg PRESIDIO_BASE_LABEL=$(PRESIDIO_LABEL) -t $(DOCKER_REGISTRY)/$*:$(PRESIDIO_LABEL) -f $*/Dockerfile .
+
+%-dpypiimage:
+ifndef PIP_EXTRA_INDEX_URL
```
eladiw: Can you elaborate?
e.g. line 72 has wheel ver. line 70 doesnt. on purpose?

VERSION
```diff
@@ -0,0 +1 @@
+0.3.dev
```
eladiw: dev?

pipelines/CI-presidio.yaml
```diff
@@ -1,73 +1,37 @@
+# CI pipeline for presidio. This pipeline triggers the build-test-publish template for every PR
+# and merge to master on https://github.com/microsoft/presidio.
+# Pipeline parameters:
+#   * PYPI_INDEX_URL - pypi artifact used to push the wheel.
```
eladiw: I think the comment should be revised a bit. 'artifact' is not so clear....

pipelines/templates/build-golang-template.yaml
```diff
@@ -0,0 +1,54 @@
...
+steps:
+- bash: |
+    mkdir -p '$(GOBIN)'
+    mkdir -p '$(GOPATH)/pkg'
+    rm -rf '$(MODULEPATH)'
+    mkdir -p '$(MODULEPATH)'
+    shopt -s extglob
+    mv !(gopath) '$(MODULEPATH)' --force
+    echo '##vso[task.prependpath]$(GOBIN)'
+    echo '##vso[task.prependpath]$(GOROOT)/bin'
```
eladiw: If I understand correctly this script setup go on the agent machine? in a bit of manual intervention steps?

Can you name this step / doc. 
e.g. 'shopt -s' what is it for?

pipelines/templates/build-golang-template.yaml
```diff
@@ -0,0 +1,54 @@
...
+    REGISTRY_NAME: ${{ parameters.registry_name }}
+    PRESIDIO_LABEL: ${{parameters.presidio_label}}
+    PRESIDIO_DEPS_LABEL: ${{parameters.deps_label}}
+    BRANCH_LABEL: ${{parameters.presidio_branch_label}}
+    DOCKER_COMMAND: ${{ parameters.make_command_to_run }}
+    GOLANG_IMAGE: ${{ parameters.image_name }}
+  workingDirectory: '$(MODULEPATH)'
+  displayName: 'Build image'
+- bash: | 
+    make DOCKER_REGISTRY=$REGISTRY_NAME PRESIDIO_LABEL=$PRESIDIO_LABEL GOLANG_IMAGE=$GOLANG_IMAGE docker-push-golang-single
```
eladiw:
```suggestion
    make docker-push-golang-single DOCKER_REGISTRY=$REGISTRY_NAME PRESIDIO_LABEL=$PRESIDIO_LABEL GOLANG_IMAGE=$GOLANG_IMAGE
```

why the single command and not the other one?

pipelines/templates/build-python-template.yaml
```diff
@@ -0,0 +1,91 @@
...
+    command: 'login'
+- task: UsePythonVersion@0
+  inputs:
+    versionSpec: '$(PYTHON.VERSION)'
+  displayName: 'Use Python $(PYTHON.VERSION)'
+- script: |
+    python -m pip install --upgrade pip
+    python -m pip install pipenv
+    pipenv --python 3
+    pip install twine
```
eladiw: what is this?

pipelines/templates/build-python-template.yaml
```diff
@@ -0,0 +1,91 @@
...
+  displayName: 'Docker login'
+  inputs:
+    containerRegistry: ${{ parameters.registry_service_principal }}
+    command: 'login'
+- task: UsePythonVersion@0
+  inputs:
+    versionSpec: '$(PYTHON.VERSION)'
+  displayName: 'Use Python $(PYTHON.VERSION)'
+- script: |
+    python -m pip install --upgrade pip
```
eladiw: here you install python 'normally' why go wasn't like that

pipelines/templates/build-python-template.yaml
```diff
@@ -0,0 +1,91 @@
...
+    python -m pip install pipenv
+    pipenv --python 3
+    pip install twine
+  displayName: 'Install python environment'
+  workingDirectory: ${{ parameters.folder_name }}
+- task: TwineAuthenticate@0
+  inputs:
+    artifactFeeds: ${{ parameters.artifact_feed }}
+  displayName: 'Authenticate RC with Azure Artifacts'
```
eladiw: minor, can we have the display name the first item? easier to read

pipelines/templates/build-test-publish.yaml
```diff
@@ -0,0 +1,441 @@
...
+        variables:
+          DEPS_LABEL: $[ dependencies.GetDepsJob.outputs['setEnv.DEPS_LABEL'] ]
+          ${{ if eq(variables['Build.SourceBranchName'], 'master') }}:
+              BUILD_LABEL: 'latest'
+          ${{ if ne(variables['Build.SourceBranchName'], 'master') }}:
+              BUILD_LABEL: '$(Build.SourceBranchName)'
+        strategy:
+          maxParallel: 10
+          matrix:
+            ANNONIMYZER:
```
eladiw:
```suggestion
            ANONYMIZER:
```

presidio-analyzer/Dockerfile.local
eladiw: 👍 
is 'local' == 'standalone'?

Dockerfile.golang.base
balteravishay: yes. functionality wise the "docker build" has the same results.
this way, if you change your code (which happens more frequently than changing gopkg), dep ensure will not run which is the major improvement here.

Makefile
balteravishay: done

Makefile
balteravishay: its more than that. line 70 uses the local docker file. 72 uses Dockerfile which in turn, uses the wheel.
added comment in line.

VERSION
balteravishay: this gets replaced in CI to either non (0.3) or RC (0.3-rc) depending on the branch.

docs/build_release.md
balteravishay: fixed

docs/build_release.md
balteravishay: replaced push with publish. deploy has no reference in this doc as we are currently not automating it so it irrelevant.

pipelines/CI-presidio.yaml
balteravishay: done

pipelines/templates/build-golang-template.yaml
balteravishay: the pipeline is couples to azure and this parameter is the service connection to the publish azure container registry. without this the pipeline can't run

pipelines/templates/build-golang-template.yaml
balteravishay: this is the ugly part of running go<1.11 on an azure pipelines agent as documented here (notice the go-older tab)

https://docs.microsoft.com/en-us/azure/devops/pipelines/ecosystems/go?view=azure-devops&tabs=go-older

notice this step is only done now when running the integ tests (this used to run on the build as well, but there is no need for it there since we're only doing docker builds now).

the reason we need this to run the integ tests is because they themselves to not run from a containers (based from the golang-base). i've already opened a bug for this here: https://dev.azure.com/csedevil/Presidio-internal/_workitems/edit/1804

added better name

balteravishay: you're right, this is not required here and is removed. see my comment to why we need this in the integration tests part:

this is the ugly part of running go<1.11 on an azure pipelines agent as documented here (notice the go-older tab)

https://docs.microsoft.com/en-us/azure/devops/pipelines/ecosystems/go?view=azure-devops&tabs=go-older

the reason we need this to run the integ tests is because they themselves to not run from a containers (based from the golang-base). i've already opened a bug for this here: https://dev.azure.com/csedevil/Presidio-internal/_workitems/edit/1804

pipelines/templates/build-golang-template.yaml
balteravishay: this is not done in parallel. see the pipeline: https://dev.azure.com/csedevil/Presidio/_build/results?buildId=8162&view=results

pipelines/templates/build-python-template.yaml
balteravishay: fixes, see my comment above

pipelines/templates/build-python-template.yaml
balteravishay: added comment on top of file. twine is used to publish the wheel.

presidio-analyzer/Dockerfile.local
balteravishay: local is on a local dev environment. non-local is the one used in CI. added doc in files

pipelines/templates/build-test-publish.yaml
balteravishay: fixed

pipelines/templates/build-python-template.yaml
balteravishay: fixed

PR #277, opened 2020-03-10
Running functional tests in container

balteravishay:
This PR closes the following bug: [1804](https://dev.azure.com/csedevil/Presidio-internal/_workitems/edit/1804)

NOTE THAT the PR is currently open in comparison to feature/pypi-publish and not master!

the following changes are suggested:

* Running golang based  functional tests in a container. the main goal is to avoid having to run "dep ensure" on a local machine (or build agent). in CI this change saved ~4 minutes.

* Adding a dockerfile to functional tests. this dockerfile runs "go test" so it need the entire dev env. output from the test is sent to a mounted volume where it can be read by the host machine.

* Change name of folder tests to functional_tests to align with the way containers are built (name of folder = name of container).

* In Makefile's functional tests, run all containers in parallel and wait only for analyzer (the biggest container). this saves time.

* publish functional tests to ADO


pipelines/templates/build-test-publish.yaml
```diff
@@ -394,8 +387,12 @@ stages:
           inputs:
             targetType: 'inline'
             script: |
-              make DOCKER_REGISTRY=$REGISTRY_NAME PRESIDIO_LABEL=$(Build.BuildID) test-functional-no-build
-            workingDirectory: '$(MODULEPATH)'
+              make DOCKER_REGISTRY=$REGISTRY_NAME PRESIDIO_LABEL=$(Build.BuildID) TEST_IN_CONTAINER=true test-functional-no-build
+        - task: PublishTestResults@2
```
tamirkamara: you can "name" this test results with _testRunTitle_ to separate them from other tests that might run (not sure have those in this project)

balteravishay: done in the container

PR #278, opened 2020-03-10
Add catch for ValueError exception.

darindf:
Got an exception when processing data while using re2. 

```
 File "\site-packages\analyzer\predefined_recognizers\uk_nhs_recognizer.py", line 26, in validate_result
    val = int(c)
ValueError: invalid literal for int() with base 10: 'U'
```

This catches that error as obviously this is not expected, and hence not valid.


balteravishay:
validated by pipeline https://dev.azure.com/csedevil/Presidio/_build/results?buildId=8287&view=results

PR #282, opened 2020-03-12
Fix bug in build deps

balteravishay:
Build python deps has an extra tab and extra tag "s" which fail the build.


ammills01:
@omri374 it looks like the extra 's' on line 355 is still in master. If I'm reading this correctly, it looks like this fix should have been merged into master already.

Issue #297, opened 2020-04-06
Reorganize protobuf files in analyzer


dhpollack:
It would be nice if the protobuf files were in a separate directory from the application files.  This requires a few tweaks to the current code base, but is not very difficult to do.  The general steps are that you need to make the generated protobufs [use relative imports](https://github.com/protocolbuffers/protobuf/issues/1491#issuecomment-438138293) and then change any reference to the protobufs to `import mydir.myproto_pb2`.  It would make the repo a lot cleaner and easier to work with for new people looking to contribute to the project.

omri374:
Thanks @dhpollack
This is in our backlog. If you're interested in proposing a change in a PR, that would be great

dhpollack:
@omri374 do you have a preference on the name of the directory?  I've actually implemented this on a local copy of presidio where I used `protobuf_models` as the directory name.  If that's fine then I can probably create a PR for this relatively quickly.

omri374:
@dhapollack Thanks! `protobul_models` sounds good.

PR #299, opened 2020-04-07
organize protobuf files

dhpollack:
Issue #297 

To make the protobufs use relative imports, I did the following.  If you
regenerate the protobuf files then you will have to run the last command
to get the relative imports to work correctly.

```sh
cd presidio-analyzer/presidio_analyzer
mkdir -p protobuf_models
mv *_pb2*.py protobuf_models/
cd protobuf_models
sed -i -E 's/^import.*_pb2/from . \0/' *.py
```

Signed-off-by: David Pollack <d.pollack@solvemate.com>


eladiw:
thanks @dhpollack we will review this PR soon.
Thanks for your contribution!

presidio-analyzer/presidio_analyzer/protobuf_models/analyze_pb2_grpc.py
```diff
@@ -1,7 +1,7 @@
 # Generated by the gRPC Python protocol compiler plugin. DO NOT EDIT!
 import grpc
 
-import analyze_pb2 as analyze__pb2
+from . import analyze_pb2 as analyze__pb2
```
eladiw, line 4: was this regenerated?

eladiw: just saw you used sed to alter these files

eladiw: why is this a must. I'd prefer to have as little manual intervention as possible

dhpollack: No, I used sed command that's in the comment

dhpollack:
The manual step is not ideal but it's a long standing issue with Google and how they generate the files. You can see the link in the issue that's attached to this PR. Instead of adding this functionality they insist that it's not need despite the hundreds of messages in the issue.

omri374:
@dhpollack If this fix: https://github.com/protocolbuffers/protobuf/issues/1491#issuecomment-542398041 would work, it's easier to maintain and doesn't require some post processing. WDYT?

dhpollack:
> @dhpollack If this fix: [protocolbuffers/protobuf#1491 (comment)](https://github.com/protocolbuffers/protobuf/issues/1491#issuecomment-542398041) would work, it's easier to maintain and doesn't require some post processing. WDYT?

The comment after that one looks more fool-proof.  I'll give that a try.

dhpollack:
It looks like the build is failing due to failure downloading the spacy models, even though all the tests are passing.  It would be nice if you used the small spacy model as a default, so one can test presidio without having to download and build 800MB large model.  Also, the English spacy models aren't actually "requirements" if someone were to use this library on non-English data.

omri374:
We have an issue with the build agent which is currently being fixed (see https://github.com/microsoft/presidio/pull/300)

The reason we're using the big spaCy model is because it has NER capabilities which the small one doesn't, and some of our tests validate the output of the spaCy model. You are on spot regarding other languages. The current deployment is coupled with the English version and supporting deployments of other models requires some changes to the deployment pipeline.

eladiw:
Hi @dhpollack,
We have fixed the build issue. can you fix the merge conflicts and we will merge this

dhpollack:
@eladiw I rebased on master and then fixed the imports in the test.  I think it should work now.

PR #300, opened 2020-04-07
Fix build agents running out of disk space

balteravishay:
This PR fixes ADO build agent running out of physical disc space during PR validation.
the fix is splitting the makefile commands to two jobs:

* local validations - goland build, golang unit tests, python lint and unit tests
* containers build and functional tests on local agent using docker network


pipelines/templates/build-test-publish.yaml
```diff
@@ -197,17 +197,44 @@ stages:
...
+                set -eux  # fail on error
 
-                # remove pipenv environment to save place on agent.
-                # it is not used in the next steps and when skipping this
-                # build will fail due to agent disk capacity.
-                pipenv --rm	
-                cd presidio-analyzer	
-                pipenv --rm	
-                cd ../
+                # Only dep v 0.5.0 works.
```
eladiw, line 221: why only 0.5.0

PR #308, opened 2020-04-20
Fix typo to recognize a key context word

gitcarbs:
otherwise bank and check get concatenated

#309


omri374:
Thanks for your contribution @gitcarbs. We have some internal issues with our CI. We'll fix it soon and be able to merge your fix.

Issue #302, opened 2020-04-09
presidio analyzer - config dir error


dhpollack:
**Describe the bug**
The presidio analyzer should have the default dir for the config to be in the user's home, but instead it creates a folder in the run dir named `~`.

**To Reproduce**
run the presidio analyzer

**Expected behavior**
`~` expands to the user home path

**Additional context**
The cli does not properly [expand the user dir](https://github.com/microsoft/presidio/blob/master/presidio-analyzer/presidio_analyzer/__main__.py#L152).  

**Solution**
`os.path.expanduser`

eladiw:
Hi @dhpollack 
Thanks for opening this issue.

We will look into this issue. If you can contribute a PR it will be appreciated.

PR #296, opened 2020-03-30
Support additional language POC

eladiw:
This PR introduces a new language (this case hebrew) with two additional predefined recognizers.
IL ID and IL domain name.

What was changed:
1) analyzer already supported languages, however, now a recognizer by default supports 'all' languages and recognizers can override this.
2) when choosing a specific language for analyze, you get those language recognizers and also the recognizers which are not language specific.
3) spacy works only with english, still need to do, load other models (if needed) to support more languages. also since we've used the NLP engine for tokenization any language other than english doesnt support context support....

PR #312, opened 2020-04-21
[WIP] analyzer - multiple languages and nlp engines

dhpollack:
Initially this was my attempt to use stanza, which is an nlp engine by
Stanford.  But generally, it's an update to allow for one to add NLP
engines and custom recognizers more easily.  Specifically, I
standardized the format of the recognizers, removed use of global
variables when possible, and removed a lot of hard-coding of defaults.

I am thinking of using presidio for several non-english projects at work
and these are several of the changes that I made.

Below is a list of the changes in list form:

* make spacy and/or stanza optional
  * remove requirement of en_core_web_lg from install
* allow predefined recognizers to take parameters
  * this allows for easily using these as non-english recognizers
* create config files for different NLP engines
* create tests for stanza
* make all spacy and stanza tests optional
* create a Dockerfile for an anaconda-based image
  * pytorch is built with MKL and is much faster on cpu from conda
* completely rewrote the IBAN recognizer
  * the current version only recognizes IBANs if they are the entirety
    of the string.  This version will find IBANs in sentences.
* fixed some tests
* created a `run.sh` file, so just run dockers without rebuilding them

"Breaking" Changes:

* I would like to use [black](https://github.com/psf/black), but it's
  not super friendly with pylint.  My suggestion is to drop pylint and
  use black instead.
* Default spacy model is `en` rather than `en_core_web_lg` and no spacy
  models are downloaded by default.  The idea is to let the user choose
  which models they want.  For non-english users, it saves a lot of time
  at installation because you don't need to install the large spacy
  model that you aren't using.

Relevant Issues:
#302

Signed-off-by: David Pollack <d.pollack@solvemate.com>


omri374:
Hi @dhpollack, thanks for your awesome contribution! We'll look at it and add comments. 
Some initial thoughts:
1. Have you looked at the [presidio-research](https://github.com/microsoft/presidio-research/) repo for evaluating NLP models? We didn't look at stanza but we did compare spaCy to Flair and CRF models. We'd like to make sure that Stanza is a. fast enough and b. accurate enough. Flair, for example, was extremely accurate but very slow, and CRF was very fast but less accurate.
2. Have you considered using [spacy-stanza](https://spacy.io/universe/project/spacy-stanza)?

dhpollack:
> Hi @dhpollack, thanks for your awesome contribution! We'll look at it and add comments.
> Some initial thoughts:
> 
>     1. Have you looked at the [presidio-research](https://github.com/microsoft/presidio-research/) repo for evaluating NLP models? We didn't look at stanza but we did compare spaCy to Flair and CRF models. We'd like to make sure that Stanza is a. fast enough and b. accurate enough. Flair, for example, was extremely accurate but very slow, and CRF was very fast but less accurate.
> 
>     2. Have you considered using [spacy-stanza](https://spacy.io/universe/project/spacy-stanza)?

@omri374 

1.  I looked at it, but it was after I started hacking away at the original library.  I also wanted the anonymization of text more than the analyzer so it was easier to use this library.  Additionally, I was also looking into deploying a service either via k8s or docker eventually.

2.  I hadn't seen this.  This looks like it would be easier to use.  I'll definitely look further into it.

For my use case, speed was less of a priority than accuracy.  Currently I'm only using a single core instance to test things out and it was fast enough for my use case, but I am sure that if one were to use a GPU machine it would be much faster.  My main goal was see how difficult it would be add different models from different libraries to the analyzer (and ultimately the anonymization) and to make things a bit more user-friendly to add non-English pipelines.  A ancillary goal was to play with stanza and see how easy it would be to use as an alternative to spacy.

Feel free to cherrypick certain parts of this PR.  It looks like there is [another PR](#296) which adds another language (hebrew), but I wanted to throw this out there as an alternative way of doing it.  Instead of creating new python classes for each language one can get the same results with something like the following:

```python
il_domain_recognizer = DomainRecognizer(pattern_groups=[("IL Domain()", u"(www.)?[\u05D0-\u05EA]{2,63}.co.il", 0.5)], supported_entity="IL_DOMAIN_NAME", supported_language="he")
```

Then you can even use config files for different languages / recognizers that are based on the predefined recognizers.  

Another option for me would be to could use the presidio-research repo and then reimplement the anonymization part in python.  I've just gotten started with this project so I'll take a look into that as well.

omri374:
One thing we had issues with regarding languages, is the context phrases. Even in patterns like credit cards (which are international), the context words are still language specific. One option would be to create a recognizer per language and the other is to provide a set of context phrases for each language. 

What do you think would be the most practical and user friendly?

dhpollack:
I made the context an input as well. I just didn't use it in the example because the PR with the example used the same context as the default.

dhpollack:
I standardized the format of the predefined recognizers. If you look at the code you should be able to see what's configurable and what's not. But do you mean at run time and loading all the predefined recognizers for each language? I hadn't tackled this problem yet but Off the top of my head I would use a configuration file with a list of recognizers and load those, otherwise lod the defaults.

elyase:
@dhpollack I would be very interested in testing this PR. I installed the python wheel and downloaded the space models but when I try:

```python
engine = AnalyzerEngine(default_language="de")
text = "My name is David and I live in Miami"
response = engine.analyze(correlation_id=0, text = text, entities=[], language='de', all_fields=True, score_threshold=0.5)
```

I am getting a:

```
2020-04-22 20:45:27 ERROR: Failed to get recognizers hash
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/elyase/Downloads/presidio/presidio-analyzer/presidio_analyzer/analyzer_engine.py", line 198, in analyze
    all_fields=all_fields)
  File "/Users/elyase/Downloads/presidio/presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry.py", line 113, in get_recognizers
    "No matching recognizers were found to serve the request.")
ValueError: No matching recognizers were found to serve the request.
```


Are you able to run that example?

dhpollack:
@elyase I've only tested this if you have a nlp engine (i.e. spacy) with a model of the target language.

```python
from presidio_analyzer.recognizer_registry.recognizer_registry import RecognizerRegistry
from presidio_analyzer.nlp_engine.spacy_nlp_engine import SpacyNlpEngine
nlp = SpacyNlpEngine({"de": "de"})
registry.load_predefined_recognizers(["de"], "spacy")
engine = AnalyzerEngine(registry=registry, nlp_engine=nlp, default_language="de")
text = "My name is David and I live in Miami"
response = engine.analyze(correlation_id=0, text = text, entities=[], language='de', all_fields=True, score_threshold=0.5)
```

I did the above and get a different error about not having a hash for the API store.  I think this is because I don't have redis running on my machine.  However, this is a simplified version of what [app.py](https://github.com/dhpollack/presidio/blob/dhp/allow_multiple_langs/presidio-analyzer/presidio_analyzer/app.py) does.  I've never tried this outside of using app.py so I'm not exactly sure how it should work.

elyase:
@dhpollack that works. The hash error is fortunately just logging output. Thanks a lot!

omri374:
We have a task on ignoring this error when running the analyzer in a standalone way

presidio-analyzer/tests/test_stanza_recognizer.py
```diff
@@ -0,0 +1,198 @@
...
+nlp_is_available = "stanza" in NLP_ENGINES and "stanza" in NLP_RECOGNIZERS
+if nlp_is_available:
+    nlp_engine = NLP_ENGINES["stanza"]()
+    test_recognizer = NLP_RECOGNIZERS["stanza"]()
+entities = ["PERSON", "DATE_TIME"]
+
+
+class TestStanzaRecognizer(TestCase):
+
+    # Test Name Entity
```
omri374, line 20: This can be removed (Should have been removed from the Spacy test too)

presidio-analyzer/tests/test_spacy_recognizer.py
```diff
@@ -23,6 +27,8 @@ class TestSpacyRecognizer(TestCase):
     #     assert len(results) == 1
     #     assert_result(results[0], entity[0], NER_STRENGTH)
 
+    @pytest.mark.skipif(not nlp_is_available,
```
omri374, line 30: How do we verify that an installation is successful? Usually unit tests fail if something is missing, however in this case we would skip it completely.

presidio-analyzer/tests/test_spacy_recognizer.py
```diff
@@ -1,13 +1,17 @@
...
 
-from presidio_analyzer.predefined_recognizers import SpacyRecognizer
+from presidio_analyzer.nlp_engine import NLP_ENGINES
+from presidio_analyzer.predefined_recognizers import NLP_RECOGNIZERS
 from presidio_analyzer.entity_recognizer import EntityRecognizer
 
 NER_STRENGTH = 0.85
-nlp_engine = TESTS_NLP_ENGINE
-spacy_recognizer = SpacyRecognizer()
+nlp_is_available = "spacy" in NLP_ENGINES and "spacy" in NLP_RECOGNIZERS
```
omri374, line 11: should be `NLP_ENGINES` or `TESTS_NLP_ENGINE`?

presidio-analyzer/requirements.txt
```diff
@@ -0,0 +1,35 @@
+argcomplete==1.11.1
```
omri374, line 1: We'd rather not have a Pipfile + requirements.txt. We moved away from requirements.txt because the installation here is challenging enough even without dependency conflicts

presidio-analyzer/presidio_analyzer/recognizer_result.py
```diff
@@ -37,7 +37,7 @@ def __str__(self):
                                   self.end,
                                   self.score)
 
-    def __repr(self):
+    def __repr__(self):
```
omri374, line 40: Good catch :)

presidio-analyzer/presidio_analyzer/predefined_recognizers/us_itin_recognizer.py
```diff
@@ -1,24 +1,45 @@
...
+            0.3,
+        ),
+        (
+            "Itin (medium)",
+            r"\b(9\d{2})[- ]{1}((7[0-9]{1}|8[0-8]{1})|(9[0-2]{1})|(9[4-9]{1}))[- ]{1}(\d{4})\b",  # noqa: E501
+            0.5,
+        ),
+    ]
+
+    CONTEXT = ["individual", "taxpayer", "itin", "tax", "payer", "taxid", "tin"]
```
omri374, line 28: General remark about context: It is language dependent. Maybe it would be better to define it as a dictionary with language as key:
```
CONTEXT = {"en":["individual", "taxpayer", "itin", "tax", "payer", "taxid", "tin"],
                      "de":[]}
```

Then, during analysis, the requested language in the request would be used to fetch the relevant context. This allows us to use the same recognizer for multiple languages.

presidio-analyzer/Dockerfile.local.conda
```diff
@@ -0,0 +1,54 @@
+FROM python:3.7-slim
```
omri374, line 1: Is this docker file used for development? How would a new user use it? Why not use Pipenv locally?

presidio-analyzer/requirements.txt
dhpollack: I can remove this.  I am using anaconda for testing because I've read in the past that the pytorch wheels don't have as many optimizations as the pytorch on anaconda.  Also pipenv doesn't play very nicely with anaconda, so I duplicated what was in the `Pipfile.lock` with a simple jq script.  But I can just have all this in my own version of the repo.

presidio-analyzer/presidio_analyzer/predefined_recognizers/us_itin_recognizer.py
dhpollack: I think you would need to redesign the whole system of recognizers to allow `supported_language` to be a set or list.  Right now, one needs to create a new instance of the recognizer for each language with the correct `supported_language`.  Also, I think maintaining a list of contexts in different languages would get unwieldy very quickly.  It is not difficult to make the change you are suggesting once, but I think you start to go down a path that will make maintenance a lot more difficult (i.e. adding a new recognizer would require adding context for every language and also making sure every recognizer had context for every supported language).

presidio-analyzer/Dockerfile.local.conda
dhpollack: I can also remove this file (same reasoning as `requirements.txt`).

But another reason that I created this file is that it's a lot faster to iterate when testing.  Your current `Dockerfile.local` reinstalls all the requirements any time there is any change to the code.  This takes relatively long especially when making changes that don't affect any of the requirements.  So I added a step that installs from the `Pipfile.lock` (via jq and requirements.txt, super hacky to use conda) and then installs models, then adds the code to the docker.  The important part is adding the code after installing the requirements and models, so the diffs between the images are super small.

Lastly, I should mention that in quarantine, I am basically running internet through my cell phone, so it's slow and I have a limited amount of data.  So I'm much more conscious of large downloads than I normally would be.  This is probably a big reason why I am so against making the `en_core_web_lg` model a requirement in the `Pipfile`.

presidio-analyzer/tests/test_spacy_recognizer.py
dhpollack: The idea behind this is that while using spacy and stanza, you only need one or the other.  But I should make the spacy tests required, especially since it makes more sense to use `spacy-stanza`.

However, I do think that requiring a large spacy model for testing is a fundamental design flaw in the tests.  It's mixing the performance/accuracy of the spacy models with the testing of the functionality of the library.  The best solution would be to mock spacy and just create the entities you want for each test ([SO answer from a spacy developer on this topic](https://stackoverflow.com/a/56733926)).  The second best solution would be to have all the tests pass with a minimal spacy model (i.e. `en`).  Then separately have scripts to benchmark how well you want the whole library to function when using the best models.  

No hard feelings if you disagree with me here.  I do understand the idea that having the best model out of the box can create a more desirable experience for users.  (see note below about my internet situation as to why I'm against making the large spacy model a requirement)

presidio-analyzer/Dockerfile.local.conda
dhpollack: One more point, I found that having the `Dockerfile.python.deps` in the base folder and `Dockerfile` and `Dockerfile` in `presidio-analyzer` confusing.  Since there is only one python service in this library, I wanted to create a single Dockerfile that I could rebuild to test my changes within the service.  In the long term, you might want to consider making each service a separate repo and then using git submodules.  The python analyzer in particular seems like a standalone product that can be useful outside of the service as a whole.

dhpollack:
@omri374 What are your thoughts on dropping pylint in favor of black?  It's also possible to continue using pylint, but there are a few things that black does that pylint complains about.  I guess we could just ignore those in the pylintrc file.  I started doing a bit of programming in Rust and C++ and found standardized code formatters remove a lot of cognitive energy that went into silly things like "should I put these parameters each on their own line, all on one line, two per line,...".  I love it especially on other OSS projects that I've worked on.

presidio-analyzer/tests/conftest.py
```diff
@@ -0,0 +1,51 @@
...
+    skip_slow = pytest.mark.skip(reason="need --runslow option to run")
+    for item in items:
+        if "slow" in item.keywords:
+            item.add_marker(skip_slow)
+
+@pytest.fixture(scope="session")
+def nlp_engines(request):
+    available_engines = {}
+    for name, engine_cls in NLP_ENGINES.items():
+        print(f"======================================\n\n{name} | {engine_cls}")
```
dhpollack, line 31: Note to self: remove this.

presidio-analyzer/tests/test_analyzer_engine.py
```diff
@@ -9,12 +8,12 @@
 from presidio_analyzer.protobuf_models.analyze_pb2 import AnalyzeRequest, \
     RecognizersAllRequest
 from presidio_analyzer.entity_recognizer import EntityRecognizer
-from presidio_analyzer.nlp_engine import NlpArtifacts
+from presidio_analyzer.nlp_engine import NlpArtifacts, NLP_ENGINES
```
dhpollack, line 11: I can remove the `NLP_ENGINES` import now that I am use the fixture.

dhpollack:
Ok, now 'm using pytest marks to mark the tests that test differently with the big model vs the small one.  Also the test times on my machine are 1.8s vs 6.3s vs 22s (spacy, spacy + stanza, spacy + stanza + en_core_web_lg).  There is currently only 1 test that gets different results when using `en_core_web_lg` vs `en`.  I am using `pytest --runslow` to add the large model tests rather than skipping them.

presidio-analyzer/tests/test_spacy_recognizer.py
omri374: I completely agree that a mock (or at least a small model) should be the one used for testing. I'll discuss this with the team as there are some implications (for example, we have test that verifies that the code on the demo website is analyzed correctly. This wouldn't work with a mock, but could be turned into an integration test). There are some other implications on the way we do CI. @balteravishay what are you thoughts on this?

presidio-analyzer/requirements.txt
omri374: I think it would be best. Either we move completely to Conda or we stay with Pipenv. Having both is confusing and difficult to maintain. Honestly Pipenv isn't a perfect tool (especially since it's not actively maintained anymore) but we have a setup which is working here and it would require some development to replace it with Conda.

presidio-analyzer/presidio_analyzer/predefined_recognizers/us_itin_recognizer.py
omri374: In your opinion, what is the alternative? context terms are always language dependent. Either we have a recognizer per language (e.g. EnCreditCardRecognizer) or we allow a recognizer to accept context in multiple languages. Is there any additional alternative?

presidio-analyzer/Dockerfile.local.conda
omri374: On your second point, I agree it's confusing to have it in the root of the repo, but the reason it's divided into two is to answer your first point. you have a base docker image which you install only once, and then you use the analyzer Dockerfile to build your code only, without having to reinstall dependencies.

omri374:
@dhpollack regarding Black, I'm not familiar enough with it. We'll add it to the backlog to investigate. Thanks for the suggestion!

presidio-analyzer/Pipfile
```diff
@@ -5,8 +5,7 @@ name = "pypi"
 
 [packages]
 cython = "*"
-spacy = "==2.2.3"
-en_core_web_lg = {file = "https://github.com/explosion/spacy-models/releases/download/en_core_web_lg-2.2.5/en_core_web_lg-2.2.5.tar.gz"}
+spacy = ">=2.2.4"
```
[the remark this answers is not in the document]
dhpollack: reverted back to `==2.2.4`

presidio-analyzer/tests/test_spacy_recognizer.py
dhpollack: I actually changed this to default to the big model but now you can use the small model with `pytest --runfast`.

presidio-analyzer/tests/test_spacy_recognizer.py
dhpollack: I refactored everything to use pytest fixtures.  You can see `tests/conftest.py` to see what I did.  There is some logic to load the small spacy model if `--runfast` is used.

presidio-analyzer/requirements.txt
dhpollack: I deleted the `requirements.txt` file

presidio-analyzer/Dockerfile.local.conda
dhpollack: I deleted this file.

presidio-analyzer/presidio_analyzer/predefined_recognizers/us_itin_recognizer.py
dhpollack: So I would create a language specific recognizer using something like the following:

```python
il_domain_recognizer = DomainRecognizer(
    pattern_groups=[("IL Domain()", u"(www.)?[\u05D0-\u05EA]{2,63}.co.il", 0.5)],
    context=["תחום", "ip"],  # this is google translated
    supported_entity="IL_DOMAIN_NAME",
    supported_language="he"
)
```

omri374: Overall it sounds good, however this would require the user to instantiate the recognizer with all parameters, as opposed to having some of them predefined within the class itself. WDYT?

presidio-analyzer/tests/test_spacy_recognizer.py
omri374: Cool, thanks. 
Are you considering to implement the spacy-stanza solution?

dhpollack:
@omri374 do you want to break out these changes separately yourself?  If you cherry-pick certain changes but not all then it gets harder to merge / rebase into master.  I maintain a separate branch internally, but I can keep this branch roughly up-to-date with that one.  

Also I think you can run your CI pipeline and it should work.

omri374:
@dhpollack I think we should consider spacy-stanza first. This is the main blocker for accepting this PR. Breaking it into sub PRs or cherry-picking might require some manual work but it would easier to review. We need another reviewer to merge this. 
Did you get a chance to try out spacy-stanza? if this is something you can introduce into this PR, I would be happy to work on breaking it into smaller PRs. WDYT?

dhpollack:
Spacy-stanza is in there. I made a few significant changes since the original PR. Spacy-stanza being one and the other being pytest style tests. But it might be good to get a fresh set of eyes on it and see what they think.

presidio-analyzer/presidio_analyzer/app.py
```diff
@@ -1,27 +1,25 @@
...
-
 from analyzer_engine import AnalyzerEngine # noqa
 from recognizer_registry.recognizer_registry import RecognizerRegistry # noqa
-from nlp_engine.spacy_nlp_engine import SpacyNlpEngine # noqa
+from nlp_engine import NLP_ENGINES # noqa
 from presidio_logger import PresidioLogger # noqa
+from protobuf_models import analyze_pb2, analyze_pb2_grpc
 
-logging.getLogger().setLevel("INFO")
+logging.getLogger().setLevel(os.environ.get("LOGLEVEL", "INFO"))
```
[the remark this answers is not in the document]
dhpollack: Are you referring to the fact that I'm not using the `PresidioLogger` class?  I can use that.  

But the code in master is actually far less efficient than this.  You are creating an unnececessary temporary variable and doing unnecessary checks as assignments when you could just pass in the level as a string.  Also it looks like the `PresidioLogger` already checks the `LOG_LEVEL` environmental variable, so I'm not sure any of that code is necessary.

presidio-analyzer/presidio_analyzer/predefined_recognizers/sg_fin_recognizer.py
```diff
@@ -1,23 +1,35 @@
...
     Recognizes SG FIN/NRIC number using regex
     """
 
-    def __init__(self):
-        patterns = [Pattern('Nric (weak) ', WEAK_REGEX, 0.3),
-                    Pattern('Nric (medium) ', MEDIUM_REGEX, 0.5), ]
-        super().__init__(supported_entity="SG_NRIC_FIN", patterns=patterns,
-                         context=CONTEXT)
+    PATTERN_GROUPS = [
+        ("Nric (weak)", r"(?i)(\b[A-Z][0-9]{7}[A-Z]\b)", 0.3),
```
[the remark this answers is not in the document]
dhpollack: I can do that.

presidio-analyzer/presidio_analyzer/predefined_recognizers/crypto_recognizer.py
```diff
@@ -1,35 +1,49 @@
...
     """
     Recognizes common crypto account numbers using regex + checksum
     """
 
-    def __init__(self):
-        patterns = [Pattern('Crypto (Medium)', REGEX, 0.5)]
-        super().__init__(supported_entity="CRYPTO", patterns=patterns,
-                         context=CONTEXT)
+    PATTERN_GROUPS = [
+        ("Crypto (Medium)", r"\b[13][a-km-zA-HJ-NP-Z1-9]{26,33}\b", 0.5),
```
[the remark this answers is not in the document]
dhpollack: I use it in the [`__init__`](https://github.com/dhpollack/presidio/blob/dhp/allow_multiple_langs/presidio-analyzer/presidio_analyzer/nlp_engine/__init__.py) of the nlp_engine, so that we know what engines are available.  It also makes it easier to add new engines if someone wanted to do that by having a consistent interface.

presidio-analyzer/presidio_analyzer/nlp_engine/spacy_nlp_engine.py
```diff
@@ -13,14 +13,20 @@ class SpacyNlpEngine(NlpEngine):
         The SpacyNlpEngine uses SpaCy as its NLP module
     """
 
-    def __init__(self):
-        logger.info("Loading NLP model: spaCy en_core_web_lg")
+    engine_name = "spacy"
+    is_available = True if spacy else False
```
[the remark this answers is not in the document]
dhpollack: I use it in the [`__init__`](https://github.com/dhpollack/presidio/blob/dhp/allow_multiple_langs/presidio-analyzer/presidio_analyzer/nlp_engine/__init__.py) of the nlp_engine, so that we know what engines are available.  It also makes it easier to add new engines if someone wanted to do that by having a consistent interface.

presidio-analyzer/setup.cfg
```diff
@@ -3,4 +3,6 @@
 # 3. If your code will not run unchanged on both Python 2 and 3, you will
 # need to generate separate wheels for each Python version that you
 # support.
-universal=1
\ No newline at end of file
+universal=1
+[flake8]
+max-line-length = 88
```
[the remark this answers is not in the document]
dhpollack: I can change this to 89.  The [default for black](https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length) is 88, which is why I set it to 88.

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry.py
```diff
@@ -41,22 +41,26 @@ def __init__(self, recognizer_store_api=RecognizerStoreApi(),
...
-            UsBankRecognizer(), UsLicenseRecognizer(),
-            UsItinRecognizer(), UsPassportRecognizer(),
-            UsPhoneRecognizer(), UsSsnRecognizer(),
-            SpacyRecognizer(), SgFinRecognizer()])
+
+        NlpRecognizer = NLP_RECOGNIZERS.get(nlp_engine, SpacyRecognizer)
+        recognizers_map = {"en": [UsBankRecognizer, UsLicenseRecognizer, UsItinRecognizer,
+                UsPassportRecognizer, UsPhoneRecognizer, UsSsnRecognizer, NhsRecognizer,
+                SgFinRecognizer],
+            "ALL": [CreditCardRecognizer, CryptoRecognizer, DomainRecognizer, EmailRecognizer,
```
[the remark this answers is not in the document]
dhpollack: Yea, it doesn't work as well in other languages, but it still does work to a certain extent.  I feel like anyone that wants to use this for other languages to it's full extent is going to have to go beyond the defaults.  Also, I find this solution better than just not having universal recognizers or not having some sort of clear way to signal to someone hacking the library how they could add a universal recognizer.

presidio-analyzer/presidio_analyzer/predefined_recognizers/sg_fin_recognizer.py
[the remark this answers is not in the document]
dhpollack: done

presidio-analyzer/Pipfile
```diff
@@ -5,8 +5,7 @@ name = "pypi"
 
 [packages]
 cython = "*"
-spacy = "==2.2.3"
-en_core_web_lg = {file = "https://github.com/explosion/spacy-models/releases/download/en_core_web_lg-2.2.5/en_core_web_lg-2.2.5.tar.gz"}
```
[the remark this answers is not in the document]
dhpollack: I added this to the deployment docs.

presidio-analyzer/presidio_analyzer/app.py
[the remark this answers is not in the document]
dhpollack: now I've change it to initialize the `PresidioLogger`

omri374:
@dhpollack  thanks for the updates. If you prefer that we go over this together over a meeting, feel free to reach out: presidio@microsoft.com

omri374:
@dhpollack CI fails on linting. Could you please verify that pylint and flake 8 pass successfully?

dhpollack:
> @dhpollack CI fails on linting. Could you please verify that pylint and flake 8 pass successfully?

@omri374 done

presidio-analyzer/presidio_analyzer/pattern_recognizer.py
```diff
@@ -110,21 +110,23 @@ def build_regex_explanation(
...
         """
+        flags = flags if flags else re.DOTALL | re.MULTILINE
         results = []
         for pattern in self.patterns:
             match_start_time = datetime.datetime.now()
             matches = re.finditer(
                 pattern.regex,
                 text,
-                flags=re.IGNORECASE | re.DOTALL | re.MULTILINE)
+                flags=flags)
```
omri374, line 129: Why remove the IGNORECASE flag?

dhpollack: It prevents one from making a case sensitive regex. If you want a case insensitive regex you can add the modifier "(?i)", so you can now do either case sensitive or case insensitive whereas before you could only do case sensitive.

omri374: Case insensitive regex would theoretically improve recall (and might reduce precision). To me it sounds like it should be the default. I'll consult with the team. Do you think each regex should explicitly add (?i) to become case insensitive?

dhpollack: yes, I think if you want a case-insensitive regex then you should write the regex to be case insensitive.  This was actually a pain point for me, because it was buried deep within the code, so I was trying to write a case-sensitive regex and the case sensitivity was automatically ignored.

omri374:
@dhpollack there's still a lint error:
`presidio_analyzer/analyzer_engine.py:13:0: E0012: Bad option value 'import-outside-toplevel' (bad-option-value)`
I think (but not sure) this is because the pylint version we are using does not support this new `import-outside-toplevel` option. If that's the case, we need to update pylint on the Pipfile. 
Another option would be to add 
```
[TYPECHECK]
disable=bad-option-value
```

dhpollack:
> @dhpollack there's still a lint error:
> `presidio_analyzer/analyzer_engine.py:13:0: E0012: Bad option value 'import-outside-toplevel' (bad-option-value)`
> I think (but not sure) this is because the pylint version we are using does not support this new `import-outside-toplevel` option. If that's the case, we need to update pylint on the Pipfile.
> Another option would be to add
> 
> ```
> [TYPECHECK]
> disable=bad-option-value
> ```

Ok, I moved the imports to the top-level and removed pylint disable flag.

omri374:
Hi @dhpollack, sorry for the delay, we're now back into looking into this.
I'd like to merge this PR this week, so I'll leave some comments and once these are fixed we can go forward with the merge.

presidio-analyzer/presidio_analyzer/app.py
```diff
@@ -0,0 +1,196 @@
...
+        super(PresidioCLIHelp, self).__init__(
+            cli_ctx=cli_ctx, privacy_statement="", welcome_message=WELCOME_MESSAGE
+        )
+
+
+def serve_command_handler(
+    enable_trace_pii,
+    env_grpc_port=False,
+    grpc_port=3000,
+    nlp_conf_path="conf/default.yaml",
```
omri374, line 65: Deciding on the model/framework to use should be better documented. Adding details here as docstring as well as on the development.md file on how to set up different spacy and stanza models would be really helpful.

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry.py
```diff
@@ -41,25 +52,48 @@ def __init__(self, recognizer_store_api=RecognizerStoreApi(),
...
+                IbanRecognizer,
+                IpRecognizer,
+                NlpRecognizer,
+            ],
+        }
+        for lang in languages:
+            lang_recognizers = [rc() for rc in recognizers_map.get(lang, [])]
+            self.recognizers.extend(lang_recognizers)
+            all_recognizers = [
+                rc(supported_language=lang) for rc in recognizers_map.get("ALL", [])
```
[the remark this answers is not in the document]
dhpollack: `languages` is a list of languages for which you have a NLP engine available.  So if I load a spacy `en` and spacy `fr` engine, then I will load everything in `"en"` and `"ALL"` for english and `"ALL"` for french.  The previous method won't load other recognizers for other languages, even though the regexes are generally language agnostic.  So you wouldn't get an IP recognizer in French, but you would get this recognizer in English.  I found it a bit confusing why that was happening when I had an NLP engine that wasn't English.

* Yes, the context is still English, but the regex scores are generally strong enough that these recognizers still fire.  Personally, I think it's more intuitive that the IP recognizer finds IP addresses in all languages.  
* It shows anyone who wants to hack the code how they add a language specific recognizer or a language agnostic recognizer

presidio-analyzer/tests/conftest.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--runfast", action="store_true", default=False, help="run fast tests"
+    )
+
+
+def pytest_configure(config):
+    config.addinivalue_line("markers", "slow: mark test as slow to run")
```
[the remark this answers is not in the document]
dhpollack: I think that I need to change the documentation here.  The default is to run the full model.  I changed the flag to `--runfast` so that it uses the small model and skips one test if you add that flag, but if there is no flag then it uses the big model and runs everything.

presidio-analyzer/presidio_analyzer/app.py
dhpollack: Yea, this is a great idea.  I'll work on that.

presidio-analyzer/conf/spacy.yaml
```diff
@@ -0,0 +1,5 @@
+nlp_engine_name: spacy
+models:
+  -
+    lang: en
+    name: en
```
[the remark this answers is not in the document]
dhpollack: I could do that.  I usually just use the short name, but this also makes it more explicit what is the language code and what is the model.  I also don't love my own variable naming scheme here.  I feel like it should be `lang_code` and `model_name`.  Let me look into that.

presidio-analyzer/presidio_analyzer/nlp_engine/stanza_nlp_engine.py
```diff
@@ -0,0 +1,37 @@
...
+
+    engine_name = "stanza"
+    is_available = bool(stanza)
+
+    def __init__(self, models=None):
+        if not models:
+            models = {"en": "en"}
+        logger.debug(f"Loading NLP models: {models.values()}")
+
+        self.nlp = {
```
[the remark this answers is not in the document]
dhpollack: The documentation is a good idea.  I'll look into the try-catch.  I feel like you get a pretty good error message from stanza (and spacy as well) for not having a model downloaded, but having an explicit error message is generally nicer.

presidio-analyzer/tests/conftest.py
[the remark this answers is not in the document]
omri374: I see now

presidio-analyzer/presidio_analyzer/nlp_engine/stanza_nlp_engine.py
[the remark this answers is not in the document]
omri374: I agree, the error message is good enough, but it'd be nice to wrap it in our own explanation.

presidio-analyzer/conf/spacy.yaml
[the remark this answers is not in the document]
omri374: lang_code and model_name sound good

presidio-analyzer/presidio_analyzer/app.py
omri374: Thanks!

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry.py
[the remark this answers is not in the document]
omri374: OK. Let's keep it as is and we'll have some discussions around multi-language recognizers (multiple instances vs. language agnostic ones). Would that work?

dhpollack:
@omri374 I made the changes based on the review.  I am getting an branch conflict error, but it might just be an error with the github frontend and my browser cache.  It should have been fixed with [this commit](https://github.com/dhpollack/presidio/commit/4200436df3e24cf81277bd407f8c64590568487f).

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry.py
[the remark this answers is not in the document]
dhpollack: done

omri374:
@balteravishay @eladiw could you please review the non-python and non-md files and add your comments?

pipelines/templates/build-python-template.yaml
```diff
@@ -46,6 +46,12 @@ steps:
   displayName: 'Install python dependencies'
   inputs:
     targetType: 'inline'
+    script: |
```
balteravishay, line 49: why are there two "script" elements to the task?

dhpollack: good catch, I think that this occurred during one of the rebasings from master.  I created a commit to use the pipeline build from master with the additional step of downloading the large spacy model.

omri374:
Merged! Thanks a lot @dhpollack !

Issue #294, opened 2020-03-26
Inconsistent anonymize results between Stream job and API calls


ammills01:
**Describe the bug**
When text is passed in that has no matching items, the API calls will return your original text. However, when you have a Stream job hooked up it only sends results to the sink if they have matching entities.

**To Reproduce**
1. Call the anonymize API endpoint with text like "It is a great afternoon" and any template
2. Observe the result, you should get back { "text": "It is a great afternoon" }
3. Now, publish an event to Kafka with the same text (I have not tested other job types)
4. Start a Streams job with a Kafka source and whatever sink you desire
5. Notice there is no result sent to your sink

**Expected behavior**
We should always get the text back, even if there are no items to anonymize.

**Additional context**
I have not tried to debug locally but looking at the code, this seems to be caused by the if statement in [./presidio-collector/cmd/presidio-collector/processor/processor.go](https://github.com/microsoft/presidio/blob/master/presidio-collector/cmd/presidio-collector/processor/processor.go), method ReceiveEventsFromStream. As far as I can tell nothing is blowing up, despite there being the possibility of a null reference exception checking analyzerResponse.AnalyzeResults.

`
		analyzerResponse, err := services.AnalyzeItem(ctx, text, streamRequest.AnalyzeTemplate)
		if err != nil {
			err = fmt.Errorf("error analyzing message: %s, error: %q", text, err.Error())
			return err
		}

		if len(analyzerResponse.AnalyzeResults) > 0 {
			anonymizerResult, err := services.AnonymizeItem(ctx, analyzerResponse.AnalyzeResults, text, streamRequest.AnonymizeTemplate)

			if err != nil {
				err = fmt.Errorf("error anonymizing item: %s/%s, error: %q", partition, sequence, err.Error())
				return err
			}

			err = services.SendResultToDatasink(ctx, analyzerResponse.AnalyzeResults, anonymizerResult, fmt.Sprintf("%s/%s", partition, sequence))
			if err != nil {
				err = fmt.Errorf("error sending message to datasink: %s/%s, error: %q", partition, sequence, err.Error())
				return err
			}
			log.Debug("%d results were sent to the datasink successfully", len(analyzerResponse.AnalyzeResults))

		}
		return nil
	})
`

As you can see, if there are no results in the analyzerResponse it skips the AnonymizerItem and SendResultToDatasink functions.

[...]

PR #322, opened 2020-05-04
Fix for bug #294 - Inconsistent anonymizer results

ammills01:
Remove the length check from the analyzer results so we always publish results to the provided sink.

To test, I setup a Streams job with a Kafka source and sink. I then published 10k records to the source topic and confirmed all 10k made it to the sink. The messages were a mix of items with and without matching entities. I verified there were no errors in the job or analyzer logs.


omri374:
Thanks for the contribution @ammills01

PR #338, opened 2020-06-04
Changed pipenv install to pipenv sync

omri374:
1. Attempt to fix failing build on Azure Pipelines due to timeouts on Pipfile lock. 
`pipenv sync` does not re-lock the Pipfile.
2. Make the pipeline script identical to the Presidio Python base Dockerfile


omri374:
Solved by #335

PR #345, opened 2020-07-07
Added Indian PII recognizers

Nilanjan96:
Added Indian Aadhar Card, PAN Card, Driving License Number, Passport, Voter Card, Phone Number recognizers

### DIFF
--- a/presidio-analyzer/presidio_analyzer/predefined_recognizers/__init__.py
+++ b/presidio-analyzer/presidio_analyzer/predefined_recognizers/__init__.py
@@ -13,12 +13,24 @@ from .us_itin_recognizer import UsItinRecognizer
 from .us_passport_recognizer import UsPassportRecognizer
 from .us_phone_recognizer import UsPhoneRecognizer
 from .us_ssn_recognizer import UsSsnRecognizer
+from .ind_driver_license_recognizer import INDLicenseRecognizer
+from .ind_phone_recognizer import INDPhoneRecognizer
+from .ind_aadhar_recognizer import INDAadharRecognizer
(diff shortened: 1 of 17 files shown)


omri374:
Hi @Nilanjan96, thanks for your contribution!

A few initial comments:
1. Have you ran any verification on these recognizers to see their precision/recall values? If any of these have unusually high false positives or false negative values? A combination of ambiguous regex (e.g. in the extreme case, a regex checking that there are 5 to 7 digits) and ambiguous context values (e.g. "number","card") are likely to introduce many false positives.
2. I think it would be best to put all the new recognizers in predefined_recognizers/india or something like that, as we are starting to have a large number of recognizers.
3. Could you verify if any of these have an ability to add checksum/other forms of validation? This could really help to increase the confidence of identification.

omri374:
Also, please run pytest, flake and pylint, as there are some syntax errors. See reference here: https://github.com/microsoft/presidio/blob/master/docs/development.md#dev-python

Nilanjan96:
Sure thing. I am sorry for being late. I will look into this now.

Nilanjan96:
As for the part 1, I did run this test on several hundreds of fake PII data. I did have some problems because US DRIVER LICENSE was marking every 14 letter word as US DRIVER LICENSE as well as a lot of other stuff. I will check the common words like "card" and "number".

omri374:
US_DRIVER_LICENSE is a weak regex, but it usually emits confidence = 0.01. A threshold on confidence would ignore these findings

omri374:
Closing as there's no response from the author

PR #347, opened 2020-07-22
Fix documentation links

diwu1989:
These links were broken because they were relative to the current docs directory vs absolute from the base of the project folder.


omri374:
Thanks!

PR #348, opened 2020-07-23
Add ABA routing number recognizer

diwu1989:
Bank routing number:
https://en.wikipedia.org/wiki/ABA_routing_transit_number

Has a specific checksum digit which will guarantee good accuracy.


presidio-analyzer/presidio_analyzer/predefined_recognizers/aba_routing_recognizer.py
```diff
@@ -0,0 +1,60 @@
...
+                r"\b[0123678]\d{3}-\d{4}-\d\b",
+                0.3,),
+    ]
+
+    CONTEXT = [
+        "aba",
+        "routing",
+        "abarouting",
+        "association",
+        "bankrouting",
```
diwu1989, line 23: is it necessary to implement these context words if we have a checksum validation?
I noticed that whenever validation passes, the accuracy always becomes 100%

presidio-analyzer/presidio_analyzer/predefined_recognizers/aba_routing_recognizer.py
```diff
@@ -0,0 +1,60 @@
...
+    Recognizes American Banking Association (ABA) routing number.
+    Also known as routing transit number (RTN) and used to identify financial
+    institutions and process transactions.
+    """
+
+    PATTERNS = [
+        Pattern("ABA routing number (weak)", r"\b[0123678]\d{8}\b", 0.05,),
+        Pattern("ABA routing number",
+                r"\b[0123678]\d{3}-\d{4}-\d\b",
+                0.3,),
```
diwu1989, line 15: do these patterns need to be named differently?
Can I name both of these patterns `ABA routing number`?
How should we choose accuracy? I saw that sometimes its 5% and sometimes its 30%

presidio-analyzer/presidio_analyzer/predefined_recognizers/aba_routing_recognizer.py
omri374: It's not necessary, however it would be used anyway (as it uses the context words for every recognizer)

presidio-analyzer/presidio_analyzer/predefined_recognizers/aba_routing_recognizer.py
omri374: The name is mainly for readability. Adding the word "weak" could help future users understand that this might introduce false positives.

The best way to choose accuracy is to test this recognizer with the Presidio evaluation framework (https://github.com/microsoft/presidio-research). There it can be evaluated using real or synthetic datasets and see how well the coverage (recall) is, as well as how many false positives this recognizer introduces.

diwu1989:
How do I trigger the presidio-CI build?

presidio-analyzer/presidio_analyzer/predefined_recognizers/aba_routing_recognizer.py
diwu1989: Okay good, i'll get this other repo set up locally and test my analyzers with it.
Thanks

omri374:
@diwu1989 I triggered it, however we have some issues with our CI so it might not go through due to other reasons. We're working on a solution.

PR #349, opened 2020-07-23
Improve SSN analyzer

diwu1989:
There are some specific SSN patterns that have been mis-used in advertising and are common as sample data.
There's also more specific invalidation checks that can be applied to prune down the number of false positives.

Implementation is using rules from https://help.symantec.com/cs/DLP15.7/DLP/v40736167_v133697641/US-Social-Security-Number-(SSN)-narrow-breadth?locale=EN_US


presidio-analyzer/tests/test_analyzer_engine.py
```diff
@@ -590,7 +590,7 @@ def test_demo_text(unit_test_guid, nlp_engine):
         "Can you please check the status on bank account 954567876544 "
         "in PresidiBank?\n\n"
         ""
-        "Kate's social security number is 078-05-1120.  "
```
diwu1989, line 593: https://www.ssa.gov/history/ssn/misused.html
this is a known mis-used SSN, it's not in circulation.

presidio-analyzer/presidio_analyzer/pattern_recognizer.py
```diff
@@ -96,6 +96,19 @@ def validate_result(self, pattern_text):
         """
         return None
 
+    # pylint: disable=unused-argument, no-self-use, assignment-from-none
+    def invalidate_result(self, pattern_text):
```
omri374, line 100: Do we currently have just one usage for this? If yes, maybe we should wait with abstraction

PR #350, opened 2020-07-25
fix Dockerfile.local

dhpollack:
I had an issue while building the docker images.  Adding the `WORKDIR` to `PYTHONPATH` fixed the issue.


dhpollack:
@omri374 can you run the CI on this?  I had a problem with the module not being found when rebuilding the docker image.

dhpollack:
Alright, I think that I found it.  You have two different pipelines with build steps that have the same name and I only modified one of them before.   Now it should work correctly on CI as well.

presidio-analyzer/Pipfile
```diff
@@ -5,7 +5,7 @@ name = "pypi"
 
 [packages]
 cython = "*"
-spacy = "==2.2.4"
+spacy = "==2.2.3"
```
omri374, line 8: Why the downgrade?

dhpollack: I upgraded it in my previous PR but didn't update it in other places so I realized there was a mismatch in versions. So I downgraded to what it was before.

Issue #355, opened 2020-08-03
presidio-analyzer docker image fails to start


akrock:
**Describe the bug**

Attempting to run the latest docker image results in a python Module error.

```
2020-08-03T02:36:53.781989647Z Traceback (most recent call last):
2020-08-03T02:36:53.78202595Z   File "app.py", line 15, in <module>
2020-08-03T02:36:53.782031594Z     from analyzer_engine import AnalyzerEngine  # noqa
2020-08-03T02:36:53.782035557Z   File "/usr/bin/presidio-analyzer/presidio_analyzer/analyzer_engine.py", line 4, in <module>
2020-08-03T02:36:53.782040406Z     from presidio_analyzer import PresidioLogger, RecognizerRegistry
2020-08-03T02:36:53.782043987Z ModuleNotFoundError: No module named 'presidio_analyzer'
```
**To Reproduce**
1. ` docker run  mcr.microsoft.com/presidio-analyzer:latest`

**Expected behavior**
Maybe an error about not being able to connect to the recognizer store? I'm not sure I have never seen it actually run.

**Additional context**
While I understand that this project is in a pre-production state and the latest builds might not always work, if the only builds that are made available are the latest builds, users would expect them to work. It might be helpful to make tags other than "latest" for known working release in the future?

omri374:
Hi @akrock, thanks for the issue. This is probably an issue with the release pipeline not releasing the latest version. We're looking into this.

omri374:
Released new version, issue is now fixed.

PR #356, opened 2020-08-03
fix: Replace sys.path.append in app.py

akrock:
This PR addresses #355 .

I am by no means a python expert, but this line to register the module's parent directory with `sys.path` seems to address the issue that was preventing the `presidio-analyzer` from running in it's docker container.

This line originally was in `__main__.py` but was not carried over with the migration to `app.py` from these changes: https://github.com/microsoft/presidio/commit/e5fe414fd6108095e9fade11280d8143a9948f88#diff-7667e7401c6e0dc1f5576509d1c53249

### DIFF
--- a/presidio-analyzer/presidio_analyzer/app.py
+++ b/presidio-analyzer/presidio_analyzer/app.py
@@ -12,6 +12,8 @@ from knack.commands import CLICommandsLoader, CommandGroup
 from knack.help import CLIHelp
 from knack.help_files import helps
 
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
 from analyzer_engine import AnalyzerEngine  # noqa
 from recognizer_registry.recognizer_registry import RecognizerRegistry  # noqa
 from nlp_engine import NLP_ENGINES  # noqa


presidio-analyzer/presidio_analyzer/app.py
```diff
@@ -12,6 +12,8 @@
 from knack.help import CLIHelp
 from knack.help_files import helps
 
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```
omri374, line 15:
```suggestion
sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))) # noqa
```

omri374: `# noqa` is also required on the line containing `from protobuf_models...`

omri374:
Hi @akrock, the CI fails because of linting. Could you please add # noqa and verify that flake8 and pylint runs successfully?
```sh
pipenv run pylint presidio_analyzer
pipenv run flake8 presidio_analyzer --exclude "*pb2*.py"
```

omri374:
Replacing with #356

PR #360, opened 2020-08-08
Analyzer sys path fix

omri374:
Fixing #355, Replacing #356 

This line originally was in __main__.py but was not carried over with the migration to app.py from these changes: e5fe414#diff-7667e7401c6e0dc1f5576509d1c53249


omri374:
Replaces #356

PR #387, opened 2020-12-10
Small bug fixes

omri374:
1. typo in `default_language` value in `AnalyzerEngine`
2. `NlpArtifacts` now pass full token (not just text) for easier downstream analysis of token attributes.
3. Formatting


omri374:
addressed in #392

PR #393, opened 2021-01-05
Initial CI for analyzer V2

omri374:
Initial CI pipeline for V2 analyzer. Failing because it runs on the old analyzer


azure-pipelines.yml
```diff
@@ -0,0 +1,61 @@
...
+
+jobs:
+- job: 'Test'
+  pool:
+    vmImage: 'ubuntu-16.04'
+  strategy:
+    matrix:
+      Python37:
+        python.version: '3.7'
+#      Python38:
```
navalev, line 17: should be also run on version 3.6?

navalev (dismissed):
looks good for the initial pipeline, as discussed we will extend it in the next iteration. Let's commit to be used for additional PRs

azure-pipelines.yml
omri374: Possibly. Let's see how long it takes first and decide if we want to add more options?

navalev: agree

azure-pipelines.yml
```diff
@@ -0,0 +1,61 @@
...
+        pipenv --python 3
+  - task: Bash@3
+    displayName: 'Install deps: Analyzer'
+    inputs:
+      targetType: 'inline'
+      workingDirectory: 'presidio-analyzer'
+      script: |
+        # Azure pipelines test publish
+        pipenv sync --dev --sequential
+        pipenv install --dev --skip-lock pytest-azurepipelines
```
balteravishay, line 43: can this be taken to a different task so we can know which tasks are "core presidio" and which are "pipeline utility"?

azure-pipelines.yml
```diff
@@ -0,0 +1,61 @@
...
+  - task: Bash@3
+    displayName: 'Pytest: Analyzer'
+    inputs:
+      targetType: 'inline'
+      workingDirectory: 'presidio-analyzer'
+      script: |
+        pipenv run pytest
+
+  - task: Bash@3
+    displayName: 'Linting: Analyzer'
```
balteravishay, line 55: shouldn't lint be before test?

azure-pipelines.yml
omri374: Fixed. moved `pipenv install --dev --skip-lock pytest-azurepipelines` to the testing task

azure-pipelines.yml
omri374: Fixed

azure-pipelines.yml
navalev: wouldnt we replace the "core presidio" with a build script?

omri374: I assume this is subject to change once we have a solid build script.

navalev (approved):

SharonHart (approved):

PR #396, opened 2021-01-11
2603 - Create anonymize method stub

shiranr:
Add method stub
Add flask as part of the dependencies
Add test for method stub


SharonHart:
Can we maybe split linting, depnedencies, and flask+stubs into different PRs?

presidio-anonymizer/tests/test_anonymizer.py
```diff
@@ -0,0 +1,9 @@
+import unittest
+from anonymizer import app
+
+
+class TestAnonymizer(unittest.TestCase):
```
SharonHart, line 5: What are we planning to test here? The controller layer?

shiranr:
need to redo this PR.

presidio-anonymizer/tests/test_anonymizer.py
shiranr: This test for now, is just a POC to see the method is working with flask and REST. It will change once we write code here.

PR #397, opened 2021-01-12
2608 - create template for anonymizer

shiranr:
Create template for anonymizer from cookiecutter and clean it up to our needs.


presidio-anonymizer/setup.cfg
```diff
@@ -0,0 +1,7 @@
+[bdist_wheel]
+# This flag says to generate wheels that support both Python 2 and Python
+# 3. If your code will not run unchanged on both Python 2 and 3, you will
+# need to generate separate wheels for each Python version that you
+# support.
+universal=1
```
omri374, line 6: This means that we can run on Python 2 which isn't the case.

omri374:
Looks great! Thanks @shiranr

presidio-anonymizer/setup.cfg
shiranr: I thought I saw this in the analyzer as well so I assumed it's ok.

shiranr: I removed it.

omri374: You right! I'll remove it from there as well.

presidio-anonymizer/Pipfile
```diff
@@ -0,0 +1,13 @@
...
+name = "pypi"
+
+[packages]
+flask = "==1.1.2"
+
+[dev-packages]
+pytest = "*"
+pylint = { version = "==2.3.1" }
+flake8= {version = "==3.7.9"}
+pytest-azurepipelines = "*"
```
omri374, line 13: Not necessary. We only use it during CI

omri374 (approved):

azure-pipelines.yml
```diff
@@ -60,3 +60,22 @@ jobs:
...
         pipenv install --dev --skip-lock pytest-azurepipelines
         pipenv run pytest
+
+  - task: Bash@3
+    displayName: 'Linting: Anonymizer'
+    inputs:
+      targetType: 'inline'
+      workingDirectory: 'presidio-anonymizer'
+      script: |
+        pipenv run pylint presidio_anonymizer
```
SharonHart, line 70: You can remove this line, we will use just flake8

presidio-anonymizer/Pipfile
```diff
@@ -0,0 +1,12 @@
...
+url = "https://pypi.python.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+flask = "==1.1.2"
+
+[dev-packages]
+pytest = "*"
+pylint = { version = "==2.3.1" }
```
SharonHart, line 11: remove pylint

navalev (approved):

presidio-anonymizer/setup.cfg
```diff
@@ -0,0 +1,2 @@
+[flake8]
+max-line-length = 89
```
SharonHart, line 2: Will have to be aligned with black, @omri374 are we going with a custom length different from PEP8 from a special reason?

SharonHart (approved):
Nice, left a small comment on pylint

SharonHart (dismissed):

presidio-anonymizer/setup.cfg
omri374: Black uses 88 which is not aligned with PEP8 but makes more sense with modern monitors. This should also be 88 and not 89. Agree.

omri374 (approved):

presidio-anonymizer/setup.cfg
shiranr: @omri374  it should also be changed in the analyzer to 88. The value is currently similar in both.

azure-pipelines.yml
```diff
@@ -60,3 +60,21 @@ jobs:
         # Install pytest and run tests
         pipenv install --dev --skip-lock pytest-azurepipelines
         pipenv run pytest
+
```
balteravishay, line 63: Not sure this duplication is required. But can be fixed in the ci user story

balteravishay (dismissed):
Approve with minor comment

PR #398, opened 2021-01-12
V2 Linting and Pre-Commit Hooks

SharonHart:
PR Objectives:

- [x] Remove pylint, leaving just flake8 as a code checker, both in the CI and locally.
- [x] Add  pre-commit configuration to project root, with black and flake8 hooks, for autoformatting and checking when committing, to find and fix issues before the build process.
- [x] Updates development documentation with local pre-commit usage.

(each in a commit to ease on the review)

Will be happy if the reviewer will use the documentation and test the logics locally to eliminate environment-related issues.


docs/development.md
```diff
@@ -0,0 +1,36 @@
+# Setting Up a Development Environment
+
+
+## Table of contents
+
+1. [Automatically format code and check for code styling](#python-styling)
```
SharonHart, line 6: Will not necessarily be #1, just a placeolder

shiranr (approved):

presidio-analyzer/Pipfile.lock
```diff
@@ -522,80 +501,13 @@
...
-        },
         "toml": {
             "hashes": [
                 "sha256:806143ae5bfb6a3c6e736a764057db0e6a0e05e338b5630894a5f779cabb4f9b",
                 "sha256:b3bda1d108d5dd99f4a20d24d9c348e91c4db7ab1b749200bded2f839ccbe68f"
             ],
             "markers": "python_version >= '2.6' and python_version not in '3.0, 3.1, 3.2, 3.3'",
             "version": "==0.10.2"
-        },
-        "typed-ast": {
```
SharonHart, line 541: @omri374 Do you have any idea why there are removals?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,11 @@
...
+  -   repo: https://github.com/ambv/black
+      rev: stable
+      hooks:
+        - id: black
+          language_version: python3.7
+  -   repo: https://gitlab.com/pycqa/flake8
+      rev: 3.7.9
+      hooks:
+        - id: flake8
+          args: ['--max-line-length=88']
```
SharonHart, line 11: Since it defined in the root, flake is not aware of /{presidio-service}/setup.cfg flake8 configuration.
This mean we either have to repeat the config here.
I've tried to define a global flake8 configuration (makes sense since we have duplication between services anyway) in the project root but it was ignored from some reason. @omri374 do you know if its doable?

docs/development.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Automatically format code and check for code styling <a name='python-styling'></a>
+
+Presidio services are PEP8 compliant and continuously enforced on style guide issues during the build process using `flake8`.
+
+Running flake8 locally, using `pipenv run flake8`, you can check for those issues prior to committing a change.
+
+To make stuff easier, you can use pre-commit hooks to verify and automatically format code upon a git commit, using `black`:
```
omri374, line 14:
```suggestion
To make things easier, you can use pre-commit hooks to verify and automatically format code upon a git commit, using `black`:
```
nit :)

docs/development.md
```diff
@@ -0,0 +1,38 @@
...
+
+1. [Install pre-commit package manager locally.](https://pre-commit.com/#install) 
+
+1. From the project's root, enable pre-commit, installing git hooks in the `.git/` directory:
+
+    `pre-commit install`
+
+2. Commit non PEP8 compliant code will cause commit failure and automatically format your code using `black`, as well as checking code formatting using `flake8`
```
omri374, line 22: Worth mentioning that although we are PEP8 compliant, the max line width is 88?

SharonHart: seems like noise here, maybe above when I state 'Presidio services are PEP8 compliant..' ?
Maybe 'Presidio services are PEP8 compliant (with minor exclusions)'

omri374: I agree. let's just keep it as PEP8 compliant.

docs/development.md
SharonHart: Done

docs/development.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Automatically format code and check for code styling <a name='python-styling'></a>
+
+Presidio services are PEP8 compliant and continuously enforced on style guide issues during the build process using `flake8`.
+
+Running flake8 locally, using `pipenv run flake8`, you can check for those issues prior to committing a change.
+
+To make things easier, you can use pre-commit hooks to verify and automatically format code upon a git commit, using `black`:
+
+1. [Install pre-commit package manager locally.](https://pre-commit.com/#install)
```
omri374, line 16: If it's just `pip install pre-commit` then I would consider writing it here.

SharonHart: You can choose, its pip install in your own venv, its brew, its curl to the installation python file.

navalev (approved):
looks great. approving per the offline teams discussion

docs/development.md
shiranr: how do we know someone actually ran this before the commit? 
Is there going to be some check in the pipeline?

SharonHart: We can't, and we can't force to use it, it's suppose to help us find and fix issues before the build (which will still fail regardless of using pre-commit)

shiranr (approved):
Looks good. 👍 
Small remark: we need to consider adding a stage to the pipeline so we fail the build if the user did not run the pre-commit script.
I don't think it is top priority but it's nice to have a ticket about it for the future.

SharonHart:
> Looks good. 👍
> Small remark: we need to consider adding a stage to the pipeline so we fail the build if the user did not run the pre-commit script.
> I don't think it is top priority but it's nice to have a ticket about it for the future.

It's already there, flake8 is currently running in the build

SharonHart:
@omri374 let's extend this, in another PR, using flake8 plugins. It's actually becoming preferred since we can have more control on what we check for, and as of now, run it as part of the pre-commit as well.

PR #400, opened 2021-01-13
Create anonymize method stub with flask.

shiranr:
Create anonymize method stub with flask

### DIFF
--- a/presidio-anonymizer/presidio_anonymizer/presidio_anonymizer.py
+++ b/presidio-anonymizer/presidio_anonymizer/presidio_anonymizer.py
@@ -1,1 +1,13 @@
-"""Main module."""
+from flask import Flask, request
+
+app = Flask(__name__)
+
+
+@app.route('/anonymize', methods=['POST'])
+def anonymize():
+    content = request.json
+    return content
+
+
+if __name__ == 'main':
+    app.run()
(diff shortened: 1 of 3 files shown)


presidio-anonymizer/presidio_anonymizer/presidio_anonymizer.py
```diff
@@ -1 +1,13 @@
-"""Main module."""
+from flask import Flask, request
```
omri374, line 1: We should decide if the flask app is part of the package (i.e. inside the `presidio_anonymizer` folder) or outside, in some `app.py` file, which is similar to how we had the grpc server set up.

omri374 (approved):
Previous

presidio-anonymizer/presidio_anonymizer/presidio_anonymizer.py
SharonHart, line 1: I would start structuring the project, have this file just run the app and move the route to views.py / routes.py

SharonHart (approved):
Minor comment + please add a PR description

presidio-anonymizer/presidio_anonymizer/presidio_anonymizer.py
shiranr: Not sure I understand the remark. 
The entire application is going to be inside the presidio_anonymizer folder just like in the anonymizer (I am trying to keep consistency between the applications).
If I am not mistaking @itye-msft and I are per-programming the flask on the analyzer, so if the layout will change, we will also change the anonymizer.

shiranr: Will be done later with the analyzer for consistency.

PR #406, opened 2021-01-18
Dockerize Anonymizer + Add Compose

SharonHart:
* Add an Anonymizer Dockerfile.
* Add docker-compose file with build context, currently starting an anonymizer instance on port 5001
* Changed api.py to app.py to match flask convention

Run (and build) locally from root with `docker-compose up --build`

**This is not enough for the production use-case, where we need to use wsgi like gunicorn, we don't need it currently and we might add this later for the demo server**


docker-compose.yml
```diff
@@ -0,0 +1,11 @@
...
+services:
+  presidio-anonymizer:
+    build:
+      context: .
+      args:
+        - NAME=presidio-anonymizer
+    environment:
+      - FLASK_RUN_PORT=5001
+    ports:
+      - "5001:5001"
```
SharonHart, line 11: For example in analyzer:
```
  presidio-analyzer:
    build:
      context: .
      args:
        - NAME=presidio-analyzer
        - SPACY_MODEL=en_core_web_lg
    environment:
      - FLASK_RUN_PORT=5002
    ports:
      - "5002:5002"

```

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM python:3.8-slim
+
+ARG NAME
+ARG SPACY_MODEL
+WORKDIR /usr/bin/${NAME}
+
+COPY ./${NAME} /usr/bin/${NAME}
+
+RUN pip install pipenv
+RUN pipenv sync
```
SharonHart, line 10: Kept pipenv instantiation from the lock file, although not the best practice. @omri374 stated that it solved issues in the past.

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+RUN if [ "$SPACY_MODEL" != "" ] ; then pipenv run python -m spacy download "$SPACY_MODEL" ; fi
```
navalev, line 11: following our offline discussion - I'm in favor of spliting this into seperate dockerfiles (analyzer and anonymizer) where each has it our arguments. In this case, there's no reason for the anonymizer to have the space model version. And I assume further service-specific arguments (for future services we will convert) will be required. 
Let's continue this offline

navalev (changes requested):

Dockerfile
SharonHart: Completely legitimate option, had the same doubt.
What we can also do, is have a common dockerfile (without this line), and a default spacy in analyzer pip. Additionaly, there will be another dockerfile for building analyzer with a different spacy model, overriding the one defined in pip (it will essentially be the dockerfile suggested here). Downside is having to update both on changes.
I don't think that we will have much difference in the future since most of the differences can be handled with pip and environment variables.

omri374: @SharonHart what do you mean by "overriding the one defined in pip"? Could you add some examples on how this would look like?

Dockerfile
omri374: If we decide to go this way, then I would use the same way which we had in V1: https://github.com/microsoft/presidio/blob/4b3f637a6337fcdf54a07f5d1e78559dd68945a1/Dockerfile.python.deps#L26

Let's discuss offline.

omri374: If we decide to go this way, then I would use the same way which we had in V1: https://github.com/microsoft/presidio/blob/4b3f637a6337fcdf54a07f5d1e78559dd68945a1/Dockerfile.python.deps#L26

Let's discuss offline.

Dockerfile
SharonHart: Sure,
I think that we should use 'en_core_web_sm' as default, since it's faster for CI and working with locally. 
It will be part of analyzer pipenv, incorporated in the Pipfile and Pipfile.lock (making the dockerfile completely common).
Additionally, we can add another spacy model, installed with pipenv as an addition to en_core_web_sm, or override en_core_web_sm. We can control what the application uses with an environment variable.
It can be done in the separate dockerfile (if we decide to separate the services), or we can have additional one for reference only, for whomever wants to use a different one.

balteravishay: vote for as many dockerfiles as required. keep the CI logic to minimum.

balteravishay (commented):
comment about dockerfile

navalev (approved):

PR #410, opened 2021-01-19
Feature/2539 analyzer http

itye-msft:
A flask wrapper for Presidio analyzer.
Integration tests are not included as the framework is not ready yet.
Solves User Story: [2539](https://dev.azure.com/csedevil/Presidio-internal/_workitems/edit/2539)


Pipfile
```diff
@@ -0,0 +1,11 @@
+[[source]]
```
SharonHart, line 1: probably here by mistake

presidio-analyzer/app.py
```diff
@@ -0,0 +1,85 @@
...
+                names = [o.name for o in recognizers_list]
+                return json.dumps(names)
+            except Exception as e:
+                self.logger.error("A fatal error occurred during execution of AnalyzerEngine.get_recognizers(). {}"
+                                  .format(e))
+                return json.dumps({"Error": e}), 500
+
+
+if __name__ == "__main__":
+    port = int(os.environ.get("PORT", DEFAULT_PORT))
```
SharonHart, line 83: by default flask has 'FLASK_RUN_PORT' env var, you can use it instead

presidio-analyzer/app.py
```diff
@@ -0,0 +1,85 @@
...
+            """
+            # Parse the request params
+            req_data = request.get_json()
+
+            text = req_data.get("text")
+            language = req_data.get("language")
+            entities = req_data.get("entities")
+            correlation_id = req_data.get("correlation_id")
+            score_threshold = req_data.get("score_threshold")
+            trace = req_data.get("trace")
```
SharonHart, line 50: since all the names are the same, maybe you can use **kwargs

SharonHart (changes requested):
Overall looks good! left some suggestions. Also I think that some of the stuff might fail linting, you can activate the pre-commit hook to catch them

presidio-analyzer/app.py
itye-msft: Yeah, it work when you run flask directly (flask run), but for our case it won't work

presidio-analyzer/app.py
itye-msft: I think it's a good practice not to rely on the caller, and let the app read and in some cased verify.

SharonHart: Currently a missing key will be passed as None either way and you will have to validate anyway in the method.
Only difference is if the argument names won't match

itye-msft: I prefer to do this middleware op, it's more safe for the long run

VERSION
```diff
@@ -1 +1 @@
-1.3.0
```
SharonHart, line 1: We bump the minor, in the middle

presidio-analyzer/README.MD
```diff
@@ -24,8 +24,14 @@ python -m spacy download en_core_web_sm
 ```
 
 ## Getting started
+### Running Presidio as an HTTP server
+```sh
+cd presidio-analyzer
+python app.py
+curl -d '{"text":"John Smith drivers license is AC432223", "language":"en"}' -H "Content-Type: application/json" -X POST http://localhost:3000/analyze
+```
```
SharonHart, line 32: We might want to direct to use compose here later on. Even if it's just a single service, we can show how to run compose with one service only and analyzer as exampel

balteravishay: i think running as a python package first is ok. but is there a difference between doing
python app.py and doing
flask run
?

balteravishay (approved):
approve with minor question on doc

presidio-analyzer/README.MD
itye-msft: If it's not flask, the doc will be incorrect. 
By running Python app.py, you can ensure the app takes care of itself.

SharonHart: From the exact reason, I would document the orchestrated way, app.py might change as well :)
Also takes care of env vars that might be needed later on, and since documented in presidio docs anyway, less confusing

presidio-analyzer/setup.py
```diff
@@ -27,6 +27,7 @@
         "spacy==2.3",
         "regex==2020.11.13",
         "tldextract==3.1.0",
+        "flask==1.1.2"
```
navalev, line 30: need to also update the NOTICE file with flask

presidio-analyzer/app.py
```diff
@@ -0,0 +1,85 @@
...
+class Server:
+
+    def __init__(self):
+        self.logger = PresidioLogger(os.environ.get("PRESIDIO_LOGGER"))
+        self.app = Flask(__name__)
+        self.logger.info("Starting analyzer engine")
+        self.engine = AnalyzerEngine()
+        self.logger.info(WELCOME_MESSAGE)
+
+        @self.app.route('/health')
```
navalev, line 30: let also add /health route to the anonymizer. might be usuful for smoke tests

presidio-analyzer/app.py
```diff
@@ -0,0 +1,85 @@
...
+
+        @self.app.route("/analyze", methods=["POST"])
+        def analyze():
+            """
+            Executes the analyzer function
+            """
+            # Parse the request params
+            req_data = request.get_json()
+
+            text = req_data.get("text")
```
navalev, line 45: I would avoid breaking the json here, and pass the request to a seperate module that will handle it. makes the routes more readable

itye-msft: Done

SharonHart (dismissed):

navalev (dismissed):

balteravishay (approved):

navalev (dismissed):

SharonHart (approved):

PR #411, opened 2021-01-19
E2E Tests Initialization

SharonHart:
Add E2E tests infra under functional-tests/
Made some decisions I want an opinion on:
1. Since we will use a fixed low number of api methods, I've decided to encapsulate all in a single common file (can be split to two later on).
2. Each method will return the status code and deserialized response, so that the scenario will not have to use json.loads each time it asserts or modifies the response. This also forces the developer to keep the scenarios simple, as we want them to be in e2e.
3. Under tests/ we will have a test_{service}.py file with service-specific scenarios, and test_integration.py file with tests calling multiple services. Scenarios will be mixed there with test suites tags defining what layer will they test, between api, integration between services, smoke.


omri374 (dismissed):
LGTM

navalev (dismissed):
looks good. Lets also add a readme outlining what is the structure of the tests, what is expected from a contributer (in a speperate PR)

navalev (approved):

omri374 (approved):

PR #419, opened 2021-01-21
Run Pipeline on Merge to V2 + Deploy to Dev. Env. Stub

SharonHart:
Decided to have a single pipeline, only have the deployment task under a branch condition:
https://docs.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops&tabs=yaml#run-for-the-main-branch-if-succeeding
In another solution, we will have to duplicate the pipeline and extract its content to a template, making it less readable.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,11 +1,6 @@
 trigger:
   - V2
 
-pr:
-  branches:
-    include:
-      - V2
-
 jobs:
     #- job: validate
     #  displayName: 'Validate PR pre-requisites'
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -1,11 +1,6 @@
 trigger:
   - V2
 
-pr:
-  branches:
-    include:
-      - V2
-
```
SharonHart, line 8: Defined in the UI

SharonHart:
Closing, will be done once we add deployment to dev. env.

navalev (approved):

omri374 (approved):

omri374:
@SharonHart should this be merged or only after we replace the stub?

SharonHart:
@omri374 Not really, but I see that Avishay's PR is going to be merged soon so this might not have that much content after all. Let's just leave it open for now

PR #420, opened 2021-01-21
Analyzer dockerfile

balteravishay:
adding a dockerfile for analyzer which allows different spacy models to be used as argument.
changing also the anonymizer dockerfile to run python instead of flask.


SharonHart:
@balteravishay can you add it to compose as well and validate? Also a simple functional scenario will be great

presidio-anonymizer/Dockerfile
```diff
@@ -11,4 +11,4 @@ RUN pipenv sync
 COPY . /usr/bin/${NAME}/
 
 EXPOSE ${FLASK_RUN_PORT}
-CMD pipenv run flask run --host 0.0.0.0
\ No newline at end of file
+CMD pipenv run python app.py --host 0.0.0.0
```
SharonHart, line 14: Problem was that app was set with name by self.app = Flask(__name__)
but started from the global context, not self :
server = Server()
server.app.run(port=port)
As long as it works, it doesn't matter

navalev (changes requested):
could you please also update the documentation, maybe add info on how we can leverege this to use other spacy models.

navalev (approved):

presidio-analyzer/README.MD
```diff
@@ -24,14 +24,38 @@ python -m spacy download en_core_web_sm
...
+Note: you can change the spacy model which is downloaded to the container at build time, using the following build command:
+
+```sh
+docker build -t presidio-analyzer --build-arg NAME=presidio-analyzer  --build-arg SPACY_MODEL=en_core_web_sm .
+```
+
+TODO: The Analyzer is hard-coded to use the large english model. Extending to any model will be worked on in user story #2675
+
+TODO: The Analyzer container is built with a single spacy model. Add support for multiple models will worked on in user story #2678
+
```
SharonHart, line 57: We can also document the compose way, `docker-compose up presidio-anonymizer`, not sure whats best:
From one end, we document compose on the 'how to run functional tests' so it will be a uniform single way of running all that you need, maybe less confusing.
On the other hand, it will require to download compose as a prerequisite on a simple change not requiring a functional test change

functional-tests/tests/test_analyzer.py
```diff
@@ -0,0 +1,13 @@
...
...\': 0, \'end\': 10, \'score\': 0.85, \'analysis_explanation\': {\'recognizer\': \'SpacyRecognizer\', \'pattern_name\': None, \'pattern\': None, \'original_score\': 0.85, \'score\': 0.85, \'textual_explanation\': "Identified as PERSON by Spacy\'s Named Entity Recognition", \'score_context_improvement\': 0, \'supportive_context_word\': \'\', \'validation_result\': None}}', "{'entity_type': 'US_DRIVER_LICENSE', 'start': 30, 'end': 38, 'score': 0.6499999999999999, 'analysis_explanation': {'recognizer': 'UsLicenseRecognizer', 'pattern_name': 'Driver License - Alphanumeric (weak)', 'pattern': '\\\\b([A-Z][0-9]{3,6}|[A-Z][0-9]{5,9}|[A-Z][0-9]{6,8}|[A-Z][0-9]{4,8}|[A-Z][0-9]{9,11}|[A-Z]{1,2}[0-9]{5,6}|H[0-9]{8}|V[0-9]{6}|X[0-9]{8}|A-Z]{2}[0-9]{2,5}|[A-Z]{2}[0-9]{3,7}|[0-9]{2}[A-Z]{3}[0-9]{5,6}|[A-Z][0-9]{13,14}|[A-Z][0-9]{18}|[A-Z][0-9]{6}R|[A-Z][0-9]{9}|[A-Z][0-9]{1,12}|[0-9]{9}[A-Z]|[A-Z]{2}[0-9]{6}[A-Z]|[0-9]{8}[A-Z]{2}|[0-9]{3}[A-Z]{2}[0-9]{4}|[A-Z][0-9][A-Z][0-9][A-Z]|[0-9]{7,8}[A-Z])\\\\b', 'original_score': 0.3, 'score': 0.6499999999999999, 'textual_explanation': None, 'score_context_improvement': 0.3499999999999999, 'supportive_context_word': 'driver', 'validation_result': None}}"]
```
SharonHart, line 13: Not part of this PR, and not its purpose, but we might want to unescape here and make it more readable. Also maybe reduce the response payload by changing the request if possible

SharonHart (approved):
Approved with suggestions

functional-tests/tests/test_analyzer.py
omri374: We have a US for reducing the payload size

PR #424, opened 2021-01-24
Add logger and log messages to anonymizer

shiranr:
Add logger and log messages to anonymizer


presidio-anonymizer/app.py
```diff
@@ -25,9 +28,11 @@ def anonymize():
                 data = AnonymizerRequest(content)
                 text = AnonymizerEngine().anonymize(data)
             except InvalidParamException as e:
-                return e.err, 400
-            except Exception:
-                # TODO add logger 2652
+                logging.debug(
```
balteravishay, line 31: maybe error or warn?

presidio-anonymizer/app.py
```diff
@@ -25,9 +28,11 @@ def anonymize():
...
+                    {f"failed to anonymize text with validation error: {e.err_msg}"})
+                return e.err_msg, 400
```
balteravishay, line 33: consider error code 422 Unprocessable Entity, since this is semantic error and not a general purpose "bad request"

balteravishay (commented):
some comments

presidio-anonymizer/app.py
shiranr: Done.

presidio-anonymizer/app.py
shiranr: Very nice! thanks. Done.

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
```diff
@@ -34,11 +34,14 @@ def anonymize(self, engine_request: AnonymizerRequest) -> str:
...
+            self.logger.debug(
+                f"for analyzer result {analyzer_result} received transformation "
+                f"{str(transformation)}")
             anonymizer_class = transformation.get("anonymizer")
             new_text = anonymizer_class().anonymize(transformation)
             end_of_text = min(analyzer_result.end, last_replacement_point)
             output_text = output_text[:analyzer_result.start] + new_text + output_text[
                                                                            end_of_text:]
             last_replacement_point = analyzer_result.start
+            self.logger.debug(f"text after replacement is {output_text}")
```
SharonHart, line 46: Maybe original_text as well?

presidio-anonymizer/app.py
```diff
@@ -25,9 +28,11 @@ def anonymize():
                 data = AnonymizerRequest(content)
                 text = AnonymizerEngine().anonymize(data)
             except InvalidParamException as e:
-                return e.err, 400
-            except Exception:
-                # TODO add logger 2652
+                logging.warning(
+                    {f"failed to anonymize text with validation error: {e.err_msg}"})
+                return e.err_msg, 422
```
SharonHart, line 33: Good change to add api tests to the functional tests module to verify that error.

presidio-anonymizer/presidio_anonymizer/entities/analyzer_result.py
```diff
@@ -109,5 +116,6 @@ def validate_and_create(cls, content):
     def __validate_fields(cls, content):
         for field in ("start", "end", "score", "entity_type"):
             if content.get(field) is None:
+                cls.logger.debug(f"invalid input, no field {field} for {content}")
```
SharonHart, line 119: just curios, any particular reason to define it in the class? unlike other cases?

SharonHart (commented):
Looks good, minor comments

presidio-anonymizer/app.py
SharonHart: Not for all cases though.. for not passing a json at all or 'analyzer_results' at all, 422 might be misleading

presidio-anonymizer/app.py
```diff
@@ -25,9 +28,11 @@ def anonymize():
                 data = AnonymizerRequest(content)
                 text = AnonymizerEngine().anonymize(data)
             except InvalidParamException as e:
-                return e.err, 400
-            except Exception:
-                # TODO add logger 2652
+                logging.warning(
+                    {f"failed to anonymize text with validation error: {e.err_msg}"})
```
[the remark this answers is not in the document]
shiranr: The message here is unique to the issue. I don't think we need the entire trace. 
It will seem as an internal exception instead of an error of the user usage.

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
shiranr: Will be done once we add all the anonymizers. They require it. I don't want to add unneeded values for now...

presidio-anonymizer/app.py
shiranr: Added.

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
SharonHart: What do you mean? don't we have original_text on the top? We can just print it once for easy debug

presidio-anonymizer/app.py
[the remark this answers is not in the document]
SharonHart: I'm used to taking the trace and paste it, instead of just looking for strings, can show me the whole callstack as well. Your choice, its for us so I don't see why not being more verbose when we debug issues.

presidio-anonymizer/app.py
SharonHart: I meant to the functional tests module

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
shiranr: The original_text is the entire text is can be with 1,000 lines and more. 
I don't think it should be printed in debug.

presidio-anonymizer/app.py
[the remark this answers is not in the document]
shiranr: I don't think exceptions should be in debug mode when it's not really an exception but a user error.

presidio-anonymizer/app.py
shiranr: yes, I know, didn't push it yet but it's added.

presidio-anonymizer/presidio_anonymizer/entities/analyzer_result.py
shiranr: Done.

presidio-anonymizer/app.py
shiranr: That's Avishays comment since the handling of an empty json is not in this section :)

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
SharonHart: Do we write debug by default to app insights?

presidio-anonymizer/app.py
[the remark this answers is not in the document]
SharonHart: There are static code analysis tools that might actually fail in that case, its considered best practice to reserve the exception, even if its info:
https://jira.sonarsource.com/browse/RSPEC-1166

presidio-anonymizer/app.py
SharonHart: What about not passing analyzer_results?
Not critical since we are not 100% restful anyway, but might be confusing

balteravishay (approved):

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
omri374: Original text contains PII. We should be careful about logging it.

SharonHart: In that case, when logging the output we might log PII which we failed to anonymize?

omri374: Yes, could be.

balteravishay (approved):

balteravishay (approved):

SharonHart (approved):

PR #432, opened 2021-01-25
Push containers in CI

balteravishay:
This PR adds push containers to private ACR during CI so that we can use those images to deploy the dev environment.
one thing to note is that the PR/CI pipeline starts having some conditions and conditional insertions which makes it harder to understand.

also notice that docker compose commands do not change. the arguments are added so that they are optional.

here is a sample of the pipeline running on a CI trigger (w. docker build and push job and pull during functional test)
https://dev.azure.com/csedevil/Presidio/_build/results?buildId=10178&view=results

here is a sample of the pipeline running on PR/manual (w. docker build during functional tests and no push)
https://dev.azure.com/csedevil/Presidio/_build/results?buildId=10182&view=results


SharonHart (dismissed):

azure-pipelines.yml
```diff
@@ -127,15 +129,67 @@ jobs:
...
+          TAG: ':$(Build.BuildId)'
       steps:
+        - task: DockerCompose@0
+          condition: ne(variables['Build.Reason'], 'IndividualCI')
+          displayName: Build Presidio Images
+          inputs:
+              action: Build services
+              dockerComposeFile: docker-compose.yml
+        - task: DockerCompose@0
+          displayName: Pull Presidio Images to ACR
```
shiranr, line 183: is it pull Persidio images from ACR
or 
Push presidio images to ACR?
I am guessing the first?

azure-pipelines.yml
```diff
@@ -127,15 +129,67 @@ jobs:
           parameters:
             SERVICE: 'Anonymizer'
             WORKING_FOLDER: 'presidio-anonymizer'
+    - job: BuildContainers
```
navalev, line 132: can we seperate this into a template?

shiranr (dismissed):

navalev (changes requested):

azure-pipelines.yml
balteravishay: done

azure-pipelines.yml
balteravishay: done

SharonHart (approved):

shiranr (approved):

navalev (approved):

PR #436, opened 2021-01-26
[DO NOT MERGE] Proposal for batch text analysis

omri374:
This PR includes a new method for Presidio analyzer, `analyze_batch` which allows the analysis of more than one value.
It currently receives a dictionary with a list of strings as each value. 

This can be useful for table like structures where the keys are the column names and the values are the different columns, or for JSON files which can be analyzed as a dictionary.

FYI @willsthompson

### DIFF
--- a/presidio-analyzer/presidio_analyzer/analyzer_engine.py
+++ b/presidio-analyzer/presidio_analyzer/analyzer_engine.py
@@ -1,5 +1,5 @@
 import json
-from typing import List, Optional
+from typing import List, Optional, Dict, Union
 
 from presidio_analyzer import (
     RecognizerRegistry,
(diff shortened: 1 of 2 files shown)


willsthompson:
This is great @omri374, thanks for getting it kicked off! One thing I'd like to see included as part of JSON support is differentiated context behavior for JSON keys.

For example, how we might want to modify a result based on a key in a key/value relationship is likely different from current assumptions made about context (based on words surrounding a matching value). Some differences I think would be helpful to address:

* It's common for keys to take on arbitrary formats (`custId`, `USER_NAME`, etc.), so I think patterns would be necessary to match on keys, rather than lists of word tokens.
* In some cases it may be too difficult or unnecessary to match an entity based on its value, and the key is the only string of significance. For these cases it would be helpful to be able to build recognizers based only on a key match and no value match.
* While surrounding words are parsed at runtime within the context enhancement function, key context is static and known ahead of time, so it may be useful to make it available for access from other recognizer functions like `validate_result()` or `invalidate_result()`

I don't have a good idea of how disruptive these changes would be to the analyzer, however. Maybe it could be done without significant design changes?

omri374:
Great points @willsthompson.
I think that context support for keys is definitely a required addition here. Let me think about this a bit and get back to you.

omri374:
We've created some documentation on this matter here: #714
Please share any feedback you have on this. Closing this PR for now until we decide on next steps.

PR #438, opened 2021-01-27
Anonymizer to receive original_text

SharonHart:
Both mask and hash require the original text for the anonymization.
This changes the interface, so that anonymize method will receive it.


presidio-anonymizer/presidio_anonymizer/anonymizers/hash.py
```diff
@@ -1,11 +1,12 @@
 """Hashes the PII text entity."""
+from presidio_anonymizer.anonymizers import Anonymizer
 
 
-class Hash(object):
+class Hash(Anonymizer):
```
SharonHart, line 5: @shiranr this was not intentional right?

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
```diff
@@ -30,17 +30,24 @@ def anonymize(self, engine_request: AnonymizerRequest) -> str:
         original_text = engine_request.get_text()
         last_replacement_point = len(original_text)
         output_text = engine_request.get_text()
-        analyzer_results = engine_request.get_analysis_results(
-        ).to_sorted_unique_results(True)
+        analyzer_results = (
```
itye-msft, line 33: What are the new braces for?

itye-msft (approved):
Approved with a question

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
SharonHart: Actually I'm not sure, black added them

navalev (approved):
apprvoing - alignes with our offline discussion

presidio-anonymizer/presidio_anonymizer/anonymizers/hash.py
shiranr: Nope, an honest mistake. :)

presidio-anonymizer/presidio_anonymizer/anonymizers/anonymizer.py
```diff
@@ -9,6 +9,6 @@ class Anonymizer(ABC):
     logger = logging.getLogger("presidio-anonymizer")
 
     @abstractmethod
-    def anonymize(self, params={}):
+    def anonymize(self, original_text=None, params={}):
```
shiranr, line 12: Please notice, not all anonymize methods require the original_text.

shiranr: The original_text can be passed through params
or better yet we can change params to **kwargs

SharonHart: I know, but can we avoid it in the interface level?

shiranr (dismissed):

presidio-anonymizer/presidio_anonymizer/anonymizers/anonymizer.py
shiranr: yes, use **kwargs instead of params.
params was a temporary place holder

SharonHart:
@shiranr I don't think **kwargs can go on an interface, also params need to be method specific, and the current framework is not supporting appending to it as it comes from transformation straight from the payload. 
What do you suggest?

presidio-anonymizer/presidio_anonymizer/anonymizers/anonymizer.py
SharonHart: **kwargs essentially let you pass anything so it misses the idea of an interface.
Interface changes between implementations, so it breaks the open close principle

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
itye-msft: Looks unnecessary

PR #439, opened 2021-01-27
Feature/remove interpetability

itye-msft:
Add a new param to remove analysis explanation from the response.
Added Unit tests + Integration tests
Update the docs

### DIFF
--- a/presidio-analyzer/presidio_analyzer/analyzer_engine.py
+++ b/presidio-analyzer/presidio_analyzer/analyzer_engine.py
@@ -102,6 +102,7 @@ class AnalyzerEngine:
         correlation_id: Optional[str] = None,
         score_threshold: Optional[float] = None,
         trace: Optional[bool] = False,
+        include_interpretability_response: Optional[bool] = True
     ) -> List[RecognizerResult]:
         """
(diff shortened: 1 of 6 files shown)


balteravishay (approved):

presidio-analyzer/presidio_analyzer/analyzer_engine.py
```diff
@@ -176,3 +179,18 @@ def __remove_low_scores(
 
         new_results = [result for result in results if result.score >= score_threshold]
         return new_results
+
+    def __remove_analysis_explanation(self, results: List[RecognizerResult],
```
navalev, line 183: Is it possible to not generate the interpretability info instead of removing it?

navalev (changes requested):

presidio-analyzer/presidio_analyzer/analyzer_engine.py
itye-msft: Internally, the traces use it.
So we want to keep it inside, but not always to return in via the HTTP payload.

navalev (approved):

PR #444, opened 2021-01-27
Removed wrong folder for linting

omri374:
Build python steps are using the WORKING_FOLDER param for linting, while this folder is the root of the package. Calling `flake8` on a missing folder skips linting completely. Our current code in V2 doesn't pass linting.


balteravishay (approved):

navalev (changes requested):
this fails the whole pipeline - I prefer to also fix the linting errors in the same PRs to avoid failing it

omri374:
@navalev yes definitely. I'll push a fix soon.

SharonHart (approved):

omri374:
Fixed unlinted files

PR #446, opened 2021-01-27
Initial docs structure

omri374:
Proposed docs structure. 
Visually:
image

Some remarks:
1. Added "installation" which looks at installing the ready solution (by pip or docker pull) holistically and not focusing on one service
2. Example python and docker/http usages for using multiple services (e.g., analyzer + anonymizer) are under `samples`. There's an additional layer with "samples/deployments".
3. The current README on presidio-analyzer and presidio-anonymizer folders should move into the docs. We can have a link there to the docs.
4. Yet to be finalized is how to document API references. Are we generating something automatically / pointing to swagger or writing something manually?
5. Design doc is planned to provide a bigger picture of the multi-service design, but maybe this is unnecessary.
6. Removed authors list. No actual need for this IMHO.


navalev (approved):
IMO We should minimize the main readme, but that can be done in a seperate cycle

navalev (approved):
python samples should not be under deployment IMO
As for API reference - in the gRPC version we generated (manually) swagger spec. We can do the same ADO1843 (backlog)

cobypeled (commented):
The initial idea was to have the Samples folder outside of Docs, wasn't it?
I think both are fine.
To Nava's point I agree with it would make more sense to have a folder for samples that will include deployment.

navalev (approved):

omri374:
@cobypeled @navalev thanks for the review.
Created another layer of samples. Now it contains Python, Docker and deployments where we currently have App Service.

@cobypeled did a check and we have to contain the samples within docs if docs are referencing to it, otherwise we'll have broken links on microsoft.github.io/presidio

cobypeled (approved):

cobypeled (approved):

navalev (approved):

PR #448, opened 2021-01-28
Add replace anonymizer

shiranr:
Add replace anonymizer 
Make the anonymizers eager loaded instead of lazy.


navalev (approved):

presidio-anonymizer/presidio_anonymizer/entities/anonymizer_request.py
```diff
@@ -106,12 +107,12 @@ def __get_anonymizer(self, transformation):
         Extract the anonymizer class from the anonymizers list.
 
         :param transformation: a single transformation value
-        :return: anonymizer_class
+        :return: anonymizer
```
SharonHart, line 110: if you've dropped the reference for it being a class, add it as type hint

shiranr: Done.

navalev (approved):

SharonHart (approved):

PR #453, opened 2021-01-28
CI - Deploy to dev environment

balteravishay:
This PR adds:
* split pipeline to stages: build&test and deploy
* deploy to app service in B/G deployment: create slot, deploy to slot, run smoke test on slot, swap slot if successful, delete slot (always).
* use variable group for pipeline instead of singular variables.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,213 +1,363 @@
 # Presidio PR validation and CI pipeline.
 # * PR Validation: security analysis, lint, unit test, functional tests
-# * CI: lint, unit test, functional tests, push to private Azure Container registry 
+# * CI: lint, unit test, functional tests, push to private Azure Container registry
 #       and deploy to internal development environment
 
 trigger:
   branches:
     include:
-    - V2
+    - V2s
 
 pr:
   branches:
     include:
       - V2
 
-jobs:
-    #- job: validate
-    #  displayName: 'Validate PR pre-requisites'
-    #  steps:
-    #    - template: .pipelines/templates/validate-version.yml
-    #      parameters:
-    #        UPSTREAM_BRANCH: $(System.PullRequest.TargetBranch)
-
-    - job: Inclusivelint
-      displayName: 'Inclusive linting validation'
-      steps:
-        - task: Bash@3
-          displayName: 'inclusive lint'
-          inputs:
-            targetType: 'inline'
-            script: |
-              npm install inclusivelint
-              npx inclusivelint -r -p .
-    
-    - job: SecurityAnalysis
-      displayName: Security Analysis
-      condition: ne(variables['Build.Reason'], 'IndividualCI')
-      pool:
-          # CredScan only runs on Windows
-          vmImage: 'windows-latest'
-      steps:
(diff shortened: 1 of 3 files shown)


navalev (dismissed):

azure-pipelines.yml
```diff
@@ -1,213 +1,347 @@
...
+        - job: WaitForDeployment
+          displayName: Wait for deployment
+          dependsOn: 
+            - DeployAnalyzerToDev
+            - DeployAnonymizerToDev
+          pool: Server
+          steps:
+          - task: Delay@1
+            inputs:
+              delayForMinutes: '5'
```
itye-msft, line 296: Isn't there any other way?
What if it won't finish in 5 minutes?

azure-pipelines.yml
```diff
@@ -1,213 +1,347 @@
...
+          dependsOn: 
+            - DeployAnalyzerToDev
+            - DeployAnonymizerToDev
+            - WaitForDeployment
+          variables:
+            ANALYZER_STAGING_URI: $[ dependencies.DeployAnalyzerToDev.outputs['createStagingSlot.STAGING_URI'] ]  
+            ANONYMIZER_STAGING_URI: $[ dependencies.DeployAnonymizerToDev.outputs['createStagingSlot.STAGING_URI'] ]  
+          steps:
+          - task: Bash@3
+            displayName: 'Smoke test'
```
itye-msft, line 309: Can this depend on external file, because if any changes to the interface occurs, you need to change the CI!

itye-msft (changes requested):
Looks good. left some comments

azure-pipelines.yml
balteravishay: i wish there is, and would take any suggestion.
the problem is that its an app service pulling and running the container. there is no "status" or "readiness" you can query.
the only workaround which may fit what you want is to put the smoke tests in a a while loop. 
do you think that's better?

azure-pipelines.yml
balteravishay: great comment! i removed the specific requests and switched to using /health endpoint of both services. (had to add it to anonymizer as well).

azure-pipelines.yml
navalev: how about adding "semi" smoke test in a loop - curl to a health check to each service?
question - can we deploy the apps using ARM templates and wait for an output from the template (say have an output variable with the slot id)?

balteravishay: @navalev the ARM template deployment exists after the app service is up and does not wait for the container to pull. there is also no "readiness" that i could find that indicates if the container is ready or not. i will do the following:
* implement a while loop on the smoke test instead of delay, with an exit of 5 minutes.
* ask a question in the relevant CSE tech domain

navalev: so we're going with the while loop (until we get a 200)

navalev (approved):

azure-pipelines.yml
balteravishay: done

navalev (approved):

itye-msft (approved):
LGTM

PR #455, opened 2021-01-30
Anonymizer - Integration Tests Infra + Mask scenarios

SharonHart:
This adds an infrastructure for anonymizer integration tests.
While 'integration' terminology was used before in the functional-tests context (where we test integration **between presidio services** through API), here, the integration is **between anonymizer components** (like engine-anonymizers, with as less mocks as possible).
Why is the functional-tests module not sufficient? Since presidio services can be used as packages, functionality can be extended without exposing new REST APIs.
In return, this additional test layer can reduce the coverage of the functional-tests module, to be mainly on the REST API, with simple and short happy scenarios and negative tests for the different http codes. 

Implementation-wise, each tested entry-point (method) will have a directory under tests/resources. In it, for each method scenario, there's an input and output json files. All the scenarios will asserted in a single parameterized test.
Not sure that files are the best option as the methods receives objects, not jsons (True vs true, None vs null) but currently, and for anonymize method this is an easily extendible option.

### DIFF
--- a/presidio-anonymizer/tests/integration/file_utils.py
+++ b/presidio-anonymizer/tests/integration/file_utils.py
@@ -1,0 +1,6 @@
+import os
+
+
+def get_scenario_file_content(scenario_method, scenario_name: str):
+    with open(os.path.join("resources", scenario_method, scenario_name)) as f:
+        return f.read()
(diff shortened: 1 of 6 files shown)


omri374 (dismissed):

omri374:
Looks great.

navalev (dismissed):
looks good. still debating weather this shoul be part of the full integration tests, or leave it in each package seperatly (and we need to add the same for the analyzer). for now lets continue with your approach

navalev (approved):

omri374 (approved):

PR #462, opened 2021-02-01
Functional Tests Refactor

SharonHart:
Now that we have several scenarios, decided to align all of them with a single structure and use common methods.
Some decisions:
* Use json strings for request and response. (@itye-msft your hunch was right). Both easy copy from postman, and no need to transform the payload to python dict (null -> None, true > True). 
* Place the jsons in the scenario, for easy read-in-one-go, and compare the request and the response easily. This has the assumption that our payloads should be fairly small, and that we focus on the API layer, where complex scenarios will be added to the service integration test. @shiranr I left some of your tests reading from files, I want your approval to transform them. 
* Added json strings equality helper methods so that we won't have to sort lists and objects in the scenarios.


.pre-commit-config.yaml
```diff
@@ -14,7 +14,7 @@ repos:
           ]
           args: ['--max-line-length=88', 
                  '--docstring-convention=numpy', 
-                 '--per-file-ignores=**/tests/**.py:D functional-tests/**:D', 
+                 '--per-file-ignores=**/tests/**.py:D functional-tests/**:D,E501,W291',
```
omri374, line 17: Should this be here or under `--extend-ignore`?

functional-tests/tests/test_analyzer.py
```diff
@@ -1,76 +1,141 @@
...
+    request_body = "{}"
 
     response_status, response_content = analyze(request_body)
 
+    expected_response = """
+        {"error": "No text provided"}
+    """
     assert response_status == 500
-    assert response_content == {"error": "No text provided"}
+    assert equal_json_strings(expected_response, response_content)
```
omri374, line 79: Could we have issues with json strings formatting such as indentation between expected and actual?

SharonHart: Json.loads takes care of the whitespaces and aligns both.

.pre-commit-config.yaml
SharonHart: This links the ignore to just the functional tests

omri374: ok

omri374 (approved):

navalev (approved):
looks good. Approving following our offline discussion. We need to align all tests thoughot the code - unit, funcitonal and integration (as much as possible)

navalev (approved):

SharonHart:
@navalev @omri374 I've committed another change moving some of the scenarios (that were jsons in the first place) from functional to anonymizer integrations. @shiranr those are yours do you agree with moving?

.pre-commit-config.yaml
```diff
@@ -14,7 +14,7 @@ repos:
           ]
           args: ['--max-line-length=88', 
                  '--docstring-convention=numpy',
```
shiranr, line 16: Can we add a remark what each means?

.pre-commit-config.yaml
shiranr, line 17: Can we add a remark what each means or a link to where you got this codes from? 
When someone reads the code he/she starts looking it online and it takes time.

SharonHart: Done

functional-tests/tests/test_anonymizer.py
```diff
@@ -1,92 +1,63 @@
...
 
 
 @pytest.mark.api
-def test_anonymize_api_fails_on_invalid_value_of_text():
-    json_path = file_path("payload.json")
-    with open(json_path) as json_file:
-        content = json.load(json_file)
-        content["text"] = ""
+def test_given_anonymize_called_with_empty_text_then_invalid_input_message_returned():
+    request_body = """
```
shiranr, line 34: Next test is with json files and this one is without I would keep it consistent. Take one valid json and explicitly do: json.text = ""
I think it will be more clear and use the same concept of json files.

functional-tests/tests/test_anonymizer.py
```diff
@@ -1,92 +1,63 @@
...
 
+    expected_response = """
+        ["mask", "fpe", "replace", "hash", "redact"]
+    """
 
-@pytest.mark.api
-def test_anonymizers():
-    expected_list = ["mask", "fpe", "replace", "hash", "redact"]
-    response_status, response_content = anonymizers()
     assert response_status == 200
```
shiranr, line 62: Very cool change

functional-tests/tests/test_anonymizer.py
```diff
@@ -1,92 +1,63 @@
...
-    assert all([a == b for a, b in zip(response_content, expected_list)])
```
shiranr, line 92: Very nice change. :)

shiranr (approved):

navalev (approved):

PR #465, opened 2021-02-02
Split CI and PR pipelines

SharonHart:
As CI pipeline is becoming more complex, this splits the pipelines, reducing the number of conditions in the pipeline steps.
* To reduce the sizes of both, extracted linting and building the services into templates as a single stage for reuse in both CI and PR pipelines. 
* Made Functional tests not dependent in the services build jobs, reduces the build time from 11 to 6 minutes, ultimately releasing build agents faster which we need, yet not 'classic CI' where we run Build -> UT -> functional 
*  Regarding CI pipeline I've currently just left the dev deployment as-is, not extracting into a template. I think that it's worth waiting for the production deployment and see what is common and what is not.


navalev (dismissed):
Looks good. as we discussed, I'm in favor of spliting the pipelines. Not sure about the naming and location of all the pipeline yml files, but this can be addressed in a seperate PR

SharonHart:
@navalev  eventually I want to move everything to /pipelines WDYT?

navalev:
@SharonHart yes, agree all can go under pipelines folder. also would change the names to somthing like  ci-pipeline and pr-pipeline

balteravishay (approved):
lgtm

navalev (approved):

PR #467, opened 2021-02-03
V1: Removed pyre2 dependency

omri374:
Fix showstopper bug on V1 before migrating to V2

### DIFF
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -69,6 +69,5 @@ Presidio leverages:
 - [spaCy](https://spacy.io/)
 - [Redis](https://redis.io/)
 - [GRPC](https://grpc.io)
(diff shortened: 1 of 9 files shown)


shiranr (approved):

docs/overview.md
```diff
@@ -69,6 +69,5 @@ Presidio leverages:
 - [spaCy](https://spacy.io/)
 - [Redis](https://redis.io/)
 - [GRPC](https://grpc.io)
```
balteravishay, line 71: let's remove these two as well?

balteravishay (approved):
approve with minor comment

docs/overview.md
omri374: Why? they are still used in V1

PR #468, opened 2021-02-03
Install NLP models in Docker

omri374:
Added a Python script which reads a conf file (e.g. conf/default.yaml) and downloads the defined models during Docker build.


presidio-analyzer/install_nlp_models.py
```diff
@@ -0,0 +1,58 @@
...
+            if not stanza:
+                raise ValueError("the Stanza package should be installed manually")
+            else:
+                stanza.download(model["model_name"])
+        else:
+            raise ValueError(
+                f"Unsupported nlp engine: {nlp_configuration['nlp_engine_name']}"
+            )
+    logger.info("finished installing models")
+
```
shiranr, line 44: The if else here is extremely confusing. how about a switch case?

shiranr (commented):
One comment, LMK what you think :)

presidio-analyzer/Dockerfile
```diff
@@ -1,15 +1,18 @@
 FROM python:3.8-slim
 
 ARG NAME
-ARG SPACY_MODEL=en_core_web_lg
+ARG NLP_CONF_FILE=conf/default.yaml
```
navalev, line 4: I would not change the dockerfile, but instead update the documetation with a sample of how to change it if a user wants to install NLP model in docker. WDYT?

navalev (changes requested):

presidio-analyzer/Dockerfile
omri374: so you mean we wouldn't have an argument here? this could be useful if we want to build it with other versions for testing

presidio-analyzer/install_nlp_models.py
omri374: I agree this is confusing, but there is no switch case in Python. How would you suggest to do it?

omri374: Let me refactor a little bit and see if you find it more readable.

navalev: how about using and array or lambda, somthing like (this will not work, just quick pseudo code)
```
nlp_modles = {
"spacy" : lambda s:  spacy.cli.download(model["model_name"]),
"stanza": lambda s:  stanza.download(model["model_name"])
}

nlp_models.get(nlp_configuration["nlp_engine_name"])
```

presidio-analyzer/Dockerfile
navalev: I would opt for just using the default, which is already configured in the code, so no need to add the config file.

navalev: On second thought, let's leave it in the dockerfile, will make it easier for users to use the language config feature by just updating the config file and rebuilding the image. I'm resoving the comment

presidio-analyzer/install_nlp_models.py
omri374: Thanks @navalev, this is a great suggestion. I did something a little different before seeing your suggestion. If you still feel it's cumbersome I could definitely refactor again.

presidio-analyzer/Dockerfile
omri374: These were my thoughts. Maybe it deserves more documentation to make sure users understand how to use it?

balteravishay (approved):
lgtm

presidio-analyzer/install_nlp_models.py
omri374: @shiranr could you please review if the new code makes more sense?

navalev: @omri374 looks good to me :)

navalev (approved):

PR #469, opened 2021-02-03
Configurable logging

omri374:
This PR adds:
1. logging configuration file (logging.ini) to the analyzer and anonymizer
2. Updates to the app.py to read from logging.ini
3. package usage (not through flask) is not emitting any logs, except for interpretability logs
4. Interpretability renamed to decision-process
5. Interpretability/decision-process is loaded in `__init__.py` because we do want it to work even on a package usage, if the user asked for `trace_pii=True` on the analyzer request.
6. Removed `PresidioLogger` as we want to handle everything through the config file and not via code.
7. default log level is INFO, can be changed via env variable.


presidio-analyzer/presidio_analyzer/__init__.py
```diff
@@ -14,10 +13,23 @@
 from presidio_analyzer.analyzer_engine import AnalyzerEngine
 from presidio_analyzer.analyzer_request import AnalyzerRequest
 
+# Define default loggers behavior
+
+# 1. presidio_analyzer logger
+
 logging.getLogger("presidio-analyzer").addHandler(logging.NullHandler())
```
balteravishay, line 20: shouldn't there be a way to change this behavior? will users of the package never want logging or just by default?

balteravishay (commented):
just one questions. looks good!

presidio-analyzer/presidio_analyzer/__init__.py
omri374: They can use the same logic as we have in app.py. We can have an if/else here with some constant but I don't know if it'll make it easier to work with. WDYT?

balteravishay: if the client of the package can override this behavior as we do in app.py, without changing this code than its ok. if they have to change this code to allow it to log, i think it can be better. approving.

balteravishay (approved):
approve with comment

presidio-analyzer/presidio_analyzer/__init__.py
omri374: Thanks. They would have to write code (toconfigure logging) but not to change existing code. Hope this is what you mean.

presidio-analyzer/logging.ini
```diff
@@ -0,0 +1,27 @@
...
+propagate=0
+
+[handler_consoleHandler]
+class=StreamHandler
+level=INFO
+formatter=simpleFormatter
+args=(sys.stdout,)
+
+[formatter_simpleFormatter]
+format=%(asctime)s - %(name)s - %(levelname)s - %(message)s
```
SharonHart, line 27: Needs checking that this is parsed correctly in app service logs

presidio-analyzer/presidio_analyzer/__init__.py
```diff
@@ -14,10 +13,23 @@
...
 
+# 2. decision_process logger.
+# Setting the decision process trace here as we would want it
+# to be activated using a parameter to AnalyzeEngine and not by default.
+
+decision_process_logger = logging.getLogger("decision_process")
+ch = logging.StreamHandler()
+formatter = logging.Formatter("[%(asctime)s][%(name)s][%(levelname)s]%(message)s")
```
SharonHart, line 28: Can this be taken from the toml definition?

presidio-analyzer/presidio_analyzer/app_tracer.py
```diff
@@ -12,9 +12,7 @@ class AppTracer:
     """
 
     def __init__(self, enabled: bool = True):
-
-        self.logger = PresidioLogger("Interpretability")
-        self.logger.set_level("INFO")
+        self.logger = logging.getLogger("decision_process")
```
SharonHart, line 15: should the rename be in the api as well? remove_interpretability_response attribute?

presidio-analyzer/presidio_analyzer/presidio_logger.py
```diff
@@ -1,86 +0,0 @@
-import logging
-import os
-import sys
-
-
-class PresidioLogger:
```
SharonHart, line 6: nice decision on this 👍

presidio-anonymizer/app.py
```diff
@@ -31,28 +40,26 @@ def anonymize():
...
+            )
             return ErrorResponse(err.err_msg).to_json(), 422
 
         @self.app.errorhandler(Exception)
         def server_error(e):
-            self.logger.error(
-                "A fatal error occurred "
-                "during execution".format(e)
-            )
+            self.logger.error("A fatal error occurred " "during execution".format(e))
```
SharonHart, line 62: missing e placeholder?

presidio-analyzer/app.py
```diff
@@ -26,7 +30,9 @@ class Server:
     """HTTP Server for calling Presidio Analyzer."""
 
     def __init__(self):
-        self.logger = PresidioLogger(os.environ.get("PRESIDIO_LOGGER"))
+        fileConfig(Path(Path(__file__).parent, LOGGING_CONF_FILE))
+        self.logger = logging.getLogger("presidio-analyzer")
+        self.logger.setLevel(os.environ.get("LOG_LEVEL", self.logger.level))
```
SharonHart, line 35: Is there an attribute for reading that in the logging conf file?

presidio-analyzer/logging.ini
omri374: Do we know if it does? We can update this later if we see it is not parsed correctly

SharonHart (approved):

presidio-analyzer/app.py
omri374: Discussed offline, no quick and easy solution.

presidio-analyzer/presidio_analyzer/app_tracer.py
omri374: Yes, we have a user story for this.

presidio-anonymizer/app.py
omri374: Not sure I follow?

presidio-analyzer/app.py
omri374: This is for when the log level is assigned in an env var. This would take precedence over the conf file

presidio-analyzer/presidio_analyzer/__init__.py
omri374: Discussed offline, no quick and easy solution.

omri374:
@balteravishay and @SharonHart could you please re-approve? had a merge conflict.

SharonHart (approved):

presidio-anonymizer/app.py
SharonHart: e is not printed, not something you changed though

omri374: Fixed

SharonHart (approved):

balteravishay (approved):

anatampa:
How can one change the default logging level via the env variable? Any help would be appreciated.

omri374:
For Python or Docker?

PR #470, opened 2021-02-04
return json response from API

shiranr:
Move error handling into handlers.
return json from API

### DIFF
--- a/functional-tests/tests/test_anonymizer.py
+++ b/functional-tests/tests/test_anonymizer.py
@@ -24,7 +24,8 @@ def test_given_anonymize_called_with_valid_request_then_expected_valid_response_
 
     response_status, response_content = anonymize(request_body)
 
-    expected_response = "hello world, my name is ANONYMIZED. My number is: 03445****"
+    expected_response = '{"text":"hello world, my name is ANONYMIZED. ' \
+                        'My number is: 03445****"}\n'
     assert response_status == 200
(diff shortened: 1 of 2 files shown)


balteravishay (approved):
nice!

functional-tests/tests/test_anonymizer.py
```diff
@@ -24,7 +24,8 @@ def test_given_anonymize_called_with_valid_request_then_expected_valid_response_
 
     response_status, response_content = anonymize(request_body)
 
-    expected_response = "hello world, my name is ANONYMIZED. My number is: 03445****"
+    expected_response = '{"text":"hello world, my name is ANONYMIZED. ' \
```
navalev, line 27: are'nt we using "response" and not "text"?

presidio-anonymizer/app.py
```diff
@@ -57,6 +49,17 @@ def anonymizers() -> Tuple[str, int]:
                 )
                 return ErrorResponse(e.args[0]).to_json(), 500
```
navalev, line 50: why not add self.app.errorhandler also here?

omri374 (approved):

functional-tests/tests/test_anonymizer.py
shiranr: Correct, will be fixed.

PR #484, opened 2021-02-07
Publish Anonymizer and Analyzer to PyPI

SharonHart:
Used the legacy pipeline template as reference:
https://github.com/microsoft/presidio/blob/master/pipelines/templates/publish-python-twine.yaml
Mainly, removed ADO artifacts publishing, which simplified the pipeline by removing conditions and most of the parameters.

Pass analyzer and anonymizer working dirs as steps arguments


.pipelines/templates/release.yml
```diff
@@ -27,6 +28,17 @@ jobs:
         tagSource: manual
         tag: $(version)
 
+- job: PublishToPyPi
+  displayName: Publish to PyPi
+  dependsOn: version
+  steps:
+    - template: ./publish-to-pypi.yml
```
balteravishay, line 35: can be a matrix and then run in paralllel.

.pipelines/templates/publish-to-pypi.yml
```diff
@@ -0,0 +1,37 @@
...
+    inputs:
+      pythonUploadServiceConnection: Pypi-Presidio
+  - task: Bash@3
+    displayName: 'Upload Wheel to PyPi'
+    inputs:
+      targetType: 'inline'
+      script: |
+        set -eux  # fail on error
+        echo ${PYPIRC_PATH}
+        ls -l
```
balteravishay, line 35: ls -l: intentional?

balteravishay (approved):
minor comments

.pipelines/templates/publish-to-pypi.yml
SharonHart: no, so is the echo, good catch

SharonHart:
@balteravishay Done, thanks

balteravishay (approved):

navalev (approved):
LGTM

PR #485, opened 2021-02-07
Dynamicly load anonymizers

navalev:
minor fix to load list of anonymizers dynamicly - load all subclasses of the Anonymizer class instead of using a hard coded list.


presidio-anonymizer/tests/test_anonymizer_request.py
```diff
@@ -1,138 +1,138 @@
-from typing import List
```
navalev, line 1: just a lint change

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
```diff
@@ -13,14 +13,14 @@ class AnonymizerEngine:
...
+
     logger = logging.getLogger("presidio-anonymizer")
-    builtin_anonymizers = {
-        "mask": Mask,
-        "fpe": FPE,
-        "replace": Replace,
-        "hash": Hash,
-        "redact": Redact,
-    }
+    builtin_anonymizers = __get_sub_classes()
```
SharonHart, line 23: Nice! 👍 
Do you think that there's a place to be more implicit, and an additional interface method, like get_display_name? In case we want to pass a different name in the API, and that the sender will be more aware of the connection between the method class and and api method?

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
```diff
@@ -13,14 +13,14 @@ class AnonymizerEngine:
     and replaces the PII entities with the desired transformations.
     """
 
+    def __get_sub_classes():
+        sub_classes = {}
+        for cls in Anonymizer.__subclasses__():
+            sub_classes[cls.__name__.lower()] = cls
```
SharonHart, line 19: Not critical, you can also use comprehension: 
{cls.__name__.lower(): cls for cls in Anonymizer.__subclasses__()}

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
navalev: yes, that's a good idea. I'll do a quick update

SharonHart (approved):
Awesome 😄

balteravishay (approved):

presidio-anonymizer/presidio_anonymizer/anonymizers/hash.py
```diff
@@ -36,5 +36,9 @@ def validate(self, params: dict = None) -> None:
         )
         pass
 
+    def anonymizer_name(self) -> str:
```
omri374, line 39: Could be a parameter on the abstract class:
```python
self.name = self.__class__.__name__.lower()
```

SharonHart: See my comment above, we wanted to implicitly state the name, WDYT?

navalev: I'm debating between this and a class method, this does seems more elegant :)

SharonHart: meaning this as default implementation and have the ability to override in the implementing class?

omri374: Yes, this way it is both configurable but also works out of the box

PR #486, opened 2021-02-07
Remove flask from presidio python packages

SharonHart:
setup.py's should not install_require flask as this is the non applicative way of using presidio


omri374 (approved):

balteravishay:
shouldn't it be documented that when doing pip install presidio-xxx it won't work as an http server? is this behaviour expected?

SharonHart:
That's expected, the app.py defining is actually outside of the package directory

navalev (approved):

PR #488, opened 2021-02-07
anonymizer engine lite refactor

navalev:
a bit of tidying up to the anonymizer engine anonymize method


presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
```diff
@@ -75,3 +81,28 @@ def __validate_position_over_text(self, analyzer_result, text_len):
...
+        self, anonymizer, output_text, analyzer_result, transformation
+    ):
+        text_to_anonymize = output_text[analyzer_result.start : analyzer_result.end]
+        anonymized_text = anonymizer.anonymize(
+            params=transformation, text=text_to_anonymize
+        )  # TODO: [ADO-2754] replace with the singleton class instance
+        return anonymized_text
+
+    def __update_output_with_anonymized_section(
+        self, analyzer_result, last_replacement_point, output_text, anonymized_text
```
shiranr, line 101: I thought about doing this, but think the method is having too many parameters. 
Perhaps moving it to a class which handles the parameters will be more correct but I think it's not such a small refactor.

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
```diff
@@ -75,3 +81,28 @@ def __validate_position_over_text(self, analyzer_result, text_len):
...
+        anonymizer.validate(params=transformation)
+        return anonymizer
+
+    def __anonymize_section(
+        self, anonymizer, output_text, analyzer_result, transformation
+    ):
+        text_to_anonymize = output_text[analyzer_result.start : analyzer_result.end]
+        anonymized_text = anonymizer.anonymize(
+            params=transformation, text=text_to_anonymize
+        )  # TODO: [ADO-2754] replace with the singleton class instance
```
shiranr, line 97: There was a huge conversation about this one...
including 
@omri374 @itye-msft @SharonHart and myself. I think this should be decided.

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
```diff
@@ -75,3 +81,28 @@ def __validate_position_over_text(self, analyzer_result, text_len):
...
                 f"original text length is only {text_len}."
             )
+
+    def __extract_anonymizer(self, transformation):
+        anonymizer = transformation.get("anonymizer")()
+        # if the anonymizer is not valid, a InvalidParamException
+        anonymizer.validate(params=transformation)
+        return anonymizer
+
+    def __anonymize_section(
```
shiranr, line 91: I agree it should have its own method but it is really overloaded in the parameters.

shiranr (dismissed):
Small comments, not sure they need to be fixed in this PR.

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
```diff
@@ -42,25 +42,31 @@ def anonymize(self, engine_request: AnonymizerRequest) -> str:
...
+
+        # loop over each analyzer result
+        # get anonymizer type class for the analyzer result (replace, redact etc.)
+        # trigger the anonymizer method on the section of the text
+        # perform the anonymization
+        # concat the anonymized string into the output string
         for analyzer_result in analyzer_results:
+            self.__validate_position_over_text(analyzer_result, text_len)
+
             transformation = engine_request.get_transformation(analyzer_result)
```
omri374, line 54: What do you think about calling the `get_transformation` function with the entity type and not the entire analyzer_result? Something like
```python
transformation = engine_request.get_transformation(analyzer_result.entity_type)
```
or
```python
entity_type = analyzer_result.entity_type
transformation = engine_request.get_transformation(entity_type)
```

Might make the code a bit more readable. Just a suggestion though.

navalev: done

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
navalev: my intention was to leave the main loop more readable, and keep the helper methods in the class to make it easier to follow up when reading the code

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
navalev: removed the comment - the decision was not to use singletons. But we need to make sure not to re-create the anonymizer on each request

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
navalev: agree, the purpuse is to make the main loop more readable

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
```diff
@@ -75,3 +79,39 @@ def __validate_position_over_text(self, analyzer_result, text_len):
...
+        self, analyzer_result, last_replacement_point, output_text, anonymized_text
+    ):
+        end_of_text = min(analyzer_result.end, last_replacement_point)
+        return (
+            output_text[: analyzer_result.start]
+            + anonymized_text
+            + output_text[end_of_text:]
+        )
+
+    def __get_transformation_from_analyzer_result(
```
omri374, line 108: nit: If this only calls another method and writes a log, maybe we don't need an extra level here and this could stay in the main `anonymize` function?

presidio-anonymizer/presidio_anonymizer/entities/anonymizer_request.py
```diff
@@ -26,23 +24,25 @@ def __init__(self,
...
         self.__validate_and_insert_input(data)
-        self.default_transformation = {"type": "replace",
-                                       "anonymizer": self.anonymizers["replace"]}
+        self.default_transformation = {
+            "type": "replace",
+            "anonymizer": self.anonymizers["replace"],
+        }
 
-    def get_transformation(self, analyzer_result: AnalyzerResult):
+    def get_transformation(self, entity_type: str):
```
omri374, line 32: nit: This could be private as it's only being called from private methods (if we stay with the __get_transformation_from_analyzer_result method)

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
navalev: yes, it is abit over kill :) wanted to make the code as clear as possible.

presidio-anonymizer/presidio_anonymizer/entities/anonymizer_request.py
navalev: I've reverted and removed the __get_transformation_from_analyzer_result  method, so leaving it public

omri374 (approved):

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
shiranr: I know, but I think passing over 3 parameters to a method makes it messy as well. 
That's why I didn't do the refactor in the first place... I think the best way for this is to create a class that holds the data and knows how to handle it. 
But it doesn't need to be in this review.
We can enter this and do it later.

shiranr (approved):

PR #491, opened 2021-02-08
add missing imports

shiranr:
Add analyzer as a dependency.
Fix imports


presidio-image-anonymizer/setup.py
```diff
@@ -4,6 +4,11 @@
 
 from setuptools import setup, find_packages
 
+requirements = ["flask==1.1.2", "pillow", "pytesseract==0.3.7",
```
SharonHart, line 7: flask is not a requirement for the package

presidio-image-anonymizer/Pipfile
```diff
@@ -7,6 +7,7 @@ name = "pypi"
 flask = "==1.1.2"
```
biancafbmd, line 7: Should we also add pillow here too?

presidio-image-anonymizer/setup.py
shiranr: Removed.

presidio-image-anonymizer/Pipfile
shiranr: Done.

presidio-image-anonymizer/setup.py
```diff
@@ -4,29 +4,41 @@
 
 from setuptools import setup, find_packages
 
+requirements = ["pillow", "pytesseract==0.3.7", "presidio-analyzer>=1.9.0"]
```
SharonHart, line 7: missing numpy?

presidio-image-anonymizer/setup.py
```diff
@@ -4,29 +4,41 @@
...
+
+test_requirements = ["pytest>=3", "flake8==3.7.9", "flask==1.1.2"]
```
SharonHart, line 9: drop flask here as well

presidio-image-anonymizer/setup.py
```diff
@@ -4,29 +4,41 @@
...
 __version__ = ""
-parent_directory = os.path.abspath(
-    os.path.join(path.abspath(path.dirname(__file__)), os.pardir)
-)
+this_directory = path.abspath(path.dirname(__file__))
+parent_directory = os.path.abspath(os.path.join(this_directory, os.pardir))
+
+with open(path.join(this_directory, 'README.MD'), encoding='utf-8') as f:
+    long_description = f.read()
+
```
SharonHart, line 17: you're not based

presidio-image-anonymizer/setup.py
shiranr: weird it's even here...

presidio-image-anonymizer/setup.py
shiranr: I was.

PR #495, opened 2021-02-08
Don't delete staging slots on CI deployment

SharonHart:
[ADO-2812]


balteravishay (approved):

navalev (commented):
so the deployment will override the existing slot? and will this also answer for ADO-2791? @balteravishay

SharonHart:
@navalev It will use the existing one, we think that creating one each deployment increases the deployment time since it does redundant container pull from acr

shiranr (approved):
LGTM

PR #497, opened 2021-02-09
Image anonymizer dockerafile with spacy and tesseract

navalev:
update image anonymizer - install spacy and tesseract

### DIFF
--- a/presidio-image-anonymizer/Dockerfile
+++ b/presidio-image-anonymizer/Dockerfile
@@ -4,9 +4,14 @@ ARG NAME
 ENV PIPENV_VENV_IN_PROJECT=1
 WORKDIR /usr/bin/${NAME}
 
+RUN apt-get update 
+RUN apt-get install tesseract-ocr -y 
+
 COPY ./Pipfile* /usr/bin/${NAME}/
 RUN pip install pipenv
 RUN pipenv sync
+
+RUN pipenv run python -m spacy download en_core_web_lg
 
 COPY . /usr/bin/${NAME}/
 EXPOSE ${PORT}


presidio-image-anonymizer/Dockerfile
```diff
@@ -4,10 +4,15 @@ ARG NAME
 ENV PIPENV_VENV_IN_PROJECT=1
 WORKDIR /usr/bin/${NAME}
 
+RUN apt-get update 
+RUN apt-get install tesseract-ocr -y
```
omri374, line 8: Should we fix a specific version? We are fixing `pytesseract` to a specific version. 
@biancafbmd WDYT? How stable is the tesseract API over time?

omri374:
@navalev looks good! One minor comment/question.

presidio-image-anonymizer/Dockerfile
biancafbmd: Ideally it should be >5, but I think 4.1.1 is the latest stable one and they mention that modernization of API in 5 cause some issues with compatibility with 4. We are using very basic functionality so think >4 might work as well. More details here: https://tesseract-ocr.github.io/tessdoc/Home.html

shiranr: please notice the versions which are > 5 are still in alpha
the 4.1.1 is the last official release.
I have installed on my machine the v5.0.0-alpha.20201127 and it works perfect.
I don't think it makes a lot of difference 
But we do need to fix it since it seems they are making breaking changes in v5

navalev: this installs the latest official version which is 4.0.0

navalev: @omri374 @biancafbmd is this sufficient? or is there a way to spcificly install the 4.1.1 version?

shiranr (approved):

presidio-image-anonymizer/Dockerfile
omri374: Let's keep it as is for now, and experiment with it later. Looks like the API is relatively stable.

omri374 (approved):

presidio-image-anonymizer/Dockerfile
omri374: This should work if we want to fix the version:
```suggestion
RUN apt-get install tesseract-ocr=4.1.1 -y 
```

navalev: @omri374 this one seems to fail with "Version '4.1.1' for 'tesseract-ocr' was not found" error. I think we will need to clone the repo with the relevant tag and build it in order to get a specific version. 
For now, I will leave the default version as is and we can tackle it later.

PR #500, opened 2021-02-10
ARM templates pull from MCR or private registry

balteravishay:
This PR changes the ARM templates to allow pull from MCR or from a private registry, as was before.
additionally, update the readme for analyzer, anonymizer and the deploy-to-app-service doc.


navalev (commented):
lets use this opertunity to also change the app service plan to use DV2 vms instead of the default A type?

navalev (approved):

SharonHart (approved):

PR #501, opened 2021-02-10
v2 image anonymizer API 

shiranr:
Add API parameters management and call to the engine.
Add relevant tests - not full yet.


presidio-image-anonymizer/presidio_image_anonymizer/image_anonymizer_engine.py
```diff
@@ -1,13 +1,15 @@
...
+
 from presidio_image_anonymizer.image_analyzer_engine import ImageAnalyzerEngine
 
 
 class ImageAnonymizerEngine:
     """ImageAnonymizerEngine class only supporting redaction currently."""
 
     def anonymize(
-        self, image: object, fill: Union[int, Tuple[int, int, int]] = (0, 0, 0)
+            self, image: Image, fill: Union[int, Tuple[int, int, int]] = (0, 0, 0)
```
biancafbmd, line 12: nit: change image docstring to PIL Image

presidio-image-anonymizer/app.py
```diff
@@ -21,8 +30,62 @@ def health() -> str:
...
+            if len(filling_str_split) == 1:
+                return int(filling_str_split[0])
+            return tuple(map(int, filling_str_split))
+        except Exception as e:
+            self.logger.error(
+                f"failed to color fill '{filling_str}' with error {e}")
+            raise InvalidParamException(f"Invalid color fill \'{filling_str}\'")
+
+    @staticmethod
+    def _image_to_byte_array(anonymized_image, im):
```
biancafbmd, line 83: small suggestion: pass im.format as parameter rather than entire image object

biancafbmd (dismissed):
Looks great :)

presidio-image-anonymizer/app.py
shiranr: Nice catch! thank you :) auto refactor...

presidio-image-anonymizer/tests/test_image_anonymizer_api.py
```diff
@@ -0,0 +1,50 @@
...
+    assert Server()._get_json_data(str_json) == {'color_fill': '0, 0, 1'}
+
+
+def test_given_invalid_json_string_then_we_get_an_invalid_param_exception():
+    with pytest.raises(InvalidParamException, match="Invalid json format 'not_json'"):
+        Server()._get_json_data("not_json")
+
+
+def test_given_empty_json_params_then_we_send_default_color_fill():
+    Server()._color_fill_string_to_value({}) == (0, 0, 0)
```
SharonHart, line 31: You're testing a private method.. and to call it you need the public api endpoint method, which is not suitable for testing here. It's a code smell for moving _color_fill_string_to_value and _get_json_data into different classes, having the service class deal with api related stuff only.

shiranr: There are many debates regarding testing private methods... I prefer having it tested. 
I think moving it to another class is incorrect and kind of detaching it from the context it needs. But I don't mind doing it either.

presidio-image-anonymizer/presidio_image_anonymizer/image_anonymizer_engine.py
shiranr: Done :)

biancafbmd (dismissed):

presidio-image-anonymizer/presidio_image_anonymizer/entities/api_request_convertor.py
```diff
@@ -0,0 +1,44 @@
...
+from presidio_image_anonymizer.entities import InvalidParamException
+
+logger = logging.getLogger("presidio-image-anonymizer")
+
+
+def get_json_data(data: str) -> dict:
+    try:
+        if not data:
+            return {}
+        return json.loads(data.replace("'", "\""))
```
SharonHart, line 16: Why do we need to replace here where in the other cases we didn't?

presidio-image-anonymizer/tests/test_api_request_convertor.py
```diff
@@ -0,0 +1,50 @@
...
+def test_given_empty_json_params_then_we_send_default_color_fill():
+    color_fill_string_to_value({}) == (0, 0, 0)
+
+
+@pytest.mark.parametrize(
+    # fmt: off
+    "json_params,expected_result",
+    [
+        ({"color_fill": "1"}, 1),
+        ({"color_fill": "1, 0, 1"}, (1, 0, 1))
```
SharonHart, line 39: What will happen when the length = 2, length > 3

presidio-image-anonymizer/tests/test_api_request_convertor.py
```diff
@@ -0,0 +1,50 @@
...
+    ],
+    # fmt: on
+)
+def test_given_json_params_then_we_extract_properly_color_fill(json_params,
+                                                               expected_result):
+    color_fill_string_to_value(json_params) == expected_result
```
TessFerrandez, line 45: are you missing an assert here?

presidio-image-anonymizer/tests/test_api_request_convertor.py
```diff
@@ -0,0 +1,50 @@
...
+    assert get_json_data(str_json) == {'color_fill': '0, 0, 1'}
+
+
+def test_given_invalid_json_string_then_we_get_an_invalid_param_exception():
+    with pytest.raises(InvalidParamException, match="Invalid json format 'not_json'"):
+        get_json_data("not_json")
+
+
+def test_given_empty_json_params_then_we_send_default_color_fill():
+    color_fill_string_to_value({}) == (0, 0, 0)
```
TessFerrandez, line 31: same here, are you missing an assert?

presidio-image-anonymizer/presidio_image_anonymizer/entities/api_request_convertor.py
```diff
@@ -0,0 +1,44 @@
...
+import logging
+from typing import Tuple, Union, io
+
+from PIL import Image
+
+from presidio_image_anonymizer.entities import InvalidParamException
+
+logger = logging.getLogger("presidio-image-anonymizer")
+
+
```
SharonHart, line 11: We don't have build verification in the PR for image anonymizer, do we have a story for that?
Meanwhile run lint locally, you're missing docstirng here

presidio-image-anonymizer/presidio_image_anonymizer/entities/api_request_convertor.py
```diff
@@ -0,0 +1,44 @@
...
+        return tuple(map(int, filling_str_split))
+    except Exception as e:
+        logger.error(
+            f"failed to color fill '{filling_str}' with error {e}")
+        raise InvalidParamException(f"Invalid color fill \'{filling_str}\'")
+
+
+def image_to_byte_array(anonymized_image: Image, image_format: str):
+    img_byte_arr = io.BytesIO()
+    anonymized_image.save(img_byte_arr,
```
SharonHart, line 41: Is this saving the image to the filesystem? For a demo site running for a while won't we need to cleanup? 
Can we do it without saving?

presidio-image-anonymizer/presidio_image_anonymizer/entities/api_request_convertor.py
```diff
@@ -0,0 +1,44 @@
...
+def color_fill_string_to_value(json_params: dict) -> Union[
+    int, Tuple[int, int, int]]:
+    filling_str = json_params.get("color_fill")
+    try:
+        if not filling_str:
+            return 0, 0, 0
+        filling_str_split = filling_str.split(',')
+        if len(filling_str_split) == 1:
+            return int(filling_str_split[0])
+        return tuple(map(int, filling_str_split))
```
TessFerrandez, line 32: This could return Tuple[int, ...] instead as the code does not guarantee 1 or 3 ints -- could be 2 or 4 for example

TessFerrandez (commented):
Looks good, just a nit and some tests with missing assert that I think should be fixed.

Regarding the conversation around testing private functions - I see both points, but agree with Sharon on this one - I think it the functions should definitely be tested, but we want to decouple them if we can, if for nothing else, so we don't have to go through the ordeal to have to mock the api endpoint - as we don't want to have to fire up the API for unit tests. 

I would also like to learn more at some point (offline) on the pros of testing private functions and reasonings there - to understand what cases this would be applicable for.

presidio-image-anonymizer/tests/test_api_request_convertor.py
shiranr: YES! thank you :)

presidio-image-anonymizer/presidio_image_anonymizer/entities/api_request_convertor.py
shiranr: Add flake8 run locally.

presidio-image-anonymizer/presidio_image_anonymizer/entities/api_request_convertor.py
shiranr: Since this is coming from multipart form and with -f 'data={"":""}' it is really easy to get the brackets wrong.
To save the user from frustration and tears, I prefer to insert this :)

presidio-image-anonymizer/tests/test_api_request_convertor.py
shiranr: Will check.

presidio-image-anonymizer/presidio_image_anonymizer/entities/api_request_convertor.py
shiranr: Nope :) saving it into byte array.

presidio-image-anonymizer/presidio_image_anonymizer/entities/api_request_convertor.py
shiranr: I've added a check if != 3 we throw an exception.

presidio-image-anonymizer/presidio_image_anonymizer/image_anonymizer_engine.py
shiranr: Done.

SharonHart (dismissed):

SharonHart (approved):

TessFerrandez (approved):

PR #502, opened 2021-02-10
Run E2E Integration Tests on Deployment

SharonHart:
Make e2e-tests read anonymizer and analyzer base urls from environment variables.
Pass those URLs after staging slot deployment and test with e2e integration test suite.


e2e-tests/tests/test_services_integration.py
```diff
@@ -0,0 +1,16 @@
+import pytest
+
+from common.methods import analyzer_supported_entities
+
+
+# TODO: [ADO-2763] Stub, remove and replace with real integration scenarios.
```
itye-msft, line 6: Is this comment still relevant?

.pipelines/templates/e2e-tests.yml
```diff
@@ -22,6 +32,12 @@ steps:
     - script: |
           set -eux  # fail on error
           source env/bin/activate
-          pytest
+          if [ -z ${TEST_SUITE} ]; then pytest -v; else pytest -v -m ${TEST_SUITE}; fi
       workingDirectory: e2e-tests
+      env:
+        ${{ if ne(parameters.analyzer_base_url, '') }}:
```
itye-msft, line 38: What will happen if it's empty? Should it continue? Perhaps add a trace message

itye-msft (changes requested):

e2e-tests/tests/test_services_integration.py
SharonHart: If it will be merged before your change then yes, and you will have to replace it with real scenarios

.pipelines/templates/e2e-tests.yml
SharonHart: In most cases it will empty and expected, and the env var will not be set - for the tests to get the default urls (localhost for running against containers)

itye-msft (dismissed):

azure-pipelines-ci.yml
```diff
@@ -148,6 +148,11 @@ stages:
                     echo analyzer status code is $ANALYZER_STATUS_CODE
                     echo anonymizer status code is $ANONYMIZER_STATUS_CODE
                 done
+          - template: .pipelines/templates/e2e-tests.yml
+            parameters:
+              test_suite: 'integration'
+              analyzer_base_url: $[ dependencies.DeployAnalyzerToDev.outputs['createStagingSlot.STAGING_URI'] ]
```
balteravishay, line 154: take these values from the variables:
$(ANALYZER_STAGING_URI)

balteravishay (approved):

itye-msft (approved):

navalev (approved):

PR #509, opened 2021-02-11
Supported entities doc for V2

omri374:
Reintroduced the supported entities (previously known as supported fields) doc with updates for V2


docs/supported_entities.md
```diff
@@ -1,13 +1,61 @@
...
 
-## Extending to additional PII entities
+### Global
 
-### Adding a custom PII entity
+|Entity Type | Description | Detection Method |
+| --- | --- | --- |
+|CREDIT_CARD |A credit card number is between 12 to 19 digits. <https://en.wikipedia.org/wiki/Payment_card_number>|Pattern match and checksum|
+|CRYPTO|A Crypto wallet number. Currently only Bitcoin address is supported|Pattern match, context and checksum|
+|DATE_TIME|Absolute or relative dates or periods or times smaller than a day.|Pattern match and context|
```
shiranr, line 17: we have a format and example for payment card number by wikipedia. 
Do we want to add a other examples? 
For example: I know there are many ways of transferring date time and dates. Users tend to fail often on it when there is no specific example.
It's also ok if you don't think this is the right place to put it.

docs/supported_entities.md
```diff
@@ -1,13 +1,61 @@
...
+|--- |--- |--- |
+|FIN/NRIC| A National Registration Identification Card | Pattern match and context |
+
+## Adding a custom PII entity
+
+See [this documentation](analyzer/adding_recognizers.md) for instructions on how to add a new Recognizer for a new type of PII entity.
 
 ### Connecting to 3rd party PII detectors
+
+See [this documentation](analyzer/adding_recognizers.md#creating-a-remote-recognizer) for instructions on how to implement an external PII detector for a new or existing type of PII entity.
```
shiranr, line 61: need to verify later the links are working properly :)

shiranr (approved):
looks good. two small comments.

docs/supported_entities.md
```diff
@@ -1,13 +1,61 @@
-# Presidio supported PII entities
+# PII entities supported by Presidio
 
-## Description
+Presidio contains predefined recognizers for PII entities.
+This page describes the different entities Presidio can detect and the method Presidio employs to detect those.
 
-## Table of contents
+In addition, Presidio allows you to add custom entity recognizers.
+For more information, refer to the [adding new recognizers documentation](analyzer/adding_recognizers.md).
```
TessFerrandez, line 7: same comment as @shiranr below - as this document is not there at the time of this PR

TessFerrandez (approved):

docs/supported_entities.md
omri374: DATE_TIME is recognized using a ML NER model, so there isn't a set of patterns to be checked. I'm afraid that if we put too much details here, we would have a gap between impl and doc in case we change the impl. WDYT?

PR #511, opened 2021-02-14
Add local build script

SharonHart:
Add a script starting the cluster with compose, then running the tests.


navalev:
the dockerfile build is timely (due to using the L version of spacy) - can we use the S version of spacy for the local build?

run.bat
```diff
@@ -0,0 +1,28 @@
+echo >/dev/null # >nul & GOTO WINDOWS & rem ^
+
+export REGISTRY_NAME=
```
itye-msft, line 3: Nice. I am missing instructions to populate the optional params.

itye-msft (changes requested):

run.bat
SharonHart: The export I'm doing here?
Those are just so that warnings won't show when running compose.
The REGISTRY_NAME, IMAGE_PREFIX, TAG are for the CI process and should be changed during local build.

SharonHart:
@navalev It should match the CI default model, since e2e tests might have static scores on the assertion that depend on the model size. Omri insisted on lg throughout all the project as far as I remember. Will add a comment on the times

itye-msft (approved):

.env
```diff
@@ -0,0 +1,3 @@
+REGISTRY_NAME=''
```
navalev, line 1: why is this file needed? seems to fail the tests

SharonHart: It's default env cars used in compose. They are used when pushing images so locally should default to empty string.
Adding this allowed removing exports in the script, and not printing warning when running compose.
It's should not fail the build, I think it fails from the sporadic issue

navalev (approved):
tested and pass on windows

itye-msft (approved):

PR #521, opened 2021-02-16
Add Flasgger + helathcheck for all services

SharonHart:
Decided to use the more lean framework, [flasgger](https://github.com/flasgger/flasgger), over [flask-restplus](https://github.com/noirbizarre/flask-restplus) which has maintainability issues and seems too robust for us. 
There is an option to have the swagger definition in method dcstrings, yaml files, or python dictionaries, will decide on which once we'll document the more heavy methods, but notice how even healthcheck is already pretty long.
By default, each service has a /apidocs endpoint by default showing swagger UI:
image

### DIFF
--- a/presidio-analyzer/Pipfile
+++ b/presidio-analyzer/Pipfile
@@ -10,6 +10,7 @@ tldextract = "*"
 numpy = "==1.19.3"
 flask = "==1.1.2"
 pyyaml = "*"
+flasgger = "*"
 
 [dev-packages]
 pytest = "*"
--- a/presidio-analyzer/Pipfile
+++ b/presidio-analyzer/Pipfile
@@ -10,7 +10,7 @@ tldextract = "*"
 numpy = "==1.19.3"
 flask = "==1.1.2"
 pyyaml = "*"
-flasgger = "*"
+flasgger = "==0.9.5"
 
 [dev-packages]
 pytest = "*"
--- a/presidio-image-anonymizer/app.py
+++ b/presidio-image-anonymizer/app.py
@@ -16,6 +16,17 @@ from presidio_image_anonymizer.entities.api_request_convertor import (
 
 DEFAULT_PORT = "3000"
 
+WELCOME_MESSAGE = r"""
+ _______  _______  _______  _______ _________ ______  _________ _______
+(  ____ )(  ____ )(  ____ \(  ____ \\__   __/(  __  \ \__   __/(  ___  )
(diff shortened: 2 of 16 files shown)


omri374:
Please add flasgger to the NOTICE file

presidio-analyzer/Pipfile
```diff
@@ -10,6 +10,7 @@ tldextract = "*"
 numpy = "==1.19.3"
 flask = "==1.1.2"
 pyyaml = "*"
+flasgger = "*"
```
omri374, line 13: Should we restrict the version? How robust is this package in terms of backward compatibility?

SharonHart: let's restrict, don't think they had breaking changes yet

SharonHart: Done

omri374 (approved):

presidio-image-anonymizer/app.py
```diff
@@ -20,27 +23,38 @@ class Server:
...
+            responses:
+              200:
+                description: OK
+                content:
+                  text/plain:
+                    schema:
+                      type: string
+                      example: ok
+            """
             return "ok"
```
navalev, line 45: if we're changing the healthcheck, lets return a nicer string. "Presidio Analyzer/Anonymizer/Image-Reducter Service is Up"

presidio-image-anonymizer/app.py
```diff
@@ -4,12 +4,15 @@
...
 from presidio_image_anonymizer.entities import InvalidParamException
-from presidio_image_anonymizer.entities.api_request_convertor import get_json_data, \
-    image_to_byte_array, color_fill_string_to_value
+from presidio_image_anonymizer.entities.api_request_convertor import (
+    get_json_data,
+    image_to_byte_array,
+    color_fill_string_to_value,
+)
 
 DEFAULT_PORT = "3000"
```
navalev, line 18: add a WELCOME_MESSAGE (same as in the other services)

navalev (approved):
looks good, added 2 minor comments. As discussed, for the rest of the APIs, the swagger spec will be in external yml files do minimize the noise in app.py

presidio-image-anonymizer/app.py
SharonHart: Done, currently leaving 'image anonymizer', we'll do the change in batch

navalev (approved):

PR #526, opened 2021-02-16
V2 transformation to anonymizer

shiranr:
Change transformation/s to anonymizer/s


presidio-anonymizer/presidio_anonymizer/entities/anonymizer_request.py
```diff
@@ -18,32 +18,32 @@ class AnonymizerRequest:
...
+        self.default_anonymizer = {
             "type": "replace",
             "anonymizer": self.anonymizers["replace"],
         }
 
-    def get_transformation(self, entity_type: str):
+    def get_anonymizer_dto(self, entity_type: str):
         """
-        Get the right transformation from the list.
+        Get the right anonymizer_dto from the list.
```
omri374, line 34: Why dto? Isn't it a class which holds the logic for anonymization?

presidio-anonymizer/presidio_anonymizer/entities/anonymizer_request.py
```diff
@@ -71,40 +71,43 @@ def __handle_analyzer_results(self, data):
...
 
-    def __handle_transformations(self, data):
+    def __handle_anonymizers(self, data):
         """
-        Go over the transformations and get the relevant anonymizer class for it.
+        Go over the anonymizers and get the relevant anonymizer class for it.
 
-        Inserts the class to the transformation so the engine will use it.
-        :param data: contains the text, transformations and analyzer_results
+        Inserts the class to the anonymizer so the engine will use it.
```
omri374, line 84: nit: Not directly related to this PR
```suggestion
        Inserts the requested anonymizer into the list of anonymizers so the engine will use it.
```
Is this what this this function does? A little difficult to understand based on the method name and docstring.

presidio-anonymizer/presidio_anonymizer/entities/anonymizer_request.py
navalev: This is a prep for an additonal work Shiran will do to change the anonymizer engine to receive parameters instead of dict.

navalev (approved):

presidio-anonymizer/presidio_anonymizer/entities/anonymizer_request.py
shiranr: This entire class is going to be gone soon... 
so lets leave it as is for now :)

SharonHart (approved):

PR #530, opened 2021-02-17
Add Image Redactor to CI pipeline

navalev:
1. Add image redactor to the CI pipeline (build & deploy to dev env)
2. Quick fix to the CI pipeline to run the e2e tests


SharonHart (approved):
Looks good, would see how the deployment looks in matrix, and if it's not too complex, but a nit.

navalev:
@SharonHart since we're using each step output for the swap stage, it can be complex. HAven't looked at it. we can always improve in a seperate PR

PR #534, opened 2021-02-18
Image redactor fix pipfile lock 

navalev:
bug fix in the Image redactor


SharonHart (approved):

presidio-image-redactor/Pipfile.lock
```diff
@@ -1,7 +1,7 @@
 {
```
omri374, line 1: Why the re-lock? It seems like different packages are installed but the Pipfile hasn't changed.

navalev: re-run lock on wsl to get this change and resolve the build issues in python 3.6 and 3.7

omri374 (approved):

PR #536, opened 2021-02-18
Auto generated API reference + small updates to docs

omri374:


docs/api/analyzer.md
```diff
@@ -0,0 +1,81 @@
+# Presidio Analyzer API Reference
```
omri374, line 1: Specifying explicitly as there are many classes which aren't important and would make this extremely long. WDYT?

TessFerrandez (approved):

docs/api/analyzer.md
navalev: wouldnt it be covered by the swagger spec?if it spepcifcly the python package API maybe rename the file to python_api.md?

navalev (approved):

PR #537, opened 2021-02-18
Move swagger to project root + Add Analyzer 

SharonHart:
Decided to move open api docs to the root of the project from several reasons:
- Some objects are shared between services - analyze/ results is part of  anonymize/ request.
- The decision to have all endpoints in a single page, and to avoid using another method to merge open api yaml files. 
- Leaving Swagger UI for Redoc, thus remove the need for swagger UI endpoint per service.
 
PR Objectives:

- [x]  Move existing Anonymizer api docs to the root of the project, currently all spread on the same yml file, might split later
- [x] Remove flaskgger from app.pys (will be removed from pipenvs in another iteration)
- [x] Add Analyzer endpoints.
- [x] Add Redoc base html file, later to be served.

[**DEMO**](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/microsoft/presidio/69242d49a1232e7b18357749f159a39486a8c30f/api-docs/api-docs.yml)


api-docs/api-docs.yml
```diff
@@ -0,0 +1,540 @@
...
+          description: "Result of a validation (e.g. checksum)"
+
+
+    EntityTypes:
+      description: "The PII detection type"
+      type: string
+      enum:
+        - NAME
+        - FIRST_NAME
+        - LAST_NAME
```
SharonHart, line 414: Those weren't in supported_entities.md, anything else I'm missing?

navalev: we support PRESON entity out of the box.

navalev:
can we place api-docs folder under docs? or must it be in the root?

SharonHart:
@navalev We can place it inside, yes

navalev (changes requested):
move api-docs under docs, and remove name, first_name and last_name from entity type list

SharonHart:
@navalev Why removing them? they can be passed as keys to anonymizer, they are not returned by supported entities?

navalev:
@SharonHart - IIRC only PERSON is supported out of the box (for en language), NAME etc does not and u need to add it

SharonHart:
@navalev Thanks, moved + removed all examples with NAMESs

navalev (approved):

docs/api-docs/api-docs.yml
```diff
@@ -0,0 +1,536 @@
...
+paths:
+
+  /analyze:
+    post:
+      servers:
+        - url: https://presidio-analyzer-prod.azurewebsites.net
+      tags:
+        - Analyzer
+      summary: "Analyze Text"
+      description: "Analyzes a text"
```
omri374, line 33:
```suggestion
      description: "Recognizes PII entities in a given text and returns their types, locations and score"
```

docs/api-docs/api-docs.yml
```diff
@@ -0,0 +1,536 @@
...
+
+paths:
+
+  /analyze:
+    post:
+      servers:
+        - url: https://presidio-analyzer-prod.azurewebsites.net
+      tags:
+        - Analyzer
+      summary: "Analyze Text"
```
omri374, line 32:
```suggestion
      summary: "Detect PII in text"
```

docs/api-docs/api-docs.yml
```diff
@@ -0,0 +1,536 @@
...
+                      }
+                    ]
+
+  /recognizers:
+    get:
+      servers:
+        - url: https://presidio-analyzer-prod.azurewebsites.net
+      tags:
+        - Analyzer
+      summary: "Get Recognizers"
```
omri374, line 87:
```suggestion
      summary: "Get the available PII recognizers for a given language"
```

docs/api-docs/api-docs.yml
```diff
@@ -0,0 +1,536 @@
...
+                    "DomainRecognizer", "UsLicenseRecognizer", "UsItinRecognizer"
+                ]
+
+  /supportedentities:
+    get:
+      servers:
+        - url: https://presidio-analyzer-prod.azurewebsites.net
+      tags:
+        - Analyzer
+      summary: "Get Supported Entities"
```
omri374, line 121:
```suggestion
      summary: "Get the list of PII entities Presidio-Analyzer is capable of detecting"
```

docs/api-docs/api-docs.yml
```diff
@@ -0,0 +1,536 @@
...
+                  "UK_NHS", "US_SSN", "US_BANK_NUMBER", "EMAIL_ADDRESS", "DATE_TIME", "IP_ADDRESS", "PERSON", "IBAN_CODE",
+                  "NRP", "US_ITIN", "DOMAIN_NAME" ]
+
+  /anonymize:
+    post:
+      servers:
+        - url: https://presidio-anonymizer-prod.azurewebsites.net
+      tags:
+        - Anonymizer
+      summary: "Anonymize Text"
```
omri374, line 151:
```suggestion
      summary: "Anonymize text using one of the supported anonymization techniques"
```

docs/api-docs/api-docs.yml
```diff
@@ -0,0 +1,536 @@
...
+          type: string
+          description: "The context word which helped increase the score"
+        validation_result:
+          type: number
+          format: double
+          description: "Result of a validation (e.g. checksum)"
+
+
+    EntityTypes:
+      description: "The PII detection type"
```
omri374, line 408: Do we want a hard coded list here?

SharonHart: If supported out of the box, I think that yes.

docs/api-docs/api-docs.yml
SharonHart: It's better to have the summary described as closest to the endpoint, as it's service as a navigator when collapsed:
image
image
WDYT?

omri374: Sure, so let's be more explicit on the description.

docs/api-docs/api-docs.yml
omri374: It is supported out of the box, but dynamic. Sort of like something that comes out of a db table (IMO)

SharonHart: Do you want to add to the description that the list the dynamic and full list can be fetched from get entities endpoint?

omri374 (approved):

PR #540, opened 2021-02-21
Remove flasgger from pipenvs

SharonHart:
Remove flasgger from pipenv dependencies, after the decision not to use it in https://github.com/microsoft/presidio/pull/537
Also reorganized analyzer dockerfile to install language models only if changed for faster docker build.


presidio-image-redactor/Pipfile.lock
```diff
@@ -687,54 +701,6 @@
...
-        },
-        "importlib-metadata": {
-            "hashes": [
-                "sha256:ace61d5fc652dc280e7b6b4ff732a9c2d40db2c0f92bc6cb74e07b73d53a1771",
-                "sha256:fa5daa4477a7414ae34e95942e4dd07f62adf589143c875c133c1e53c4eff38d"
-            ],
-            "markers": "python_version < '3.8' and python_version < '3.8'",
-            "version": "==3.4.0"
-        },
-        "numpy": {
```
navalev, line 699: how was the piplock created? I see that numpy was removed, it will fail the image-redactor

navalev (changes requested):

presidio-image-redactor/Pipfile.lock
SharonHart: Removed the lockfile, did pipenv lock
numpy is showing in a different place, but not sure why it was removed from here, since I doubt that flasgger depends on it.
Anyway ran docker-compose and saw that the image redactor started successfully.

omri374 (approved):

navalev (approved):

PR #543, opened 2021-02-22
V2 image redactor e2e

shiranr:
Add e2e tests for image redactor


e2e-tests/common/methods.py
```diff
@@ -33,3 +39,30 @@ def analyzer_supported_entities(data):
         f"{ANALYZER_BASE_URL}/supportedentities?{data}", headers=DEFAULT_HEADERS
     )
     return response.status_code, response.content
+
+
+def image_redactor(file_name: str, color_fill=None):
```
SharonHart, line 44: would rename to 'redact' to match to the endpoint name, align with the other methods

e2e-tests/common/methods.py
```diff
@@ -33,3 +39,30 @@ def analyzer_supported_entities(data):
...
+    multipart_form_data = get_multipart_form_data(file_name)
+    payload = get_payload(color_fill)
```
SharonHart, line 46: get_redact_payload, you can inline it as well since its short and will be used only here...

e2e-tests/tests/test_image_redactor.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def test_given_image_with_tuple_color_filling_then_we_get_redact_image():
+    # purple redact
+    expected_result_image = get_file("purple_redact.png")
+    response = image_redactor("original_image.png", "102, 0, 102")
+    validate_and_close(expected_result_image, response)
+
+
+def test_given_image_with_no_color_filling_then_we_get_redact_image():
```
SharonHart, line 20: I think that one happy scenario is enough if we have those tests inside the service

e2e-tests/tests/test_image_redactor.py
```diff
@@ -0,0 +1,44 @@
...
+    expected_response = '{"error": "Invalid parameter, please insert image data"}'
+    response = image_redactor("")
+    assert response.status_code == 422
+    assert response.content.decode() == expected_response
+
+
+def get_file(file_name: str):
+    current_dir = os.path.dirname(__file__)
+    file_path = os.path.join(current_dir, "..", "resources", file_name)
+    return open(file_path, "rb")
```
SharonHart, line 38: You can use the same method here for the request, and pass the content to the 'methods utils', making it just handle the methods and not read from files, avoiding duplication as well.

SharonHart: Also if we know that the image length in bytes is not too large and stable on running the tests locally and in CI, you can read the content and close the file in this method, using 'with', and not depend on other method closing the file (than one might forget when adding a new scenario that's not using validate_and_close)

e2e-tests/tests/test_image_redactor.py
```diff
@@ -0,0 +1,44 @@
...
+    expected_result_image = get_file("black_redact.png")
+    response = image_redactor("original_image.png")
+    validate_and_close(expected_result_image, response)
+
+
+def test_given_no_image_then_we_fail():
+    # black redact
+    expected_response = '{"error": "Invalid parameter, please insert image data"}'
+    response = image_redactor("")
+    assert response.status_code == 422
```
SharonHart, line 31: That should give 400 IMO

e2e-tests/tests/test_image_redactor.py
SharonHart: You are missing @api marks on all tests

e2e-tests/common/methods.py
```diff
@@ -1,10 +1,16 @@
...
-from common.constants import ANONYMIZER_BASE_URL, ANALYZER_BASE_URL
+
+from common.constants import ANONYMIZER_BASE_URL, ANALYZER_BASE_URL, \
+    IMAGE_REDACTOR_BASE_URL
 
 DEFAULT_HEADERS = {"Content-Type": "application/json"}
+MULTIPART_HEADERS = {"Content-Type": "multipart/form-data"}
 ANALYZER_BASE_URL = os.environ.get("ANALYZER_BASE_URL", ANALYZER_BASE_URL)
 ANONYMIZER_BASE_URL = os.environ.get("ANONYMIZER_BASE_URL", ANONYMIZER_BASE_URL)
+IMAGE_REDACTOR_BASE_URL = os.environ.get("IMAGE_REDACTOR_BASE_URL",
```
SharonHart, line 12: The other side of reading the url from env var, is setting it in the CI pipeline when smoke testing a deployed instance of our services.
In analyzer and anonymizer cases we run @integration marked tests on the staging slot during deployment, and I think it might be worth to choose one happy scenario that you've added and mark it as integration just to have a smoke test for image redactor. See deploy-stage.yml:103 for reference.

SharonHart (changes requested):

e2e-tests/common/methods.py
shiranr: it was inline, the method looked very messy. 
I like it this way :)
But you are correct regarding the name and also I made it private.

e2e-tests/common/methods.py
shiranr: Done.

e2e-tests/tests/test_image_redactor.py
shiranr: Left only one happy scenario and added the api mark

navalev:
since the e2e tests fail. I suspect its the tesseract version - let's try to re-create the tests against the docker image, fixing the tesseract version to 4.0.0

e2e-tests/tests/test_image_redactor.py
shiranr: I actually think 422 Unprocessable Entity - fits here no?

SharonHart: Nit, not critical

e2e-tests/tests/test_image_redactor.py
shiranr: Done.

shiranr: The image can be large... so not sure we should be duplicating the bytes, wdyt?

e2e-tests/common/methods.py
shiranr: I want to make sure since we are not really defining the image redactor as 100% ready yet...
@navalev WDYT should we add smoke test for the image aredactor

navalev: Currently we do not add the image redactor to the prod demo site and i would avoid it in the integration tests as the tesseract issue is still open. Can be added in future versions once its more stable

e2e-tests/tests/test_image_redactor.py
SharonHart: Not a must

SharonHart (approved):

e2e-tests/tests/test_image_redactor.py
```diff
@@ -0,0 +1,44 @@
...
+
+def test_given_image_with_tuple_color_filling_then_we_get_redact_image():
+    # purple redact
+    expected_result_image = get_file("purple_redact.png")
+    response = image_redactor("original_image.png", "102, 0, 102")
+    validate_and_close(expected_result_image, response)
+
+
+def test_given_image_with_no_color_filling_then_we_get_redact_image():
+    # black redact
```
navalev: 422 fits here

e2e-tests/tests/test_image_redactor.py
navalev: if its just for this test, use a small image, but really not a must

presidio-image-redactor/app.py
```diff
@@ -59,6 +59,9 @@ def redact():
             params = get_json_data(request.form.get("data"))
             color_fill = color_fill_string_to_value(params)
             image_file = request.files.get("image")
+            if not image_file:
+                raise InvalidParamException(
+                    "Invalid parameter, please insert image data")
```
navalev, line 64:
```suggestion
                    "Invalid parameter, please add image data")
```

e2e-tests/tests/test_image_redactor.py
navalev: 422 is appropriate here

navalev (changes requested):

navalev (approved):
approved,  with minor edits

presidio-image-redactor/app.py
shiranr: Done.

e2e-tests/tests/test_image_redactor.py
shiranr: @navalev We don't want a default? 
The user must send the filling?

navalev: what should be the default?

shiranr: (0, 0, 0) black

shiranr: I can re-add this test.

PR #544, opened 2021-02-22
update notebook sample with anonymizer changes

navalev:


docs/samples/python/presidio_notebook.ipynb
```diff
@@ -30,7 +34,7 @@
    "source": [
     "from presidio_analyzer import AnalyzerEngine, PatternRecognizer\n",
     "from presidio_anonymizer import AnonymizerEngine\n",
-    "from presidio_anonymizer.entities import AnonymizerRequest"
```
SharonHart, line 33: imports can be done mid-code, and in this case moved to the bottom source clause , so that it will be clearer how to use each service in separate.

docs/samples/python/presidio_notebook.ipynb
```diff
@@ -152,9 +156,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "converted_analyzer_results = [result.to_dict() for result in analyzer_results]\n",
+    "new_results = [AnalyzerResult(result.entity_type, result.start, result.end, result.score) for result in analyzer_results]\n",
     "\n",
-    "print(converted_analyzer_results)\n"
+    "[print(result) for result in new_results]"
```
SharonHart, line 161: print(new_results) won't give the same output?

SharonHart (approved):
Looks good, left some suggestions

omri374 (approved):

docs/samples/python/presidio_notebook.ipynb
```diff
@@ -152,9 +156,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "converted_analyzer_results = [result.to_dict() for result in analyzer_results]\n",
+    "new_results = [AnalyzerResult(result.entity_type, result.start, result.end, result.score) for result in analyzer_results]\n",
```
omri374, line 159: I think we can skip this part and just pass the analyzer results directly to the AnonymizerEngine

docs/samples/python/presidio_notebook.ipynb
navalev: they can, but the notebook is more organized this way.

docs/samples/python/presidio_notebook.ipynb
navalev: no, it prints the objects

docs/samples/python/presidio_notebook.ipynb
navalev: analyzer and anonymizer dont have common classes, so we need to convert the analyzer results (list of RecognizerResult) to a anonymizer expected input (list of AnalyzerResult)

omri374: It would work even without it. wild wild west Python 😊

PR #547, opened 2021-02-23
V2 anonymizer improvements

shiranr:
improve runtime for analyzer results loop by reducing the list size and keeping the code clean.


presidio-anonymizer/presidio_anonymizer/entities/analyzer_results.py
```diff
@@ -54,11 +54,12 @@ def _remove_conflicts(self):
         :return: List
         """
         unique_elements = []
+        other_elements = AnalyzerResults(self)
         for result_a in self:
-            other_elements = AnalyzerResults(self)
             other_elements.remove(result_a)
             if not any([result_a.has_conflict(other_element) for other_element in
                         other_elements]):
+                other_elements.append(result_a)
```
navalev, line 62: why return an already checked element back?

navalev (commented):
what is the expected improvement in performance?

presidio-anonymizer/presidio_anonymizer/entities/analyzer_results.py
shiranr: Because others need to compare themselves with it..
we checked one side that result_a is not contained in another result
But we need to make sure other results are not contained in result_a.

SharonHart:
I don't see how this is improving anything drastically, the list we are manipulating is not expected to be long, and even if it was, the change is not affecting complexity. Overall adds more stateful operations with more chance for a bug next time someone modifies the method.

shiranr:
@SharonHart moving the initialization and duplication of the list from each loop to a one time thing, changes a lot.
I think this is a really good change and we should keep it.
I won't be doing any further changes on this one for now...

SharonHart (approved):
Approving since logically it works and does less operations. However, time complexity dodn't change, and we are operating in a more mutable manner, which might be more confusing and error prone.

navalev (approved):
approving based on our offline discussion - reducing run time in 50%

PR #548, opened 2021-02-24
V2 remove fpe from anonymizer

shiranr:
Please notice, this branch is over the v2_anonymizer_improvements branch so the loop change is included here as well.
Remove FPE from anonymizer since we currently do not support it.


navalev:
@shiran - can u do this change over the main branch, so it will not include other changes? as we want to introduce this change now

PR #553, opened 2021-02-24
Feature/rename interpretability

omri374:
- Renamed interpretability to "Decision Process"
- Default response doesn't contain the decision process
- Aligned tracing vs. returning
- Updated code, e2e tests and documentation


navalev (approved):

docs/api-docs/api-docs.yml
```diff
@@ -299,10 +299,7 @@ components:
           items:
             $ref: "#/components/schemas/EntityTypes"
           description: "A list of entities to analyze"
-        trace:
-          type: boolean
-          description: "Whether to trace the request"
-        remove_interpretability_response:
+        return_decision_process:
```
SharonHart, line 302: So trace is no longer supported?

e2e-tests/README.md
```diff
@@ -0,0 +1,23 @@
+# End-to-end tests
```
SharonHart, line 1: 👍

SharonHart (approved):

docs/api-docs/api-docs.yml
navalev: it does - the feature name was changed, and it is now an opt-in (default should be false)

omri374: It doesn't. In order to start it you would have to change the AnalyzerEngine initialization and then it would work for all requests. I didn't think it makes sense to have it on each request, as someone can make calls without it and we would have tracing only for some of the calls which misses the point IMHO. What are your thoughts on this?

SharonHart: Makes sense, fealt wierd as part of the request anyway. If someone has access to the logs, then he runs the application himself and can set it by himself

navalev: @omri374 wasnt that the purpuse of this feature - to allow certain calls to be traced?

omri374: Yes, but activating it via the request could cause some requests to be missed. So the user can either ask for this to be returned (return_decision_process) or to be logged, which requires a different configuration.

omri374: Just making sure we're taking about the same thing. It can be returned via the rest api (that's where the rename + opt in goes), but for tracing the decision process into logs something else is needed

PR #558, opened 2021-02-28
Documentation - Add design diagrams

SharonHart:
Add design diagrams to desgin.md and to each service description


presidio-anonymizer/README.MD
```diff
@@ -5,6 +5,8 @@
 The Presidio anonymizer is a Python based module for anonymizing detected PII text
 entities with desired values.
 
+![Anonymizer Design](../docs/assets/anonymizer-design.png)
```
SharonHart, line 8: Both here, and for the image redactor, should be change to '../assets/analyzer-design.png' as for the analyzer, after the docs will move the docs dir

navalev: why cant this be done now?

omri374 (approved):

presidio-anonymizer/README.MD
SharonHart: It can, understood from @shiranr in the daily that shes taking it?

docs/design.md
```diff
@@ -1,5 +1,11 @@
 # Presidio design
 
-## Table of contents
+## Analyzer
+[![Analyzer Design](assets/analyzer-design.png)](analyzer/index.md)
```
SharonHart, line 4: @omri374 Should I use absolute path here as well? How is the linking handled in mkdocs?

navalev (approved):

PR #562, opened 2021-02-28
API Documentation Styling

SharonHart:
Match redoc colors and fonts to mkdocs.

Demo here:
https://www.w3schools.com/code/tryit.asp?filename=GO4LLATLM2J0

Let me know if you think of other styling related changes that can be done.


omri374 (approved):

navalev (approved):
nice!

PR #565, opened 2021-02-28
 image-redactor release pipeline

navalev:
Workaround to have the image-redactor using a different version than the other services.
added a secodn VERSION file for the image-redactor, the setup.py is reading from it. for ACR release added a new job in the release pipeline to get a different version. Left the ci pipeline as is


omri374 (approved):

docker-compose-image-redactor.yml
```diff
@@ -0,0 +1,12 @@
+version: '3'
```
omri374, line 1: maybe rename to `docker-compose-text` and `docker-compose-image`?

.pipelines/templates/release-image-redactor.yml
```diff
@@ -0,0 +1,54 @@
...
+stages:
+    - stage: Release
+      displayName: Release
+      jobs:
+          - job: version
+            displayName: 'Get the version number'
+            steps:
+                - bash: |
+                      set -eux  # exit on error
+                      ver=$(cat VERSION-IMAGE-REDACTOR)
```
omri374, line 17: Is this the only change between this service and others? Maybe we can have it as a parameter and make it more generic.

docker-compose-image-redactor.yml
navalev: done

.pipelines/templates/release-image-redactor.yml
navalev: eventually I reverted to a single pipeline with a seperate job for the image-redactor

.pipelines/templates/build-and-push-containers.yml
```diff
@@ -10,12 +10,15 @@ parameters:
 - name: IMAGE_PREFIX
   type: string
   default: ''
+- name: DOCKER_COMPOSE_FILE
+  type: string
+  default: 'docker-compose.yml'
```
SharonHart, line 15: Thinking out load of options to avoid compose file duplication.
Optionally, compose receives service names so it can partially operate on part of the services. It could help here but I haven't found this option in the AZO task documentation page. 'Run' has 'serviceName' argument but no the others.
We can also use '[profiles](https://docs.docker.com/compose/profiles/)' and set COMPOSE_PROFILES env var to build part all / image / text. 
But this works as well...

navalev: a possible V2.5 taks? :) lets brainstorm offline

SharonHart (approved):
Looks good. I would just see if we can use a single file (added suggestion), but not critical.
Can you also explain why did we want to differentiate the versions? Also for future reference.

.pipelines/templates/build-and-push-containers.yml
SharonHart: Sure,
If the 'beta' indication is more important than the major.minor versioning (somewhat makes sense since it uses v2 of the other packages), the we can also do:
1. 
```yml
  presidio-analyzer:
    image: ${REGISTRY_NAME}${IMAGE_PREFIX}presidio-analyzer${TAG}
...
  presidio-image-redactor:
    image: ${REGISTRY_NAME}${IMAGE_PREFIX}presidio-image-redactor${TAG}-beta
...

```
2. hardcode beta in setup.py

navalev: we can just update the image-redactor version file

navalev:
@SharonHart u mean use a single docker-compose file?
the diff in version originates from the image-redactor still being in beta version, and we want it to not take the same version as the analyzer and anonmizer. this is a workaround that will be removed eventually.

PR #567, opened 2021-02-28
Update release.yml for Azure Pipelines

SharonHart:


navalev:
why hard code the version in the pipeline?

PR #571, opened 2021-02-28
inclusive lint fixes

navalev:


SharonHart (approved):

shiranr:
@navalev why use deny list instead of black list? (AFAIK, black list is an agreed term, deny list is something less clear).

navalev:
@shiranr this is one of the outputs from the inclusive linting done as part of the pipeline. deny list is a common term that is more inclusive

presidio-analyzer/presidio_analyzer/pattern_recognizer.py
```diff
@@ -22,7 +22,7 @@ class PatternRecognizer(LocalRecognizer):
 
     :param patterns: A list of patterns to detect
     :param deny_list: A list of words to detect,
-    in case our recognizer uses a predefined list of words (black list)
+    in case our recognizer uses a predefined list of words (deny list)
```
omri374, line 25: This was here on purpose 😊 if somebody doesn't know thay deny list is the new black list...

omri374 (approved):

PR #573, opened 2021-02-28
Anonymizer should return original text if analyzer results are missing

navalev:
resolve bug 2940 - Anonymizer should return original text if analyzer results are missing


shiranr (approved):
LGTM

e2e-tests/tests/test_anonymizer.py
```diff
@@ -69,8 +69,8 @@ def test_given_anonymize_called_with_empty_analyzer_results_then_invalid_input_m
     """
     response_status, response_content = anonymize(request_body)
 
-    expected_response = '{"error": "Invalid input, analyzer results can not be empty"}'
-    assert response_status == 422
+    expected_response = """{"result":"hello world, my name is Jane Doe. My number is: 034453334"}"""
```
omri374, line 72: Maybe put a string without PII in it? Otherwise it might be confusing

omri374 (approved):

PR #577, opened 2021-03-01
Reduce docs build time

omri374:
Docs take a lot of time to build because it installs Presidio's modules for automated API reference generation.
This PR proposes:
1. Run `pip install` on presidio-analyzer, presidio-anonymizer and presidio-image-redactor with `--no-dependencies`
2. Split between the requirements-docs.txt which have mkdocs + plugins, and the installation of Presidio

### DIFF
--- a/docs/requirements-docs.txt
+++ b/docs/requirements-docs.txt
@@ -3,6 +3,7 @@ mkdocs-material
 pymdown-extensions
 markdown
 mkdocstrings
-presidio_analyzer
-presidio_anonymizer
-presidio_image_redactor
+
+# presidio_analyzer 
+# presidio_anonymizer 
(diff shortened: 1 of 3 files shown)


docs/requirements-docs.txt
```diff
@@ -3,6 +3,7 @@ mkdocs-material
 pymdown-extensions
 markdown
 mkdocstrings
-presidio_analyzer
-presidio_anonymizer
-presidio_image_redactor
\ No newline at end of file
+
+# presidio_analyzer
```
omri374, line 7: uncommented here so that we could install them locally if needed to generate docs.

navalev (approved):

PR #587, opened 2021-03-02
Release Pipeline - continue on error for GH release

SharonHart:
When running the release pipeline, and when the VERSION is the same, ignore GH release step failure to continue to the deployment

### DIFF
--- a/.pipelines/templates/release.yml
+++ b/.pipelines/templates/release.yml
@@ -33,6 +33,8 @@ stages:
 
     - job: githubRelease
       displayName: GitHub Release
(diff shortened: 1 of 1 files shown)


navalev (approved):

omri374:
Are we sure we want to overwrite releases?

SharonHart:
@omri374  Shouldn't be the go-to scenario, GH is the only step that fails if a release exists. In case we want to redeploy if something happened, which is not the common usecase, this will allow that.

omri374 (approved):

Issue #590, opened 2021-03-03
Running a slightly modified version of the demo code results in an error


jhorowitz:
**Describe the bug**
Running into an AttributeError in what appears to be a trivial case. Perhaps I'm doing something wrong?

**To Reproduce**

[This colab reproduces the issue](https://colab.research.google.com/drive/1FmzrHkJV7uno5EIiFrQu6amNglVyx0Nj?usp=sharing) if the error below isn't enough.

Code:
```
text="My phone number is 212-555-5555 and the IBAN is IL150120690000003111111 "
results = analyzer.analyze(text=text, language='en')
anonymized_text = anonymizer.anonymize(text=text, analyzer_results=results)
```

Output:

```
AttributeError                            Traceback (most recent call last)
<ipython-input-4-834d0ad09e29> in <module>()
      1 text="My phone number is 212-555-5555 and the IBAN is IL150120690000003111111 "
      2 results = analyzer.analyze(text=text, language='en')
----> 3 anonymized_text = anonymizer.anonymize(text=text, analyzer_results=results)

3 frames
/usr/local/lib/python3.7/dist-packages/presidio_anonymizer/entities/analyzer_results.py in <listcomp>(.0)
     61         for result_a in self:
     62             other_elements.remove(result_a)
---> 63             if not any([result_a.has_conflict(other_element) for other_element in
     64                         other_elements]):
     65                 other_elements.append(result_a)

AttributeError: 'RecognizerResult' object has no attribute 'has_conflict'
```

hkarakose:
I have the problem too. Can we rollback to a previous version to get rid of the problem until bug fix has been released?

omri374:
Hi, we are working on a fix. A workaround: 

```python
from presidio_analyzer import AnalyzerEngine
from presidio_anonymizer import AnonymizerEngine
from presidio_anonymizer.entities import AnalyzerResult

analyzer = AnalyzerEngine()
anonymizer = AnonymizerEngine()


text="My phone number is 212-555-5555 and the IBAN is IL150120690000003111111 "
results = analyzer.analyze(text=text, language='en')

# translate results to fit the anonymizer (this will be unecessary on the next version)
new_results = [AnalyzerResult(start=r.start, end=r.end, entity_type=r.entity_type, score=r.score) for r in results]

anonymized_text = anonymizer.anonymize(text=text, analyzer_results=new_results)
```

This is just a workaround until we fix the issue.

jhorowitz:
[...]

PR #591, opened 2021-03-03
Align analyzer and anonymizer RecognizerResult objects

omri374:
Solves #590

This PR proposes:
- Rename `AnalyzerResult` to `RecognizerResult` on the anonymizer side
- Align (as much as possible) the two classes to have the same functionality
- Fixed a bug where the `RecognizerRegistry` would return the same recognizer multiple times if it supports more than one entities that are requested
- Moved the demo website test to e2e tests due to issues with the demo test


navalev (approved):
lgtm, this brings up again the discussion of having a common package, or unifying the analyzer with the anonymizer to reduce code duplications and friction.

shiranr (approved):
LGTM

PR #592, opened 2021-03-03
installation of spacy 3.0 instead of 2.3

m0canu1:


omri374:
Hi @m0canu1 thanks for your contribution!
In order to introduce a new dependency (spacy 3 in this case), one has to update the Pipfile, and then lock it. Both the Pipfile and Pipfile.lock should be committed into the repo.

More on this can be found here: https://microsoft.github.io/presidio/development/#Setting-up-Pipenv

PR #595, opened 2021-03-04
Anonymizer - Add AES Encryption Logics

SharonHart:
Added AESCypher service, that uses pycryptodome for AES-CBC encryption, to be used by an 'encrypt anonymizer' (will be integrated in a next iteration)
Placed the class in a 'services' dir, and moved validations there as well.


presidio-anonymizer/Pipfile
```diff
@@ -5,6 +5,7 @@ name = "pypi"
 
 [packages]
 flask = "==1.1.2"
+pycryptodome = "==3.10.1"
```
SharonHart, line 8: @omri374  Verifying that the licensing is valid for us. Its BSD-2, which is less restrictive than flask's BSD-3 for instance

shiranr (dismissed):
We need to verify the license. Other than that, all good.

presidio-anonymizer/presidio_anonymizer/services/aes_cipher.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def __init__(self, key: bytes) -> None:
+        """
+        Instantiate AESCipher.
+
+        :param key: AES key in bytes
+        """
+        self.key = key
+
+    def encrypt(self, text: str) -> str:
```
navalev, line 21: How about having the ket be part of the encrpt/decrypt API, this way the same ciper class can be re-used?

navalev (commented):
This looks great, just one question before approving

presidio-anonymizer/presidio_anonymizer/services/aes_cipher.py
SharonHart: Exactly why I chose to put it as a class instance.. makes more sense when you use as package, downside is instantiating it multiple times when using as a service. What do you think is better?
```python
aes = AESCipher(key)
aes.encrypt("text")
aes.decrypt("text")
```

navalev: even when used as a package, users do not use it directly, but inderctly using the Encrypt anonymizer. This mean that every anonymizer call will instatiate a new Cipher object, To be more alighn with the rest of the implemetation I would go with aving a stateless approach, where the encrypt/decrypt methods recieve the key as an argument.

SharonHart: Makes sense, will push a change

SharonHart: @navalev Done

navalev (approved):
looks great!

presidio-anonymizer/presidio_anonymizer/services/aes_cipher.py
navalev: approved

shiranr (approved):

PR #604, opened 2021-03-10
Anonymizer Indexes in response (WIP)

itye-msft:


presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
```diff
@@ -67,11 +71,26 @@ def anonymize(
...
+            result_item = AnonymizerResultItem(anonymizer_config.anonymizer_name,
+                                               analyzer_result.entity_type,
+                                               0,
+                                               index_from_end,
+                                               anonymized_text)
+
+            anonymizer_result.add_item(result_item)
+
+        # normalize the indexes to be index from start
+        text_len = len(text_builder.output_text)
```
shiranr, line 87: Please consider moving this logic to another method. 
It can even be a AnonymizerResult method which gets the params and know how to normalize them.

presidio-anonymizer/presidio_anonymizer/entities/anonymizer_result.py
```diff
@@ -0,0 +1,31 @@
+"""Handle a serializable anonymizer result."""
+import json
+
+
+class AnonymizerResult:
+    """Anonymizer result."""
+
+    def __init__(self, text=None, items=None):
```
shiranr, line 8: Please add hints for the parameters.

presidio-anonymizer/presidio_anonymizer/entities/anonymizer_result.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def to_json(self) -> str:
+        """Return a json string serializing this instance."""
+        return json.dumps(self, default=lambda x: x.__dict__)
+
+    def __eq__(self, other):
+        """Verify two instances are equal."""
+
+        return self.text == other.text and all(
+            map(lambda x, y: x == y, self.items, other.items))
```
shiranr, line 31: not sure, but isn't it enough to do self.items == other.items?

presidio-anonymizer/presidio_anonymizer/entities/anonymizer_result_item.py
```diff
@@ -0,0 +1,19 @@
+class AnonymizerResultItem:
+    """Anonymizer result item."""
+
+    def __init__(self, anonymizer, entity_type, start, end, anonymized_text):
```
shiranr, line 4: Please add hints

presidio-anonymizer/tests/integration/test_engine.py
```diff
@@ -64,7 +106,7 @@ def test_given_anonymize_with_encrypt_then_text_returned_with_encrypted_content(
 
     actual_anonymize_result = AnonymizerEngine().anonymize(
         text, analyzer_results, anonymizers_config
-    )
+    ).text
```
shiranr, line 109: Instead of doing .text consider checking the indices here as well.
Same for the other tests.

shiranr (commented):
Looks great! Added some comments...

presidio-anonymizer/presidio_anonymizer/entities/anonymizer_result.py
itye-msft: "Items" are by-ref values and need deep comparison. Simple comparison will fail because these are difference instances

presidio-anonymizer/tests/test_anonymizer_result.py
```diff
@@ -0,0 +1,24 @@
+
+from presidio_anonymizer.entities import AnonymizerResult
+from presidio_anonymizer.entities.anonymizer_result_item import AnonymizerResultItem
+
```
shiranr, line 4: I would add tests for the equal, as_json and add_item to verify that if anyone changes them in the future and break something, the developer will be aware.

presidio-anonymizer/presidio_anonymizer/entities/anonymizer_result.py
```diff
@@ -0,0 +1,47 @@
...
+        self.text = text
+
+    def add_item(self, item: AnonymizerResultItem):
+        """Add an item.
+
+        :param item: an item to add to the list.
+        """
+        self.items.append(item)
+
+    def normalize_item_indexes(self):
```
omri374, line 32: Can this be done on the fly, instead of in a post process? We have an intermediate object which holds the wrong indices (0 start for example). Might help with mitigating future bugs.

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
```diff
@@ -67,11 +71,21 @@ def anonymize(
...
-            text_builder.replace_text(
+            index_from_end = text_builder.replace_text(
                 anonymized_text, analyzer_result.start, analyzer_result.end
             )
 
-        return text_builder.output_text
+            result_item = AnonymizerResultItem(anonymizer_config.anonymizer_name,
+                                               analyzer_result.entity_type,
+                                               0,
+                                               index_from_end,
```
omri374, line 81: 1. Please define the parameter name (start=0 and not positional)
2. Can we already fill in the start here and skip the normalization later? isn't `start = index_from_end - len(anonymized_text)`?

itye-msft: 1. ok
2. no, because the while in the loop the text is being changed, and only at the end of the loop we can tell the len.

presidio-anonymizer/presidio_anonymizer/entities/anonymizer_result.py
shiranr: It can be done for each AnalyzerResult when it's created...

presidio-anonymizer/presidio_anonymizer/entities/anonymized_text_builder.py
```diff
@@ -40,11 +40,16 @@ def replace_text(self, anonymized_text: str, start: int, end: int):
...
 
-        self.output_text = self.output_text[
-                           : start] + anonymized_text + self.output_text[end_of_text:]
+        before_text = self.output_text[: start]
+        after_text = self.output_text[end_of_text_index:]
+        self.output_text = before_text + anonymized_text + after_text
+
+        # The replace algorithm is replacing the text from end to start.
+        # calculate and return the start point from the end.
+        return len(after_text) + len(anonymized_text)
```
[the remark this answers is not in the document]
itye-msft: This is the only place where you get the insertion index. If you won't return it, it's lost.

presidio-anonymizer/presidio_anonymizer/entities/anonymizer_result_item.py
```diff
@@ -0,0 +1,27 @@
+class AnonymizerResultItem:
```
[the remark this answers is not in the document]
itye-msft: But we already have a class with that name. Maybe I didn't understand ?

presidio-anonymizer/presidio_anonymizer/entities/anonymizer_result_item.py
```diff
@@ -0,0 +1,27 @@
...
+        :param end: end index in the anonymized text.
+        :param anonymized_text: the PII anonymized text.
+        """
+        self.anonymizer = anonymizer
+        self.entity_type = entity_type
+        self.start = start
+        self.end = end
+        self.anonymized_text = anonymized_text
+
+    def __eq__(self, other):
```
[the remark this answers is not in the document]
itye-msft: Added return value, but hint is not possible here as it's the same type of the class. (python error)

presidio-anonymizer/presidio_anonymizer/entities/anonymizer_result.py
itye-msft: No, because the text keep changing during the loop. When the loop is done, you can do this calculation

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
```diff
@@ -6,6 +6,8 @@
 from presidio_anonymizer.entities import RecognizerResult, AnalyzerResults
 from presidio_anonymizer.entities import AnonymizedTextBuilder
 from presidio_anonymizer.entities.anonymizer_config import AnonymizerConfig
+from presidio_anonymizer.entities.anonymizer_result import AnonymizerResult
```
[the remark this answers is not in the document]
itye-msft: no matter how I changed it, I kept getting circular init errors. So I left it as is.

presidio-anonymizer/tests/test_anonymizer_result.py
itye-msft: I added more tests like you suggested except from as_json because it contains no logic at all.

shiranr (approved):
LGTM just need to make sure the build is passing.

presidio-anonymizer/presidio_anonymizer/entities/anonymized_text_index_item.py
```diff
@@ -0,0 +1,31 @@
+class AnonymizedTextIndexItem:
+    """Anonymizer result item."""
```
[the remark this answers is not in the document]
omri374: AnonymizedEntity?

omri374:
@itye-msft please update the swagger spec too: https://github.com/microsoft/presidio/blob/main/docs/api-docs/api-docs.yml

shiranr (approved):
LGTM

docs/api-docs/api-docs.yml
```diff
@@ -559,8 +559,32 @@ components:
...
       type: object
       properties:
-        result:
+        text:
+          type: string
+        items:
+          type: array
+          description: "Array of anonymized entities"
+          items:
+            $ref: "#/components/schemas/AnonymizedEntitiy"
```
omri374, line 568: Typo: AnonymizedEntity

omri374 (approved):

shiranr (approved):

PR #611, opened 2021-03-14
Fixed installation guide url

hkarakose:
Old url points to navalev's repo

New url points to microsoft's repo


navalev (approved):
good catch! thanks

omri374 (approved):

PR #612, opened 2021-03-14
Fixed port number used by docker images

hkarakose:
Docker port numbers are not documented correctly. I noticed that port numbers are changed to 3000 in `Dockerfile`s


navalev (approved):

SharonHart (approved):

omri374:
@navalev @SharonHart are these the only places where this should be updated in?

navalev:
> @navalev @SharonHart are these the only places where this should be updated in?
there's also the individual services readmes. I'll handle it in a seperate PR

PR #615, opened 2021-03-16
documentation update - docker use port 3000

navalev:


SharonHart (dismissed):

docs/anonymizer/index.md
```diff
@@ -30,7 +30,7 @@ Persidio anonymizer comes with predefined anonymizers but can easily be extended
     docker pull mcr.microsoft.com/presidio-anonymizer
     
     # Run the container with the default port
-    docker run -d -p 5001:5001 mcr.microsoft.com/presidio-anonymizer:latest
+    docker run -d -p 5001:3000 mcr.microsoft.com/presidio-anonymizer:latest
```
omri374, line 33: Do we want both to have the same default port?

navalev: currently the default port for each service is 3000 (set in app.py). we set it to a different one in the dockercompose.yml, but when running it like this - the port is 3000, redirected to 5001 (or whatever port we want). otherwize this does not work

omri374: Sorry if I wasn't clear. Now both the analyzer and the anonymizer redirect from 3000 to 5001. Do we want the anonymizer to redirect to 5002?

navalev: got it, fixed

omri374 (approved):

PR #617, opened 2021-03-16
Sample - Spark on Azure Databricks

balteravishay:
This PR adds a sample (notebook and the required provisioning scripts using CLI) to show using presidio on Spark/Azure Databricks.
The sample reads input text files from an Azure Storage and writes back anonymized text files to an output folder.


navalev:
the docs/samples/deployments/index.md file should be updated as well with the new sample

docs/samples/deployments/spark/presidio_anonymize_blobs.ipynb
```diff
@@ -0,0 +1,156 @@
...
+    "  analyzer = AnalyzerEngine()\n",
+    "  anonymizer = AnonymizerEngine()\n",
+    "  analyzer_results = analyzer.analyze(text=text, language='en')\n",
+    "  anonymized_results = anonymizer.anonymize(\n",
+    "    text=text,\n",
+    "    analyzer_results=analyzer_results,    \n",
+    "    anonymizers_config={\"DEFAULT\": AnonymizerConfig(\"replace\", {\"new_value\": \"<ANONYMIZED>\"})}\n",
+    "  )\n",
+    "  return anonymized_results\n",
+    "anonymized_rdd = input_rdd.rdd.map(lambda x: (x[\"value\"], x[\"filename\"], anonymize_text(x[\"value\"]))).toDF([\"text\", \"filename\", \"anonymized_text\"])\n",
```
omri374, line 95: Not a spark expert, so I'm not sure what would be the best option. Maybe UDFs or PandasUDFs are going to be faster? I assume they are going to be easier to read which is another consideration here. 
Some references here:
1. https://medium.com/@fqaiser94/udfs-vs-map-vs-custom-spark-native-functions-91ab2c154b44
2. https://medium.com/@paolo_tenti/multi-label-multi-class-batch-classification-in-spark-with-pandas-udfs-f463f1abc09c

syedhassaanahmed: +1 to what @omri374 said. RDD is a Spark 1.x concept and would be difficult to grasp for someone who started with Spark 2.x. Here are some suggestions,
- When reading the text file we can directly load it into a DataFrame
```python
df = spark.read.csv(blobs, sep=r'\t')
```
- Refactor `anonymize_text()` into a [Pandas UDF](https://databricks.com/blog/2017/10/30/introducing-vectorized-udfs-for-pyspark.html) (recommended over regular Spark UDF due to performance)
- For small text files it probably won't matter but if possible let's reuse the `AnalyzerEngine` and `AnonymizerEngine` instances, otherwise there will be memory allocations on each call of `anonymize_text()`. In case of Pandas UDF, `anonymize_text()` will be called as many number of times as the Spark partitions.
- Instead of `anonymized_rdd.show()`, let's use the built-in Databricks `display(df)`.

docs/samples/deployments/spark/index.md
```diff
@@ -0,0 +1,148 @@
...
+    \"autoscale\": {
+        \"min_workers\": 2,
+        \"max_workers\": 8
+    },
+    \"azure_attributes\": { 
+        \"first_on_demand\": 1, 
+        \"availability\": \"ON_DEMAND_AZURE\", 
+        \"spot_bid_max_price\": -1 
+    }, 
+    \"node_type_id\": \"Standard_F4s_v2\",
```
syedhassaanahmed, line 63: Did we have a specific reason for choosing the `Standard_F4s_v2` SKU? If so we could mention it in this readme, or we could leave it to the General purpose D-series VM sizes.

docs/samples/deployments/spark/index.md
```diff
@@ -0,0 +1,148 @@
...
```
syedhassaanahmed, line 95: Same comment about VM SKU as above.

syedhassaanahmed:
Thanks @balteravishay for doing the great work of creating the Presidio sample on Databricks. I've added some small comments. It would improve the Notebook's readability if it was a Python source file instead of `.ipynb` which adds a lot of formatting noise and wraps the actual code inside JSON. Databricks supports exporting regular source files that are annotated in such a way that importing them back as a Databricks Notebook works and the cell formatting is still preserved. Shameless plug :) [here is an example](https://github.com/syedhassaanahmed/databricks-notebooks/blob/master/iothub_to_cosmosdb.py) PySpark notebook that was exported from Databricks.

docs/samples/deployments/spark/presidio_anonymize_blobs.ipynb
omri374: It makes sense to initialize the analyzer once and broadcast it to the other nodes. It's pretty heavy.

docs/samples/deployments/spark/presidio_anonymize_blobs.py
```diff
@@ -0,0 +1,153 @@
...
+
+dbutils.widgets.text(
+    "storage_account_name", "", "Blob Storage Account Name"
+)  # noqa: F821
+dbutils.widgets.text("storage_container_name", "", "Blob Container Name")  # noqa: F821
+dbutils.widgets.text(
+    "storage_account_access_key", "", "Storage Account Access Key"
+)  # noqa: F821
+dbutils.widgets.text(
+    "storage_input_folder", "input", "Storage Account Access Key"
```
syedhassaanahmed, line 33: Perhaps you intended to have "Storage input folder" as text instead of "Storage Account Access Key" ?

docs/samples/deployments/spark/presidio_anonymize_blobs.py
```diff
@@ -0,0 +1,153 @@
...
+# COMMAND ----------
+
+# MAGIC %md
+# MAGIC # Anonymize Text using Presidio
+
+# COMMAND ----------
+
+
+def anonymize_text(text: str) -> str:
+    analyzer = AnalyzerEngine()
```
syedhassaanahmed, line 92: Since @omri374 mentioned the `AnalyzerEngine` is heavy, should we instantiate it outside `anonymize_text()` and [broadcast](http://people.duke.edu/~ccc14/sta-663-2018/notebooks/S15E_Efficiency_In_Spark.html#Broadcast-Variables) it?

docs/samples/deployments/spark/presidio_anonymize_blobs.py
```diff
@@ -0,0 +1,153 @@
...
+        res.append(upload_to_blob(s1_item, s2_item))
+    return pd.Series(res)
+
+
+save_udf = pandas_udf(upload_series, returnType=StringType())
+
+out_df = anonymized_df.withColumn(
+    "processed", save_udf(col("anonymized_text"), col("filename"))
+)
+out_df.collect()
```
syedhassaanahmed, line 153: I can see the intention here is to store the files back to the output directory and preserve the filenames. This one is tricky to do with native Spark. In order to avoid creating Blob Client and writing another Pandas UDF, what if we tried something like this;
- Group the final dataframe by filenames e.g. `anonymized_df.groupBy(col("filename"))`
- Then for each group, use an aggregate function to consolidate all text
- The above will output a condensed dataframe which can be simply looped over and stored to blob storage with vanilla Python filesystem API. You might have to remove the `wasbs:///` to use Python FS API but that can be solved with [mounting the blob storage to DBFS](https://docs.databricks.com/data/data-sources/azure/azure-storage.html#mount-azure-blob-storage-containers-to-dbfs).

docs/samples/deployments/spark/presidio_anonymize_blobs.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+blob_service_client = BlobServiceClient(
+    account_url="https://" + storage_account_name + ".blob.core.windows.net/",
+    credential=storage_account_access_key,
+)
+container_client = blob_service_client.get_container_client(storage_container_name)
+
+blob_names = container_client.list_blobs(
+    name_starts_with=dbutils.widgets.get("storage_input_folder") + "/"
```
syedhassaanahmed, line 61: What if we instead mounted the blob storage to DBFS and then use `dbutils.fs.ls()` to obtain the list of blobs?

balteravishay: works for reading, not for writing. commented in code

docs/samples/deployments/spark/presidio_anonymize_blobs.py
balteravishay: can't load spacy models in parallel due to a limitation. commented in code

docs/samples/deployments/spark/presidio_anonymize_blobs.ipynb
balteravishay: done

docs/samples/deployments/spark/index.md
balteravishay: done

docs/samples/deployments/spark/presidio_anonymize_blobs.py
```diff
@@ -0,0 +1,156 @@
...
+# MAGIC # Import the text files from Azure Blob storage
+
+# COMMAND ----------
+
+storage_account_name = dbutils.widgets.get("storage_account_name")
+storage_container_name = dbutils.widgets.get("storage_container_name")
+storage_account_access_key = dbutils.widgets.get("storage_account_access_key")
+
+# mount the container
+spark.conf.set(
```
syedhassaanahmed, line 43: This config only needs to be set when accessing blob storage directly i.e. without mounting. When mounting the `dbutils.fs.mount()` method is enough. In order to avoid any access issues I also recommend to run an adhoc `spark.conf.remove("fs.azure.account.key." + storage_account_name + ".blob.core.windows.net")` in your own Databricks environment.

docs/samples/deployments/spark/presidio_anonymize_blobs.py
syedhassaanahmed: Noticed that the `presidio-analyzer` package [uses spaCy 2.3](https://github.com/microsoft/presidio/blob/bf46a61a6d06ad47f3c12b5cd975423ff767f243/presidio-analyzer/Pipfile#L7) Is it possible to update `presidio-analyzer` to use [spaCy v2.3.5](https://github.com/explosion/spaCy/releases/tag/v2.3.5) or higher? The reason I think its worth trying is because the parallel thread issue might have been fixed in the thinc dependency as [stated here](`presidio-analyzer`) and spaCy v2.3.5 contains a higher version of thinc.

docs/samples/deployments/spark/presidio_anonymize_blobs.py
syedhassaanahmed: That could be a permission issue on blob storage. What is the error that you got when writing?

balteravishay: the thing is that i'm trying to write a text file per row, with the name of the file provided as a column in the df.
that is not supported ootb since only dataframes can be written to hdfs.
opening a file using native python always resulted in "file not found" error, and i could not get the worker nodes to know the hdfs. is is just me? do you have a reference to how writing text files within pandas udf (i.e. on worker nodes) can be done?

docs/samples/deployments/spark/presidio_anonymize_blobs.py
balteravishay: using that to init the blobservice client for uploading. if we can't write single output files, or choose to output the sample as batched string files, we can remove this. 
@omri374 wdyt?

docs/samples/deployments/spark/presidio_anonymize_blobs.py
balteravishay: @omri374  worth a check?

docs/samples/deployments/spark/presidio_anonymize_blobs.py
syedhassaanahmed: > since only dataframes can be written to hdfs

Yes that is correct, hence I mentioned that for writing we can use the native Python API. The reason you might have received a "file not found" is because when using Python API you might need to use the path like this "/dbfs/..." instead of "wasbs:///" which you probably have in your filename column. This should help https://docs.databricks.com/data/databricks-file-system.html#access-files-on-mounted-object-storage

syedhassaanahmed: Here is an example notebook that shows how to write a text file per row with the name of the file in a column. It mounts Blob Storage to DBFS then writes using the Python fs API https://github.com/syedhassaanahmed/databricks-notebooks/blob/master/df_row_to_multiple_files.py

docs/samples/deployments/spark/presidio_anonymize_blobs.py
navalev: we can defintly add it to our backlog

syedhassaanahmed (approved):
Looks good. Once again, great work @balteravishay 🚀

navalev (approved):
Great work!

syedhassaanahmed (approved):

docs/samples/deployments/spark/presidio_anonymize_blobs.py
omri374: @navalev @balteravishay worth looking into, but instead of pushing a new version with spacy 2.3.5 we can first check by just pip installing spacy 2.3.5 on Databricks and see if it solves the issue. @balteravishay WDYT? worth the try? I don't want to keep this PR from closing so could be a future suggestion.

omri374 (approved):

PR #618, opened 2021-03-16
added e2e package test

itye-msft:
This PR introduces a simple "good day" scenario for integration of the analyzer and anonymizer packages.

In order to include the packages in the e2e project we had to do a workaround the config.py files, because in the venv context there is no VERSION file. The workaround introduces a fall back in case the VERSION file is not found. In this case we would use an environment variable, and the final fall back is for a default version which is considered pre-release so it wouldn't affect PyPI.

### DIFF
--- a/.pipelines/templates/e2e-tests.yml
+++ b/.pipelines/templates/e2e-tests.yml
@@ -29,7 +29,7 @@ steps:
           python -m pip install --upgrade pip
           pip install -r requirements.txt
           pip install pytest-azurepipelines
-          python -m spacy download en_core_web_sm
+          python -m spacy download en_core_web_lg
       workingDirectory: e2e-tests
       displayName: Install dependencies
 
--- a/.pipelines/templates/e2e-tests.yml
+++ b/.pipelines/templates/e2e-tests.yml
@@ -29,6 +29,7 @@ steps:
           python -m pip install --upgrade pip
           pip install -r requirements.txt
           pip install pytest-azurepipelines
+          python -m spacy download en_core_web_sm
       workingDirectory: e2e-tests
       displayName: Install dependencies
 
--- a/.pipelines/templates/e2e-tests.yml
+++ b/.pipelines/templates/e2e-tests.yml
@@ -29,7 +29,7 @@ steps:
           python -m pip install --upgrade pip
           pip install -r requirements.txt
           pip install pytest-azurepipelines
-          python -m spacy download en_core_web_lg
+          python -m spacy download en_core_web_sm
       workingDirectory: e2e-tests
       displayName: Install dependencies
 
--- a/e2e-tests/tests/test_e2e_integration_flows.py
+++ b/e2e-tests/tests/test_e2e_integration_flows.py
@@ -264,7 +264,7 @@ def test_given_text_with_pii_using_package_then_analyze_and_anonymize_complete_s
     expected_response.set_text("<PERSON> drivers license is <US_DRIVER_LICENSE>")
     expected_response.add_item(AnonymizedEntity("replace", "US_DRIVER_LICENSE", 28, 47, "<US_DRIVER_LICENSE>"))
     expected_response.add_item(AnonymizedEntity("replace", "PERSON", 0, 8, "<PERSON>"))
-    
+
     anonymizer = AnonymizerEngine()
     anonymizer_results = anonymizer.anonymize(analyzer_request["text"], analyzer_results)
     assert anonymizer_results == expected_response
--- a/e2e-tests/tests/test_e2e_integration_flows.py
+++ b/e2e-tests/tests/test_e2e_integration_flows.py
@@ -3,8 +3,9 @@ from pathlib import Path
 import pytest
 import json
 
-from presidio_analyzer import AnalyzerEngine
+from presidio_analyzer import AnalyzerEngine, RecognizerResult
 from presidio_anonymizer import AnonymizerEngine
+from presidio_anonymizer.entities import AnonymizerResult, AnonymizedEntity
 
 from common.assertions import equal_json_strings
 from common.methods import analyze, anonymize, analyzer_supported_entities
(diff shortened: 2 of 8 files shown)


.pipelines/templates/e2e-tests.yml
```diff
@@ -29,6 +29,7 @@ steps:
           python -m pip install --upgrade pip
           pip install -r requirements.txt
           pip install pytest-azurepipelines
+          python -m spacy download en_core_web_lg
```
navalev, line 32: this will download spacy  module for the anonymizer as well

presidio-analyzer/setup.py
```diff
@@ -11,8 +11,11 @@
 with open(path.join(this_directory, 'README.MD'), encoding='utf-8') as f:
     long_description = f.read()
 
-with open(os.path.join(parent_directory, "VERSION")) as version_file:
-    __version__ = version_file.read().strip()
+try:
+    with open(os.path.join(parent_directory, "VERSION")) as version_file:
+        __version__ = version_file.read().strip()
+except:  # noqa: E722
+    __version__ = os.environ.get("VERSION", "0.0.1-alpha")
```
navalev, line 18: is this to avoid the image-redactor current version? what if the version will change?

presidio-anonymizer/setup.py
```diff
@@ -14,8 +14,11 @@
 with open(path.join(this_directory, "README.MD"), encoding="utf-8") as f:
     long_description = f.read()
 
-with open(os.path.join(parent_directory, "VERSION")) as version_file:
-    __version__ = version_file.read().strip()
+try:
+    with open(os.path.join(parent_directory, "VERSION")) as version_file:
+        __version__ = version_file.read().strip()
+except:  # noqa: E722
+    __version__ = os.environ.get("VERSION", "0.0.1-alpha")
```
navalev, line 21: same  - what if the version will change?

.pipelines/templates/e2e-tests.yml
itye-msft: Excatly

presidio-analyzer/setup.py
itye-msft: This is called by requirements.txt of the e2e test project, and VERSION file does not exists in this context. For the rest of the projects, the VERSION file will exists and the value will be taken from there.

.pipelines/templates/e2e-tests.yml
omri374: Have you tried with the smaller mode? faster to download and run

e2e-tests/tests/test_e2e_integration_flows.py
```diff
@@ -239,3 +241,29 @@ def test_demo_website_text_returns_correct_anonymized_version():
...
+
+
+@pytest.mark.package
+def test_given_text_with_pii_using_package_then_analyze_and_anonymize_complete_successfully():
+    analyzer_request = {
+        "text": "John Smith drivers license is AC432223",
+        "language": "en",
+    }
+
+    expected_response = [
```
omri374, line 253: How about taking some of the examples we have in the samples and turning them into tests? Less json oriented and more python development oriented. In addition we should have maybe some additional tests with different parameters, making sure things to break in other scenarios, for example when the analyzer engine output is empty, or when we have intersecting entities detected.

presidio-analyzer/setup.py
omri374, line 18: Consider renaming the env variable to PRESIDIO_VERSION or something similar. Users might have this kind of variable set up before running this.

presidio-anonymizer/setup.py
omri374: @navalev this introduces a fall back in case the VERSION file is not found. In this case we would use an environment variable, and the final fall back is for a default version which is considered pre-release so it wouldn't affect PyPI

e2e-tests/tests/test_e2e_integration_flows.py
```diff
@@ -239,3 +241,29 @@ def test_demo_website_text_returns_correct_anonymized_version():
...
+        {"entity_type": "US_DRIVER_LICENSE", "start": 30, "end": 38, "score": 0.6499999999999999}
+    ]
+
+    analyzer = AnalyzerEngine()
+    analyzer_results = analyzer.analyze(analyzer_request["text"], analyzer_request["language"])
+    for i in range(len(analyzer_results)):
+        assert analyzer_results[i].entity_type == expected_response[i]['entity_type']
+        assert analyzer_results[i].start == expected_response[i]['start']
+        assert analyzer_results[i].end == expected_response[i]['end']
+        assert analyzer_results[i].score == expected_response[i]['score']
```
SharonHart, line 264: RecognizerResult has an __eq__ method

.pipelines/templates/e2e-tests.yml
navalev: ok, scratch that - the e2e test run only once, not for every package. but still worth testing with the S size spacy model

e2e-tests/tests/test_e2e_integration_flows.py
```diff
@@ -239,3 +241,29 @@ def test_demo_website_text_returns_correct_anonymized_version():
...
+
+    expected_response = """{"text": "<PERSON> drivers license is <US_DRIVER_LICENSE>", "items": [{"anonymizer": "replace", "entity_type": "US_DRIVER_LICENSE", "start": 28, "end": 47, "anonymized_text": "<US_DRIVER_LICENSE>"}, {"anonymizer": "replace", "entity_type": "PERSON", "start": 0, "end": 8, "anonymized_text": "<PERSON>"}]}"""
+    anonymizer = AnonymizerEngine()
+    anonymizer_results = anonymizer.anonymize(analyzer_request["text"], analyzer_results)
+    assert anonymizer_results.to_json() == expected_response
```
SharonHart, line 269: asserting on the object is complex? package tests assertion makes more sense on objects than jsons

presidio-analyzer/setup.py
```diff
@@ -11,8 +11,11 @@
 with open(path.join(this_directory, 'README.MD'), encoding='utf-8') as f:
     long_description = f.read()
 
-with open(os.path.join(parent_directory, "VERSION")) as version_file:
-    __version__ = version_file.read().strip()
+try:
+    with open(os.path.join(parent_directory, "VERSION")) as version_file:
+        __version__ = version_file.read().strip()
+except:  # noqa: E722
```
SharonHart, line 17: You can add except Exception: to avoid bare exception pep8 issue

.pipelines/templates/e2e-tests.yml
SharonHart: let's use the sm model for now

e2e-tests/tests/test_e2e_integration_flows.py
itye-msft: The `en_core_web_sm` didn't work because it looks like some other tests are looking for the LG one, and they failed. 

Don't we have unit tests that cover such regression cases (empty output, intersections etc)? I think that following the spirit of e2e tests we are not looking for fail scenarios.

For an e2e scenario I would choose one "good day" path. If the samples provide other "good day" scenarios that are meaningful, we can add it.

itye-msft: I converted the json into objects

navalev (dismissed):

SharonHart (approved):

omri374:
@itye-msft @navalev please don't merge until you review my comments. No major blockers but having the smaller model would make this much faster to install and run.

presidio-analyzer/setup.py
```diff
@@ -11,8 +11,11 @@
 with open(path.join(this_directory, 'README.MD'), encoding='utf-8') as f:
     long_description = f.read()
 
-with open(os.path.join(parent_directory, "VERSION")) as version_file:
-    __version__ = version_file.read().strip()
+try:
+    with open(os.path.join(parent_directory, "VERSION")) as version_file:
+        __version__ = version_file.read().strip()
+except Exception:
+    __version__ = os.environ.get("PRESIDIO_VERSION", "0.0.1-alpha")
```
omri374, line 18: Do we want to inject this env variable during e2e testing or use the default 0.0.1-alpha?

omri374 (dismissed):

presidio-analyzer/setup.py
itye-msft: This should be added to the pipeline settings, and will be read here

e2e-tests/tests/test_e2e_integration_flows.py
```diff
@@ -239,3 +243,38 @@ def test_demo_website_text_returns_correct_anonymized_version():
     # Assert equal
 
     assert expected_anonymized_text == actual_anonymized_text
+
+
+@pytest.mark.package
```
omri374, line 248: pytest complains it doesn't know what this is

SharonHart (dismissed):

e2e-tests/tests/test_e2e_integration_flows.py
itye-msft: fixed

SharonHart (approved):

omri374 (approved):

PR #624, opened 2021-03-20
[Draft]: Bump spaCy version to 3.0.5 (transformer based nlp models)

omri374:
Pros: 
- Higher F1 scores on PII entities (person, org, location) with new model `en_core_web_trf`.

Cons: 
- Runtime impact needs to be validated
- Much bigger containers (3.4 Gb instead of ~1 Gb)
- Non-transformer model (en_core_web_lg) has a regression on spaCy 3.0.5, should be fixed in 3.1.0 (see [this issue](https://github.com/explosion/spaCy/issues/7493)).


SharonHart:
That's a big increase in the image size, do we know what's the improvement in scores for our use-cases?

navalev:
Is there an option to make this configurable, to allow the end user to select the spacy version? only because of the model size. Is there a S size model for this version as well?

omri374:
@navalev the model size isn't necessarily the issue here. I think it is actually smaller than the previous one. The thing (if I'm not mistkane) is with spacy 3 in general, which introduces a whole lot more dependencies on deep learning packages.

omri374:
@navalev @SharonHart coming to think about it, all the big packages are dependencies of the en_core_web_trf model, and not spacy's. Assuming the accuracy is the same for the regular model (en_core_web_lg) between the two versions, we should be able to upgrade without the increase in size and complexity, and allow users to use the new trf model if they want.

SharonHart:
@omri374 Great so reverting to the lg spacy model? This will also partially unblock version 3.9 support

omri374:
@SharonHart yes still trying :)

PR #626, opened 2021-03-21
Return application/json mime type in all endpoints

SharonHart:
By default flask returns text/html mimetype for the content-type header. 
To return application/json, in flask, this PR changes the usage for either 
* ```flask.jsonify``` whenever the response content can be easily serialized (```__dict__``` on a custom simple object without composition), which automatically adds the applicaiton/json header.
* ```flask.Response``` which allows setting implicit header, in case we return a complex object with serialization that needs to be done with json.dumps, and instead of implementing a custom JSONEncoder

Also done in this PR:
* Removed ErrorResponse as now it its stale (to_json() is redundant using ```__dict__```, so now it holds a single fields that can be passed to jsonify) 
* Added an exception handler for all http errors, which used to be intercepted by the 'internal server error' exception handler. Calling an non existing path will now return 404 instead of 500 for example


balteravishay (commented):
nice!

balteravishay (approved):

shiranr (approved):

PR #630, opened 2021-03-22
Enable deploying Presidio services on a single App Service Plan

balteravishay:
This PR enables users to deploy presidio services on either a single App Service Plan (sku set to P2V3 to support both analyzer and image redactor), or on 3 different plans with different skus.
The default is now a single App Service Plan and takes affect on all deploy to azure scenarios and buttons.


docs/samples/deployments/app-service/presidio-services.json
```diff
@@ -204,7 +277,7 @@
                         "value": "[parameters('imagesArePublic')]"
                     },
                     "image": {
-                        "value": "[parameters('anonymizerImage')]"
+                        "value": "[parameters('imageRedactorImage')]"
```
navalev, line 280: seems that the imageRedactor deployment is done twice? there's a second image-redactor deployment later in the file

docs/samples/deployments/app-service/presidio-services.json
```diff
@@ -135,12 +142,69 @@
...
+                    }
+                }
+            }
+        },
         {
             "type": "Microsoft.Resources/deployments",
             "apiVersion": "2019-10-01",
             "name": "presidio-analyzer",
+            "dependsOn": [
+                "[resourceId('Microsoft.Resources/deployments', 'presidio-anonymizer')]"
```
navalev, line 206: what happens if the anonymizer will not be deployed?

balteravishay: nice catch!

balteravishay: didn't need the dependsOn. ARM knows how to handle it.

docs/samples/deployments/app-service/presidio-services.json
balteravishay: fixed

navalev (dismissed):

docs/samples/deployments/app-service/presidio-services.json
```diff
@@ -229,15 +251,18 @@
             "properties": {
                 "mode": "Incremental",
                 "templateLink": {
-                    "uri": "https://raw.githubusercontent.com/microsoft/presidio/main/docs/samples/deployments/app-service/presidio-app-service.json",
+                    "uri": "https://raw.githubusercontent.com/microsoft/presidio/feature/app-services-single-plan/docs/samples/deployments/app-service/presidio-app-service.json",
```
SharonHart, line 254: this needs to be changed before merging right?

SharonHart (approved):

omri374 (approved):

navalev (approved):

SharonHart (approved):

PR #636, opened 2021-03-24
Analyzer and Anonymizer Python 3.9 support

SharonHart:
Analyzer an Anonymizer to support python 3.9.
Image redactor will be in a different iteration as it needs Analyzer version to be bumped in pypi.
Essentially, spacy was the blocker, which was updated to the first version supporting 3.9 (we might keep upgrading in another iteration)


omri374:
@SharonHart if the issue is with the demo website's text, we can consider changing it a little bit.

SharonHart:
@omri374 Apparently [iban is a nationality](https://en.wikipedia.org/wiki/Iban_people) :)

e2e-tests/resources/demo.txt
```diff
@@ -7,7 +7,7 @@ On September 18 I visited microsoft.com and sent an email to test@presidio.site,
 
 My passport: 191280345 and my phone number: (212) 555-1234.
 
-This is a valid IBAN: IL150120690000003111111 . Can you please check the status on bank account 954567876544?
+This is a valid International Bank Account Number: IL150120690000003111111 . Can you please check the status on bank account 954567876544?
```
SharonHart, line 10: Needs to be changed in the website as well

navalev:
Let's remove the 3.6 support once this is added

e2e-tests/resources/demo.txt
omri374: Worth doing a PR and linking it here so we won't forget.

omri374 (approved):

navalev (approved):

Issue #633, opened 2021-03-23
Anonymizing multiple emails with Faker in a text message


lucazav:
Hi,

I'm trying to anonymize emails using Faker. All is ok if in my text there is only one email address. If it contains multiple email addresses, the same fake email is used multiple times. Here the code I'm using:

```
def anonymizeEmail(text_to_anonymize): 
    analyzer_results = analyzer.analyze(text=text_to_anonymize, entities=["EMAIL_ADDRESS"], language='en')

    anonymized_results = anonymizer.anonymize(
        text=text_to_anonymize,
        analyzer_results=analyzer_results,    
        anonymizers_config={"EMAIL_ADDRESS": AnonymizerConfig("replace", {"new_value": fake.safe_email()})}
    )

    return anonymized_results

fake = Faker('en_US')
fake.add_provider(internet)

analyzer = AnalyzerEngine()
anonymizer = AnonymizerEngine()

text = 'The user has the following two emails: email1@gmail.com and email2@gmail.com'
anonymizeEmail(text)
```

The output is:

> The user has the following two emails: kenneth74@example.net and kenneth74@example.net

How can I change the `AnonymizerConfig` in order to generate a fake email address for each one found into the string?

Thanks.

efung:
One way you could extend Presidio is by allowing the `AnonymizerConfig` objects to take lambda functions, not just static strings. 

Here's a diff that implements the idea:

```diff
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
modified: presidio-anonymizer/presidio_anonymizer/anonymizers/replace.py
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
@ presidio-anonymizer/presidio_anonymizer/anonymizers/replace.py:18 @ class Replace(Anonymizer):
         new_val = params.get(self.NEW_VALUE)
         if not new_val:
             return f"<{params.get('entity_type')}>"
-        return new_val
+        return new_val(text) if callable(new_val) else new_val

     def validate(self, params: Dict = None) -> None:
         """Validate the new value is string."""
-        validate_type(params.get(self.NEW_VALUE), self.NEW_VALUE, str)
+        new_val = params.get(self.NEW_VALUE)
+        if callable(new_val):
+            return
+        else:
+            validate_type(new_val, self.NEW_VALUE, str)
         pass

     def anonymizer_name(self) -> str:
```

and here's how you would call it:

```python
    anonymized_result = anonymizer.anonymize(
[...]

PR #654, opened 2021-04-05
Custom anonymizer with lambda type

navalev:
Based on issue #633 - allow passing a lambda for anonymizer type.
This PR suggests adding a new type of anonymizer - custom - that excepts a lmbda type.
This is only supported in the python package - the REST API will not work with accepting a labda (though it is not blocked by in the code)


e2e-tests/tests/test_anonymizer.py
```diff
@@ -109,7 +109,7 @@ def test_given_anonymizers_called_then_expected_anonymizers_list_returned():
     response_status, response_content = anonymizers()
 
     expected_response = """
-        ["hash", "mask", "redact", "replace", "encrypt"]
+        ["hash", "mask", "redact", "replace", "encrypt", "replace", "custom"]
```
SharonHart, line 112: I would leave custom out here, as its operational with package usage, not rest api. Maybe can be controlled with the operator_name interface method.

presidio-anonymizer/presidio_anonymizer/operators/custom.py
```diff
@@ -0,0 +1,44 @@
...
+    If the new value is not a lambda, this will act as a replace method
+    """
+
+    NEW_VALUE = "new_value"
+
+    def operate(self, text: str = None, params: Dict = None) -> str:
+        """:return: new_value."""
+        new_val = params.get(self.NEW_VALUE)
+        if not new_val:
+            return f"<{params.get('entity_type')}>"
```
SharonHart, line 23: You can call Relace.operate() to avoid duplication. Also, in other cases we assume that existence and validity are checked in validate().. so falling back to another implementation might be misleading (in case I accidentally  provided the function in a different variable than 'new_value' or for  an invalid statured function)

presidio-anonymizer/presidio_anonymizer/operators/custom.py
```diff
@@ -0,0 +1,44 @@
...
+        return new_val(text) if callable(new_val) else new_val
+
+    def validate(self, params: Dict = None) -> None:
+        """Validate the new value is string."""
```
SharonHart, line 27: is a function returning string

e2e-tests/tests/test_anonymizer.py
navalev: yes, still in draft. this functionality is only for the package. WIP

presidio-anonymizer/tests/test_anonymizer_engine.py
```diff
@@ -17,7 +17,7 @@
 
 def test_given_request_anonymizers_return_list():
     engine = AnonymizerEngine()
-    expected_list = ["hash", "mask", "redact", "replace", "encrypt"]
+    expected_list = ["hash", "mask", "redact", "replace", "encrypt" "replace", "custom"]
```
shiranr, line 20: Replace appear twice and a comma is missing...
```suggestion
    expected_list = ["hash", "mask", "redact", "replace", "encrypt", "custom"]
```

navalev: fixed

presidio-anonymizer/presidio_anonymizer/operators/custom.py
navalev: yep, debating if to allow other values expect for a function. will make it simpler.

presidio-anonymizer/presidio_anonymizer/operators/custom.py
```diff
@@ -0,0 +1,44 @@
+"""Replaces the PII text with lambda result."""
+from typing import Dict
+
+from presidio_anonymizer.operators import Operator, OperatorType
+from presidio_anonymizer.services.validators import validate_type
+from presidio_anonymizer.entities import InvalidParamException
+
+
+class Custom(Operator):
```
shiranr, line 9: nit: are we sure about the name? why not call this lambda?

presidio-anonymizer/presidio_anonymizer/operators/custom.py
shiranr: Why are we falling back here?
might be the user wants empty values for some values and not empty for other no?

presidio-anonymizer/presidio_anonymizer/operators/custom.py
navalev: name still WIP, open for suggestions :) isn't lambda a reserved name?

presidio-anonymizer/presidio_anonymizer/operators/custom.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class Custom(Operator):
+    """
+    Replace PII text entity with the results of a function executed on the PII text.
+
+    The function retrun type must be a string    
+    """
+
+    NEW_VALUE = "new_value"
```
shiranr, line 16: "new_val" perhaps LAMBDA = "lambda" ?

presidio-anonymizer/presidio_anonymizer/operators/custom.py
```diff
@@ -0,0 +1,40 @@
...
+    def operate(self, text: str = None, params: Dict = None) -> str:
+        """:return: result of function executed on the text."""
+        new_val = params.get(self.NEW_VALUE)         
+        return new_val(text)
+
+    def validate(self, params: Dict = None) -> None:
+        """Validate the provided function is returning a string."""
+        new_val = params.get(self.NEW_VALUE)
+        if callable(new_val):
+            if (type(new_val("PII")) == str):
```
shiranr, line 27: nit: 
```
            if (type(new_val("PII")) != str):            
                 raise InvalidParamException("Invalid function return type. must be a str")
```

presidio-anonymizer/presidio_anonymizer/operators/custom.py
navalev: I removed this. If the user wants to use any other method of anonymization, they can select a different operator. This one will only serve functions

presidio-anonymizer/presidio_anonymizer/operators/custom.py
shiranr: lambda yes, Lambda no :) we can put Lambda as a class it won't conflict. But need to think about it.

e2e-tests/tests/test_anonymizer.py
```diff
@@ -109,7 +109,7 @@ def test_given_anonymizers_called_then_expected_anonymizers_list_returned():
     response_status, response_content = anonymizers()
 
     expected_response = """
-        ["hash", "mask", "redact", "replace", "encrypt"]
+        ["hash", "mask", "redact", "replace", "encrypt", "custom"]
```
SharonHart, line 112: Won't it be misleading to get 'custom' in the REST API, but don't support it as an operation in the REST API?

navalev: get_anonymizers is a package API, so not only used by the REST API. We can potentially just block it in the REST API itself. WIP

navalev: request is blocked

presidio-anonymizer/presidio_anonymizer/services/app_entities_convertor.py
```diff
@@ -60,3 +60,14 @@ def deanonymize_entities_from_json(json: Dict) -> List['AnonymizerResult']:
...
+    @staticmethod
+    def check_custom_operator(operators: Dict[str, OperatorConfig]):
+        """
+        Check if an operator is of type custom.
+        """
+        for key, config in operators.items():
+            print(config)
+            if (config.operator_name == "custom"):
+                return True            
+        return False
```
SharonHart: drop print + tests.
Also suggestion:
```python
return any([config.operator_name == "custom" for config in operators.values()])
```

navalev: good catch, left some prints

presidio-anonymizer/presidio_anonymizer/operators/custom.py
SharonHart: Also vote for a name change, either lambda, function, callback...

presidio-anonymizer/presidio_anonymizer/operators/custom.py
```diff
@@ -0,0 +1,39 @@
...
+                return
+            else:
+                raise InvalidParamException("Function return type must be a str")
```
SharonHart, line 29:
```suggestion
            if not type(new_val("PII")) == str:
                raise InvalidParamException("Function return type must be a str")

```

presidio-anonymizer/presidio_anonymizer/operators/custom.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+
+    NEW_VALUE = "new_value"
+
+    def operate(self, text: str = None, params: Dict = None) -> str:
+        """:return: result of function executed on the text."""
+        new_val = params.get(self.NEW_VALUE)
+        return new_val(text)
+
+    def validate(self, params: Dict = None) -> None:
```
SharonHart, line 22:
```suggestion
    def validate(self, params: Dict) -> None:

```

presidio-anonymizer/presidio_anonymizer/services/app_entities_convertor.py
```diff
@@ -60,3 +60,8 @@ def deanonymize_entities_from_json(json: Dict) -> List['AnonymizerResult']:
             for result in decrypt_entity:
                 items.append(AnonymizerResult.from_json(result))
         return items
+
+    @staticmethod
+    def check_custom_operator(operators: Dict[str, OperatorConfig]):
```
SharonHart, line 65: missing UT

SharonHart:
@navalev Left some minor comments
BTW did you run 'Validate samples' manually on the branch?

navalev:
> @navalev Left some minor comments
> BTW did you run 'Validate samples' manually on the branch?

yep, just run it and no issues.

presidio-anonymizer/presidio_anonymizer/operators/custom.py
navalev: not sure about the name, discuss offline

presidio-anonymizer/presidio_anonymizer/services/app_entities_convertor.py
navalev: good catch

presidio-anonymizer/presidio_anonymizer/operators/custom.py
navalev: changed parameter name to lambda

SharonHart (approved):

shiranr (approved):

PR #668, opened 2021-04-13
Add operators dynamically

shiranr:
THIS IS A PREVIEW FOR THE MEETING ON MONDAY, NO CODE REVIEW IS NEEDED.
Remove operators from the __init__ file and make sure they are loaded when setting up the anonymizers and deanonymizers dictionaries.
Add dynamic add of operators, so when someone is using the python package, he can access the anonymizer/deanonymizer engine and add_anonymizer/deanonymizer and remove_anonymizer/deanonymizer


djejaquino:
I wonder why we can't have something like this by default. Plugging a new operator shouldn't require having your own fork.

omri374:
@djejaquino I agree. A workaround (without having to fork) is to use custom operators. See [this as an example](https://github.com/8400TheHealthNetwork/HebSafeHarbor/blob/4a0ba72e0f0ab5e8421f7b45637c463decaddbe3/hebsafeharbor/anonymizer/phi_anonymizer.py#L43)

Having said that, it is on our roadmap, just no ETA.

PR #675, opened 2021-04-21
Add Git tag on release

SharonHart:
Add a step for tagging git with Presidio version.


balteravishay (commented):
Not sure that we should allow release twice with the same version. Wdyt?

SharonHart:
@balteravishay What about bumping just the image redactor, not changing the common version, that still needs version update. Also might happen in case of failed deployment and in case we want to rerun the release pipeline

balteravishay (approved):

navalev (approved):

PR #679, opened 2021-04-25
Documentation code and samples regression test pipeline

### LINKED ISSUE #673: Outdated import in python sample `presidio_notebook.ipynb`
- **Problem:** In the `presidio_notebook.ipynb` notebook under `docs\samples\python`, the following import statement is not used and does not work: `from presidio_anonymizer.entities import AnonymizerConfig`.
- **Solution:** This import statement should be replaced by `from presidio_anonymizer.entities.engine import OperatorConfig` which is used in the last cell.

navalev:
## Change Description

This PR suggests an ADO pipeline to validate code samples integrity and check for regressions. This pipeline will be manualy triggered since it directly executes code. It validates:
 - python code snippets in readmes
 - python scripts samples
 - notebooks samples
 - docker (API requests) samples
 
The PR includes updates to readmes and samples 

Note - several readmes were updated with "exdown-skip" to skip code snipets from testing. 

## Issue reference

This PR fixes issue ADO3122
resolved #673
## Checklist

- [ ] I have signed the CLA


.pipelines/templates/test-md.py
```diff
@@ -0,0 +1,16 @@
+import os
+import exdown
+
+
+for root, dirs, files in os.walk("."):
+  for file in files:
+      if (file.endswith(".md") or file.endswith(".MD")):
```
shiranr, line 7: nit: file.str.lower().endswith(".md")

.pipelines/templates/test-docs.yml
```diff
@@ -0,0 +1,76 @@
+trigger: none
+pr: none
+
+pool:
+  vmImage: ubuntu-latest
+
+steps:
+- task: UsePythonVersion@0
+  inputs:
+    versionSpec: '3.x'
```
balteravishay, line 10: wouldn't you want to run this pipeline in multiple versions?

docs/analyzer/languages.md
```diff
@@ -96,7 +97,7 @@ and [Stanza](https://github.com/stanfordnlp/stanza). To set up new models, follo
         from presidio_analyzer.nlp_engine import NlpEngineProvider
 
         # Create NLP engine based on configuration file
-        provider = NlpEngineProvider(conf_file="PATH_TO_YAML")
+        provider = NlpEngineProvider(conf_file="./docs/analyzer/languages-config.yml")
```
balteravishay, line 100: maybe use environment variable?

balteravishay (commented):
looks great! 
I would suggest adding test-docs.yml as a yaml-template instead of pipeline, so that we can eventually run from our pipeline. isn't that the purpose?

.pipelines/templates/test-docs.yml
navalev: good point, I'll add a metrix for py versions

docs/analyzer/languages.md
navalev: @balteravishay in this case I will need to set the env var in the pipeline, making it less generic. What do you have in mind?

balteravishay: so maybe just a static const var?

.pipelines/templates/test-docs.yml
balteravishay: does it change the result or the code? why not test for all supported version?

navalev: There shoundt be an impact from the python version, and since it just a matter of validating the samples against the latest version of presidio, testing on one python version should be sufficient.

navalev:
> looks great!
> I would suggest adding test-docs.yml as a yaml-template instead of pipeline, so that we can eventually run from our pipeline. isn't that the purpose?

Not sure we want to test it as part of our pipeline, since it executes code, which can be a potential for malicious code etc. The purpuse if to run it manualy, before we do new documentation release (this will be added to the release process checklist)

docs/analyzer/languages.md
navalev: done

.pipelines/templates/test-docs.yml
```diff
@@ -0,0 +1,76 @@
+trigger: none
```
SharonHart, line 1: This is still not attached to the PR pipeline, right?

presidio-image-redactor/README.MD
```diff
@@ -62,8 +62,8 @@ redacted_image = engine.redact(image, (255, 192, 203))
 
 # save the redacted image 
 redacted_image.save("new_image.png")
-# open the image for viewing
-redacted_image.show()
+# uncomments to open the image for viewing
```
SharonHart, line 65: uncomment

SharonHart (dismissed):

balteravishay:
How is that different from unit or integration tests?
The problem is that we may end up with failures after the pr was done.
> > looks great!
> 
> > I would suggest adding test-docs.yml as a yaml-template instead of pipeline, so that we can eventually run from our pipeline. isn't that the purpose?
> 
> 
> 
> Not sure we want to test it as part of our pipeline, since it executes code, which can be a potential for malicious code etc. The purpuse is to run it manualy, before we do new documentation release (this will be added to the release process checklist)

docs/analyzer/languages.md
```diff
@@ -145,7 +148,7 @@ analyzer = AnalyzerEngine(
     supported_languages=["en","es"],
     nlp_engine=nlp_engine_with_spanish)
 
-analyzer.analyze(...)
+analyzer.analyze(text="My name is David", language="en")
```
omri374, line 151: Should be some name other than David :)

.pipelines/templates/test-docs.yml
navalev: nope

omri374 (approved):

balteravishay (approved):

PR #696, opened 2021-05-09
Update samples validation pipeline

navalev:
## Update samples validation pipeline

update exdown with the new version pytest-codeblocks


omri374 (approved):

.pipelines/templates/test-docs.yml
```diff
@@ -68,9 +69,7 @@ steps:
   displayName: Execute postman collection
   continueOnError: true  
 
-- task: PythonScript@0
-  displayName: 'Test python code in markdown files'
-  continueOnError: true
-  inputs:
-    scriptSource: 'filePath'
-    scriptPath: './.pipelines/templates/test-md.py'
```
SharonHart, line 76: Remove the file as well

.pipelines/templates/test-docs.yml
```diff
@@ -25,7 +25,8 @@ steps:
     
 - script: |
     set -eux  # fail on error    
-    pip install exdown 
+    pip install pytest-codeblocks
```
SharonHart, line 28: We can lock the version do avoid regressions

SharonHart (changes requested):

.pipelines/templates/test-docs.yml
navalev: done

SharonHart (approved):

PR #710, opened 2021-05-19
update dependencies - Bump pydantic from 1.7.3 to 1.7.4

navalev:
## Change Description

Bump pydantic from 1.7.3 to 1.7.4

## Issue reference

This PR fixes #706 and #707


presidio-analyzer/Pipfile
```diff
@@ -4,6 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
+pydantic = "==1.7.4"
```
SharonHart, line 7: wonder if we should add it to setup.py.
Its kind of weird because this is not a direct dependency.
I ran a dependency tree analysis and daw that its coming from en_core_web_lg whl, but seems like we are using the latest version

navalev: not sure. I only added it explicitly in the pipfile to make sure we use 1.7.4. An alternative would be to only update the piplock, as was done by the dependancybot

omri374: Where is `pydantic` coming from? Maybe we can update our direct dependencies and solve it this way? 
spaCy just committed an update to the pydantic version: https://github.com/explosion/spaCy/commit/06324e5a5eda1a8cfa6be34e06ddf78f331cb497

SharonHart: Indeed coming from spacy as well, though they haven't released a version with that change yet. We'll have to wait.

navalev: I suggest we update the piplock to comply with the dependency warnings, and add a task to recreate once space release the new version

SharonHart (approved):

omri374 (approved):

PR #732, opened 2021-06-09
Update release-docs.yml for Azure Pipelines

SharonHart:


balteravishay (approved):

navalev (approved):

SharonHart:
@omri374 Check the last docs release pipeline builds, the first of the recent builds failed on this missing dependency. Opened this branch as a workaround for running the pipeline, haven't checked if it's something that's reproducing locally - if so, we might need to add it to docs-requierments.txt

PR #747, opened 2021-07-20
Update phone_recognizer.py

jiankebe:
## Change Description

When only one country code is used, such as CN, supported_entities will be parsed as "C" and "N", so DEFAULT_SUPPORTED_ COUNTRY_ CODE  should be array

## Issue reference

This PR fixes issue #XX

## Checklist

- [ ] I have signed the CLA

### DIFF
--- a/presidio-analyzer/presidio_analyzer/predefined_recognizers/phone_recognizer.py
+++ b/presidio-analyzer/presidio_analyzer/predefined_recognizers/phone_recognizer.py
@@ -21,7 +21,7 @@ class PhoneRecognizer(LocalRecognizer):
 
     SCORE = 0.6
     CONTEXT = ["phone", "number", "telephone", "cell", "cellphone", "mobile", "call"]
-    DEFAULT_SUPPORTED_COUNTRY_CODES = ("US", "UK", "DE", "FE", "IL")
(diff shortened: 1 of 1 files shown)


jiankebe:
We can like this DEFAULT_SUPPORTED_COUNTRY_CODES = ('CN',) , But I think it's better to change it to array

omri374:
Hi @jiankebe, thanks for your PR! Could you please clarify why the default tuple should be changed?
In case a user would want a different set of country codes, they should be able to pass them through the init function of the `PHONE_RECOGNIZER` without having to change the default one:

```python
from presidio_analyzer.predefined_recognizers import PhoneRecognizer

# Create a new recognizer for CN numbers as CN is not a default region supported by the PhoneRecognizer
cn_phone_recognizer = PhoneRecognizer(supported_entities=["CN_PHONE_NUMBER"])
analyzer = AnalyzerEngine()
analyzer.registry.add_recognizer(cn_phone_recognizer)

res = analyzer.analyze(text="My number in Beijing is 10 6552 9988", language="en")
print(res)
```

Which results in
`[type: CN_PHONE_NUMBER, start: 24, end: 36, score: 0.95,
 type: LOCATION, start: 13, end: 20, score: 0.85]`

SharonHart:
That was exactly the intention, thanks @omri374.
However, @jiankebe, if you can explain the motivation for transforming into a list, I guess that it's not that of a drastic change to push.

Issue #754, opened 2021-08-11
Adding - Australian custom recognizers 


rakan41:
There are various Australian-specific PII entities that would benefit from PII detection. These include:

- Australian Business Number ("ABN")
- Australian Company Number ("ACN")
- Tax File Number ("TFN")
- Medicare Number 

The above entities would be detected using by creating custom "PatternRecognizers". All the mentioned entities will use regex patterns, context words, and checksums. Each of the entities would be prefixed with "AU_" to delineate that they are Australian-specific. 

For each entity, the following test strategy will be used:
- Create body of text with examples of the above entities. 
- Both invalid and valid examples will be included (from checksum perspective). 
- Test to see whether candidate PII entities are detected (regardless if valid). 
- Check that the valid PII entities have obtained a score of 1.0. 
- Ensure no false positives are detected.

PR #755, opened 2021-08-16
Issue 754

rakan41:
## Change Description

This PR adds 4 Australian-specific PII entities to the predefined recognizers. The four entities are:

- AU_ABN: Australian Business Number
- AU_ACN: Australian Company Number
- AU_MEDICARE: Australian medicare number
- AU_TFN: Australian Tax File Number
Tests have been added for each new reocgnizer. 

## Issue reference

This PR fixes issue #754 

## Checklist

- [X] I have signed the CLA


omri374:
Thanks @rakan41! Amazing contribution! We'll review it soon.

rakan41:
Amended the e2e tests to accommodate the new AU entities. E2E now pass locally.

e2e-tests/tests/test_e2e_integration_flows.py
```diff
@@ -200,11 +200,16 @@ def test_demo_website_text_returns_correct_anonymized_version():
     text = " ".join(text_into_rows)
     language = "en"
     score_threshold = 0.35
+    entities = ["PHONE_NUMBER", "US_DRIVER_LICENSE", "US_PASSPORT",
```
omri374, line 203: This test makes sure the demo website's text is anonymized correctly. While this test would pass by selecting specific entities, it would not guarantee that the demo website would work properly (as it uses all the default entities). 
Two possible solutions I can think of:
1. Make sure the new entities don't change the demo website's result by changing the logic for AU entities to avoid false positives
2. Alter the demo website's text to make sure there are no false positives. If you want to go this path we can sync offline and think together about that.
3. Have the new AU entities not part of the default presidio installation. 
 
False positives, especially for new users trying out the service, are considered bad experience, although we know that these are difficult to avoid on a real system.

presidio-analyzer/presidio_analyzer/predefined_recognizers/au_tfn_recognizer.py
```diff
@@ -0,0 +1,89 @@
...
+
+    PATTERNS = [
+        Pattern(
+            "TFN (Medium)",
+            r"\b\d{3}\s\d{3}\s\d{3}\b",
+            0.5,
+        ),
+        Pattern(
+            "TFN (Low)",
+            r"\b\d{9}\b",
```
omri374, line 29: Very weak regex. Please user a lower confidence (0.01)

presidio-analyzer/presidio_analyzer/predefined_recognizers/au_abn_recognizer.py
```diff
@@ -0,0 +1,90 @@
...
+        """
+        # Pre-processing before validation checks
+        text = self.__sanitize_value(pattern_text, self.replacement_pairs)
+        abn_list = [int(digit) for digit in text]
+
+        # Set weights based on digit position
+        weight = [10, 1, 3, 5, 7, 9, 11, 13, 15, 17, 19]
+
+        # Perform checksums
+        abn_list[0] = 9 if abn_list[0] == 0 else abn_list[0] - 1
```
omri374, line 75: The checksum logic for all these new entities seem very similar (please correct me if I'm wrong). Could we perhaps create a class with default logic and have the different AU entity recognizers inherit from it?

presidio-analyzer/tests/test_au_acn_recognizer.py
```diff
@@ -0,0 +1,51 @@
...
+        ("824 753 557", 1, ((0, 11),), ((0.5, 0.5),),),
+        ("824753557", 1, ((0, 9),), ((0.3, 0.3),),),
+        # Invalid formatting and ACNs.  
+        ("5282475355632", 0, (), (),),
+        ("52824753556AF", 0, (), (),),
+        ("51 824 753 5564", 0, (), (),),
+    ],
+)
+def test_when_all_acns_then_succeed(
+    text,
```
omri374, line 34: great work on tests!

omri374 (changes requested):

e2e-tests/tests/test_e2e_integration_flows.py
rakan41: Thanks @omri374 for the feedback. Yes, that makes sense, too many false positives would be a bad user experience. I think changing the score for the weak regex in the other comments will help resolve this (since they will be below the 0.35 score threshold). I think should resolve it because the new entities have quite generic patterns, but the checksum helps distinguish them quite well.

presidio-analyzer/presidio_analyzer/predefined_recognizers/au_abn_recognizer.py
```diff
@@ -0,0 +1,90 @@
...
+    PATTERNS = [
+        Pattern(
+            "ABN (Medium)",
+            r"\b\d{2}\s\d{3}\s\d{3}\s\d{3}\b",
+            0.5,
+        ),
+        Pattern(
+            "ABN (Low)",
+            r"\b\d{11}\b",
+            0.3,
```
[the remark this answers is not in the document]
rakan41: Noted. Makes sense, I'll update accordingly.

presidio-analyzer/presidio_analyzer/predefined_recognizers/au_tfn_recognizer.py
rakan41: Noted, will do.

presidio-analyzer/presidio_analyzer/predefined_recognizers/au_abn_recognizer.py
```diff
@@ -0,0 +1,90 @@
+from typing import Optional, List, Tuple
+
+from presidio_analyzer import Pattern, PatternRecognizer
+
+
+class AuAbnRecognizer(PatternRecognizer):
+    """
+    Recognizes Australian Business Number ("ABN").
```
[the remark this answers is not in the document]
rakan41: I've now fleshed out the description and reference in the comments of the files.

e2e-tests/tests/test_e2e_integration_flows.py
rakan41: Hi, update on the above. Changing the scores seems to have resolved the issue except for the "passport" test case. It clashes with "AU_TFN" because it just so happens that it is a valid TFN (big coincidence). This could be resolved by changing the test passport number by one digit (given passport does not have a checksum validation). Please let me know your thoughts.

presidio-analyzer/presidio_analyzer/predefined_recognizers/au_abn_recognizer.py
rakan41: There are similarities as all the checksum use a sum product and modulo operation. However, the following differs:

- Check digit appears at different locations for each entity
- Weights are different across the different entities
- Modulo number is different.
- AU_ACN checks the complement of the remainder of Modulo, rather than just remainder. 

Given the above, do you recommend creating a class?

presidio-analyzer/presidio_analyzer/predefined_recognizers/au_abn_recognizer.py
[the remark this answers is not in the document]
omri374: Thanks! looks great

presidio-analyzer/presidio_analyzer/predefined_recognizers/au_abn_recognizer.py
omri374: For the first three bullets I would say yes, but I assume that the fourth changes the logic, so I assume it would be an overkill to create a class. Let's keep it as is.

e2e-tests/tests/test_e2e_integration_flows.py
omri374: Sounds good. Let's go ahead with it. I'll just need the new passport number to update the demo website's text before merging this one.

omri374:
@rakan41 thanks for the code updates! There are a few linting issues.

rakan41:
Hi @omri374 ,
1) I've updated demo passport number to 191280342. e2e test now passes for taht. 
2) I've also fixed the linting in new recognizers

navalev:
Hi @rakan41 there are still some linting issues failing the build

navalev:
looks great! could u please also update the list of supported entities in our documentation as part of this PR? under docs/supported_entities.md

rakan41:
> looks great! could u please also update the list of supported entities in our documentation as part of this PR? under docs/supported_entities.md

Yes, will do sometime over next few days!

rakan41:
> 
> 
> looks great! could u please also update the list of supported entities in our documentation as part of this PR? under docs/supported_entities.md

All done now! Added them to md document.

omri374 (approved):

navalev (approved):

PR #771, opened 2021-09-19
Format preserving encryption

PuspenduBanerjee:
## Change Description

Added Support for Format Preserving Encryption as per NIST-FF3-1 guidelines.

## Issue reference

Feature

## Checklist

- [x] I have signed the CLA


omri374:
Hi @PuspenduBanerjee! Thanks for your contribution. FPE is definitely something we'd like to see integrated into presidio. I see it relies on the pyFPE package, but I couldn't find it on PyPI/Github. Could you please provide some references to the dependencies required for this change?

PuspenduBanerjee:
Hello @omri374 ,
Thanks for reaching out. Reference to PyFPE are as follows:
- [github](https://github.com/PuspenduBanerjee/python-fpe)
- [pypi](https://pypi.org/project/pyFPE/0.10.1/)

omri374:
Thanks! Looks like a brand new package. We'll definitely check it out.

PuspenduBanerjee:
> Thanks! Looks like a brand new package. We'll definitely check it out.

Yes. Let me know if you have any further questions on pyFPE.

navalev:
@PuspenduBanerjee, Thanks for your contribution! Will it be possible to convert the integration with PyFPE into a presidio usage sample? we will include it in our documentation once complete.

PuspenduBanerjee:
> @PuspenduBanerjee, Thanks for your contribution! Will it be possible to convert the integration with PyFPE into a presidio usage sample? we will include it in our documentation once complete.

@navalev here is a sample for anonymization with presidio [presidio-pyfpe-anonymization](https://github.com/PuspenduBanerjee/python-fpe/blob/presidio_integration/samples/anonymizer-presidio.py)

SharonHart:
@PuspenduBanerjee Thanks for that!
On anonymizer V1 we had FPE as the main encryption method used, but decided to go for AES at least as the default method.
One reason was the lack of proper and commonly used python package to use (we also found the one you've forked), so that our dependencies will be well-backed for bug fixes and improvements. Personally, since the original implementation is one file and uses pycryptodome that we already use, we can even implement it ourselves. 
More reasons to not use FPE at first, were the use of a predefined alpha-numeric radix (were presidio fits a multilingual, special characters use).
I think that @navalev's intention, which I back, was to have this as part of the [samples](https://microsoft.github.io/presidio/samples/) at first, as a recommendation for extension.

PuspenduBanerjee:
> @PuspenduBanerjee Thanks for that!
> On anonymizer V1 we had FPE as the main encryption method used, but decided to go for AES at least as the default method.
> One reason was the lack of proper and commonly used python package to use (we also found the one you've forked), so that our dependencies will be well-backed for bug fixes and improvements. Personally, since the original implementation is one file and uses pycryptodome that we already use, we can even implement it ourselves.
> More reasons to not use FPE at first, were the use of a predefined alpha-numeric radix (were presidio fits a multilingual, special characters use).
> I think that @navalev's intention, which I back, was to have this as part of the [samples](https://microsoft.github.io/presidio/samples/) at first, as a recommendation for extension.

@SharonHart Though the original implementation is one file, but my goal for pyFPE is to not only support vaultless tokenization , rather to support semi vaulted and fully vaulted tokenization as well .  Regarding 
>Personally, since the original implementation is one file and uses pycryptodome that we already use, we can even implement it ourselves.
Definitely, you can, but the beauty of OSS community is kind of crowd sourcing, let me know if you have a different opinion.

omri374:
Adding to @SharonHart's and @navalev's comments, I also think that adding a sample would be a good first step and a great addition, also as an example of a user-defined anonymizer/deanonymizer. This is often the path we go with before officially supporting a new functionality (see the [notebook on structured data](https://microsoft.github.io/presidio/samples/python/batch_processing/) as an example of this approach). 

Once we get some positive user feedback on it and see we can rely on 3rd party packages used, we can consider adding it as an official feature.

PuspenduBanerjee:
> Adding to @SharonHart's and @navalev's comments, I also think that adding a sample would be a good first step and a great addition, also as an example of a user-defined anonymizer/deanonymizer. This is often the path we go with before officially supporting a new functionality (see the [notebook on structured data](https://microsoft.github.io/presidio/samples/python/batch_processing/) as an example of this approach).
> 
> Once we get some positive user feedback on it and see we can rely on 3rd party packages used, we can consider adding it as an official feature.

That clarifies in a great detail. Let me put that together.

Issue #772, opened 2021-09-23
Missing documentation for the phone number recognizer


omri374:
Presidio uses the [phonenumbers](https://github.com/daviddrysdale/python-phonenumbers) python package for better handling and coverage of phone numbers. [See code here](https://github.com/microsoft/presidio/blob/main/presidio-analyzer/presidio_analyzer/predefined_recognizers/phone_recognizer.py). 

It currently uses five regions ("US", "UK", "DE", "FE", "IL") as default, but could be extended by the user to every region supported by the `phonenumbers` package.

We're missing documentation for this, and should also consider having this recognizer be used as default (see [this comment on the PR](https://github.com/microsoft/presidio/pull/722#issuecomment-924331601)).

lsmith77:
is this now fixed?

omri374:
Not yet. If you have a specific question we'd be happy to provide guidance.

ananyap18:
is this fixed ?

omri374:
Anything specific documentation missing?

PR #775, opened 2021-09-25
Phone Number Recognizer - Multi-Regional And set as Default

SharonHart:
## Change Description

This PR is introducing another iteration of the phone number recognizer, finalizing it and making it used by default.
Until now, a country code was appended to the entity name, f.e. US_PHONE_NUMBER, making this recognizer return multiple entity types. 
This was changed to return a single  PHONE_NUMBER entity, yet the recognizer still considers all regions, supported by phonenumbers external library. 
The results are also aggregated to not include duplicates, and their assurance score is enhanced with a base context and the country and region names. 
In this PR, US_PHONE_NUMBER is removed as it is now deprecated and  included in the new versioned recognizer.

## Issue reference

This PR fixes issue #772

## Checklist

- [x] I have signed the CLA


presidio-analyzer/presidio_analyzer/predefined_recognizers/phone_recognizer.py
```diff
@@ -21,20 +21,15 @@ class PhoneRecognizer(LocalRecognizer):
...
 
     def __init__(
         self,
         context: Optional[List[str]] = CONTEXT,
         supported_language: str = "en",
-        supported_entities: List[str] = [
-            code + ENTITY_TYPE_SUFFIX for code in DEFAULT_SUPPORTED_COUNTRY_CODES
-        ],
-        support_international=True
+        supported_regions=SUPPORTED_REGIONS
```
SharonHart, line 29: @omri374 By default, I thought about supporting all regions in phonenumbers. The results were pretty fast still and we are removing duplicates anyways. WDYT?

presidio-analyzer/presidio_analyzer/predefined_recognizers/phone_recognizer.py
```diff
@@ -66,51 +54,34 @@ def analyze(
...
-                        ]
-
-        return results
-
-    def _get_regional_recognizer_result(self, match, entity, text, nlp_artifacts):
+        for region in self.supported_regions:
+            for match in phonenumbers.PhoneNumberMatcher(text, region, leniency=0):
+                results += [self._get_recognizer_result(match, text, region, nlp_artifacts)]
+
+        return EntityRecognizer.remove_duplicates(results)
```
SharonHart, line 61: Eventually the implementation of this method didn't allow easy aggregation of the analysis explanation to include all matched regions. Decided to not include regions in the explanation.

presidio-analyzer/presidio_analyzer/predefined_recognizers/phone_recognizer.py
SharonHart: Eventually decided to return to the hardcoded list. 
Japanese number was matched with credit card, Mongolian with US driving licence, and these are just the tests we have.

presidio-analyzer/presidio_analyzer/predefined_recognizers/phone_recognizer.py
```diff
@@ -19,22 +23,19 @@ class PhoneRecognizer(LocalRecognizer):
     :param supported_entities: The entities this recognizer can detect
     """
 
-    SCORE = 0.6
```
SharonHart, line 22: see test_analyzer_engine

presidio-analyzer/presidio_analyzer/predefined_recognizers/phone_recognizer.py
```diff
@@ -66,52 +60,36 @@ def analyze(
...
-                            self._get_regional_recognizer_result(
-                                match, entity, text, nlp_artifacts
-                            )
-                        ]
-
-        return results
-
-    def _get_regional_recognizer_result(self, match, entity, text, nlp_artifacts):
+        for region in self.supported_regions:
+            for match in phonenumbers.PhoneNumberMatcher(text, region, leniency=1):
```
SharonHart, line 64: Raised leniency since phone numbers matched '1234567', credit cards and us driving numbers. 
After raising, only some regions matched '1234567', which I excluded above (see LOW_LENIENCY_CODES)

omri374: This could have an impact with false positives. Based on your tests, was that the case? Maybe worth checking more than just 1234567 (e.g. take a few samples from a dataset). Theoretically we could keep all the changes in this PR, but still stick with the five regions we had in the first place + documentation on how to add more, right?

presidio-analyzer/presidio_analyzer/predefined_recognizers/phone_recognizer.py
```diff
@@ -66,52 +60,36 @@ def analyze(
...
+    def _get_recognizer_result(self, match, text, region, nlp_artifacts):
         number = match.number
-        main_region_code = COUNTRY_CODE_TO_REGION_CODE.get(number.country_code)[0]
         result = RecognizerResult(
-            entity_type=entity,
+            entity_type="PHONE_NUMBER",
             start=match.start,
             end=match.end,
             score=self.SCORE,
             analysis_explanation=self._get_analysis_explanation(),
```
omri374, line 78: If this method gets the region as input, could we pass it to the analysis explanation somehow? I think it would be useful but it's not mandatory.

omri374:
Thanks! Great addition and looks awesome.

presidio-analyzer/presidio_analyzer/predefined_recognizers/phone_recognizer.py
SharonHart: The purpose of raising the leniency was to avoid FPs. When it was lower, lots of regions matched random numbers, credit cards, licence numbers. Having the fixed regions will help with reducing those FPs so I might revert back to using them after all, at least as a first step.

presidio-analyzer/presidio_analyzer/predefined_recognizers/phone_recognizer.py
SharonHart: That was the initial plan. However, since more than one region can be matched on the same number or as a substring, when removing duplicates we will lose that explanation. I tried to see if this can be added to the duplicates removal method but it was too complex to add.

omri374: Even if we lose it, it's better than not having it at all in the first place, isn't it?

SharonHart: It can reach a false explanation where you tried to find US numbers but the analysis stated that it matched only CA number, maybe we can phrase something differently?

omri374 (approved):
Looks great. Should we also update the supported entities list? and how should we document the customization of this recognizer? Could definitely be in a different PR and I would be happy to help.

presidio-analyzer/presidio_analyzer/predefined_recognizers/phone_recognizer.py
omri374: If you tried to find US numbers but it matched CA numbers, I think the user would benefit from knowing that. Users could change the supported regions if they want a more fine grain explanation. WDYT?

SharonHart:
@omri374 entities list is already updated, with PHONE_NUMBER entity only. 
Where would you document the recognizer customization?

presidio-analyzer/presidio_analyzer/predefined_recognizers/phone_recognizer.py
SharonHart: Done

omri374 (approved):

navalev (approved):

PR #789, opened 2021-11-01
Analyzer and Anonymizer deps bump

omri374:
## Change Description
Updates to Pipfile and setup.py for analyzer and anonymizer


SharonHart (dismissed):

navalev (approved):

omri374:
@navalev @SharonHart after doing this I was thinking, maybe we shouldn't limit the deps versions to the maximum? We know it's working on lower versions. WDYT would be the best approach here?
For example, we had spacy version 3.0.6 first and now it's tested on 3.1.x. Maybe we should specify the version to be `spacy>=3.0.6`?

SharonHart:
@omri374 In our case, maybe locking to minor version is the best, since we saw how text, and specifically the demo one, tend to recognized differently with each version of spacy. Other packages might be more relaxed-versioned

omri374:
@SharonHart how about `spacy >=3.0.6, <=3.3`?
It works on 3.2

SharonHart:
@omri374 Missing lock file for image-redactor?

omri374:
@SharonHart yes please ignore for now. I can't seem to lock and make it work on all python versions. Will keep on trying

SharonHart:
I'll try myself tomorrow as well

SharonHart (approved):

navalev (approved):

omri374:
Closing for now to introduce some updates

PR #793, opened 2021-11-09
Bug/fix deps

SharonHart:


presidio-image-redactor/tests/integration/test_image_analyzer_engine_integration.py
```diff
@@ -24,8 +24,8 @@ def __get_expected_ocr_test_image_analysis_results():
         ImageRecognizerResult("PERSON", 31, 44, 0.85, 472, 20, 91, 31),
         ImageRecognizerResult("PERSON", 31, 44, 0.85, 576, 20, 147, 31),
         ImageRecognizerResult("DOMAIN_NAME", 303, 320, 1.0, 28, 299, 438, 38),
-        ImageRecognizerResult("PHONE_NUMBER", 332, 346, 0.7, 666, 298, 88, 40),
-        ImageRecognizerResult("PHONE_NUMBER", 332, 346, 0.7, 769, 301, 169, 29),
```
SharonHart, line 28: New phone recognizer score

presidio-analyzer/tests/test_spacy_recognizer.py
```diff
@@ -38,7 +38,7 @@ def prepare_and_analyze(nlp, recognizer, text, ents):
         ("I bought my car in May", 1, ((19, 22),), 1),
         ("May 1st", 1, ((0, 7),), 1),
         ("May 1st, 1977", 1, ((0, 13),), 1),
-        ("I bought my car on May 1st, 1977", 1, ((19, 32),), 1),
+        ("I bought my car on May 1st, 1977", 2, ((19, 26), (28, 32)), 1),
```
SharonHart, line 41: spacy now recognizes it in two parts and recognizes 1977 as a date, unfortunately. Our recognizer, which is on by default should return better results and prioritized in anonymizer normalization

omri374: spaCy gets priority because it has a higher confidence value?

omri374 (approved):

presidio-analyzer/tests/test_spacy_recognizer.py
SharonHart: I think these tests only use spacy recognizer. I mentioned that our recogznizer will possibly be prioritized over spacy's because of this split.

PR #794, opened 2021-11-09
Analyzer deps bump

omri374:


SharonHart:
@omri374 Build is now passing. 
Eventually I've realized that the timeout doesn't occur on newly uploaded model versions.
So upgrading to spacy 3.2.0 actually helped with model downloading in analyzer. 
What's left is image-redactor that resolves spacy version from analyzer (which is not yet bumped). The workaround of installing 3.0.0 was to install from tar.gz.
After merging and releasing, we can revert the tar.gz installation.

presidio-analyzer/Pipfile
```diff
@@ -4,17 +4,17 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-spacy = "==3.0.6"
+spacy = "==3.2.0"
```
omri374, line 7: Can we allow more flexibility here?

presidio-analyzer/setup.py
```diff
@@ -29,14 +29,13 @@
...
-    tests_require=["pytest", "flake8==3.7.9"],
+    tests_require=["pytest", "flake8>=3.7.9"],
     install_requires=[
-        "spacy==3.0.6",
-        "regex>=2020.11.13",
-        "tldextract==3.1.0",
-        "pyyaml==5.4.1",
-        "pydantic==1.7.4",
-        "phonenumbers==8.12.24",
+        "spacy==3.2.0",
```
omri374, line 34: here too

navalev (dismissed):

presidio-analyzer/Pipfile
SharonHart: I thought that for the dependencies changing the detections regularly, at least in the previous bumps we did, to lock it to minor.

SharonHart: Anyway, done

SharonHart (approved):

navalev (approved):

PR #797, opened 2021-11-16
Bump analyzer in image redactor

SharonHart:


omri374 (approved):

navalev (dismissed):

omri374:
Why has the lock file changed but not the pipfile? Is it just re-locking?

SharonHart:
@omri374 We defined a relaxed versioning in the pipfile for analyzer, which we bumped this week. Locking again updated the version in the lock file.

SharonHart:
@omri374 Implicitly added typing extensions so that it would be included in the lockfile for python 3.9, as I did for analyzer. Can be avoided with a different dependency installation method in the future.

omri374 (approved):

PR #806, opened 2021-11-30
Fix IbanRecognizer not recongnizing IBAN followed by all caps

vtols:
## Change Description

Fix IbanRecognizer not recongnizing IBAN followed by all caps

## Checklist

- [x] I have signed the CLA


omri374:
Thanks! We'll review it soon.

vtols:
@omri374 

Just as a reminder, can you please review this PR? Thanks

presidio-analyzer/presidio_analyzer/predefined_recognizers/iban_recognizer.py
```diff
@@ -40,7 +40,9 @@ class IbanRecognizer(PatternRecognizer):
     PATTERNS = [
         Pattern(
             "IBAN Generic",
-            r"\b([A-Z]{2}[ \-]?[0-9]{2})(?=(?:[ \-]?[A-Z0-9]){9,30})((?:[ \-]?[A-Z0-9]{3,5}){2,7})([ \-]?[A-Z0-9]{1,3})?\b",  # noqa
+            r"\b([A-Z]{2}[ \-]?[0-9]{2})(?=(?:[ \-]?[A-Z0-9]){9,30})((?:[ \-]?[A-Z0-9]{3,5}){2})"  # noqa
+            r"([ \-]?[A-Z0-9]{3,5})?([ \-]?[A-Z0-9]{3,5})?([ \-]?[A-Z0-9]{3,5})?([ \-]?[A-Z0-9]{3,5})?([ \-]?[A-Z0-9]{3,5})?"  # noqa
+            r"([ \-]?[A-Z0-9]{1,3})?\b",  # noqa
```
omri374, line 45: This seems to be a very weak regex (would catch any 1-3 word with a space before it). Is it required in your opinion for IBAN detection?

vtols: This is a regex that matches the same text as before. And both expressions can indeed match more than necessary, because there's no way to tell where is the end of IBAN. The difference between these expressions is in the capture groups. The new one has one of the capture groups expanded. This allows the algorithms implemented in the recognizer to drop the groups one by one from the matched string and check whether a reduced matched string is a valid IBAN code. This doesn't work properly with the old expression, because the algorithm only drops capture groups as a whole, but can't do if the group is just repeated with the `{n,m}` operator. That's because of how capturing in regular expressions works.

So
`(?:[ \-]?[A-Z0-9]{3,5}){2,7})`
becomes
`(?:[ \-]?[A-Z0-9]{3,5}){2})([ \-]?[A-Z0-9]{3,5})?([ \-]?[A-Z0-9]{3,5})?([ \-]?[A-Z0-9]{3,5})?([ \-]?[A-Z0-9]{3,5})?([ \-]?[A-Z0-9]{3,5})?([ \-]?[A-Z0-9]{3,5})?([ \-]?[A-Z0-9]{3,5})?([ \-]?[A-Z0-9]{3,5})?([ \-]?[A-Z0-9]{3,5})?([ \-]?[A-Z0-9]{3,5})?`

omri374: Thanks for clarifying! It makes sense. Would you mind adding a few more tests for cases that might be considered false positives?

omri374 (approved):
Did a few tests and it looks great. Thanks @vtols for the contribution

omri374:
@SharonHart @navalev could you please review?

SharonHart (approved):

PR #818, opened 2022-01-20
Including entity tag equivalences with Presidio

Idomingog:
## Change Description

Modify the entity tag names to be the same than Presidio. 
Now appears PERSON instead of PER and LOCATION instead of LOC.
I don't pass the ORG and MISC type because this entities are not supported by Presidio yet. https://microsoft.github.io/presidio/supported_entities/

Modify in the analysis explanation  results the value of 'recognizer'. Now appears 'FlairRecognizer'.

## Issue reference

This PR fixes issue #XX

## Checklist

- [ ] I have signed the CLA


omri374:
Thanks @Idomingog! We'll review it soon.

omri374:
Looks great! Regarding ORGANIZATION, you could consider keeping it. The reason it's not there in the first place is because spaCy's detection rates aren't always good enough, but Flair has better performance on this.

omri374 (approved):

PR #829, opened 2022-02-16
Fix issue when we have a PII at the end of a line, the indexes can be…

shiranr:
… the same if the word was reducted. Added tests as well

## Change Description

Fix bug when we reduct data from the end of a sentence, it had an unneeded restriction.

### DIFF
--- a/presidio-anonymizer/tests/integration/test_anonymize_engine.py
+++ b/presidio-anonymizer/tests/integration/test_anonymize_engine.py
@@ -9,6 +9,28 @@ from presidio_anonymizer.entities import (
 from presidio_anonymizer.operators import AESCipher
 
 
+def test_given_name_and_phone_number_then_we_anonymize_correctly_2():
+    text = "The url is http://microsofy.com"
+    anonymizer_config = {
+        "DOMAIN_NAME": OperatorConfig(
+            "redact"
+        ),
+        "URL": OperatorConfig(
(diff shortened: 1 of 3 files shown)


presidio-anonymizer/tests/integration/test_anonymize_engine.py
```diff
@@ -9,6 +9,28 @@
 from presidio_anonymizer.operators import AESCipher
 
 
+def test_given_name_and_phone_number_then_we_anonymize_correctly_2():
```
SharonHart, line 12: Maybe a name reflecting the scenario better?

SharonHart (approved):
LGTM, might only change the test name

presidio-anonymizer/tests/integration/test_anonymize_engine.py
shiranr: lol! 😂  missed that tnx.

shiranr: Done!

anatbal (dismissed):
LGTM, thank you for the quick fix!

SharonHart (dismissed):

SharonHart (approved):

anatbal (approved):

omri374 (approved):

PR #878, opened 2022-06-02
Batch analyze - Support for lists and dicts (partial)

omri374:
## Change Description

Added the [batch analysis](https://microsoft.github.io/presidio/samples/python/batch_processing/) into the presidio-analyzer code.

This logic can handle:
1. List of values (also leveraging spaCy `nlp.pipe`)
2. Dictionaries (including nested dictionaries) and dictionaries holding lists of primitive objects.

The logic currently cannot handle:
- Lists holding non-primitive objects, such as a list of dictionaries.
- Batch anonymization (currently exists only on the sample notebook, could be added in a future PR)

Potential future enhancements:
1. Better handling of column names / keys as context
2. Sampling from lists to understand if PII might exists in a column
3. `keys_to_include` in parallel to the existing `keys_to_skip`
4. Auto-detect potential sensitive columns
5. Support for cross-row context, using the `ad_hoc_recognizer` interface: Have value from one column (say `first_name`) be used to detect this first name on another column (say `user_comments`)
6. Sensitivity reports on entire tables

## Checklist

- [X] I have signed the CLA


presidio-analyzer/presidio_analyzer/analyzer_engine.py
```diff
@@ -223,6 +227,116 @@ def analyze(
...
+            )
+
+            list_results.append(results)
+
+        return list_results
+
+    @staticmethod
+    def _validate_type(obj: Any) -> bool:
+        if not type(obj) in (int, float, bool, str):
+            raise ValueError(
```
balteravishay, line 265: should this throw or just return False? if we throw, won't the calling code stop execution of the for loop?

presidio-analyzer/presidio_analyzer/analyzer_engine.py
```diff
@@ -223,6 +227,116 @@ def analyze(
...
+        Non-string values are returned as is.
+
+        :param input_dict: The input dictionary for analysis
+        :param language: Input language
+        :param keys_to_skip: Keys to ignore during analysis
+        :param kwargs: Additional keyword arguments
+        for the `AnalyzerEngine.analyze` method
+        """
+
+        if "context" in kwargs:
```
balteravishay, line 291: is this documented?

presidio-analyzer/presidio_analyzer/analyzer_engine.py
```diff
@@ -223,6 +227,116 @@ def analyze(
...
+    def _validate_type(obj: Any) -> bool:
+        if not type(obj) in (int, float, bool, str):
+            raise ValueError(
+                "Analyzer.analyze_iterator only works on primitive types "
+                "(int, float, bool, str). "
+                "Lists of objects are not yet supported."
+            )
+        return True
+
+    def analyze_dict(
```
balteravishay, line 272: seems like a helper method or a special-case-wrapper for the analyze-iterator. more than a "core" analyzer functionality to me, but let's discuss at the review session.

omri374: This is something the user is likely to use for any JSON file, or any tabular data that's turned into a dictionary of key:List[str]
See some examples here: https://github.com/microsoft/presidio/blob/2c68459bdd9f85c09dfe82277b006d14da42e0ff/docs/samples/python/batch_processing.ipynb

Documentation is still missing, wanted to get everybody's feedback before finalizing it.

presidio-analyzer/presidio_analyzer/analyzer_engine.py
```diff
@@ -132,6 +133,7 @@ def analyze(
         return_decision_process: Optional[bool] = False,
         ad_hoc_recognizers: Optional[List[EntityRecognizer]] = None,
         context: Optional[List[str]] = None,
+        nlp_artifacts: Optional[NlpArtifacts] = None,
```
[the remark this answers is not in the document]
omri374: I don't think so. It's an object which contains the spaCy output, not something someone would send over REST IMHO

presidio-analyzer/presidio_analyzer/analyzer_engine.py
omri374: Yes, the idea is to stop the execution if the user sends something we cannot analyze. I was trying to find a better way to do this (without the return statement) but couldn't find a good approach, since this is all iterators. Any suggestions?

presidio-analyzer/presidio_analyzer/analyzer_engine.py
omri374: Not yet, will be after the internal review

evgri243:
Hey! Looks amazing! I'll try to give it a try next week (wonder how to publish it to the compliant until it has been built and packaged; have some ideas). 

My use case is quite simple now:
* take pandas data frame, extract a column (email body in my case) as a list and run analyzer on it in a batch with spacy.require_gpu enabled. Take the collection of results.
* apply anonymizer on the original text with that collection. I don't see any issues with the anonymizer yet in my hot paths, so that is more of a question of some convenient API.

presidio-analyzer/presidio_analyzer/analyzer_engine.py
shiranr: You don't need to return anything from this. Just do raise or return void.

presidio-analyzer/presidio_analyzer/analyzer_engine.py
shiranr: I wanted to ask about it as well :) the kwargs seems to me too generic in the method doc

shiranr, line 291: I would suggest:
context = []
if "context" in kwargs ... 
instead of having if else.

presidio-analyzer/presidio_analyzer/analyzer_engine.py
omri374: The function is called in a way which would only return values for which this returns True. I tried playing around with it but couldn't find a good solution. It's probably my lack of knowledge of Python generators TBH.

presidio-analyzer/presidio_analyzer/analyzer_engine.py
```diff
@@ -223,6 +227,116 @@ def analyze(
...
+            elif isinstance(value, Iterable):
+                # Recursively iterate nested dicts
+
+                results: List[List[RecognizerResult]] = self.analyze_iterator(
+                    list_of_texts=value,
+                    language=language,
+                    context=specific_context,
+                    **kwargs,
+                )
+            else:
```
shiranr, line 335: I really want to simplify this a little in terms of code...
thinking of ideas :)

presidio-analyzer/presidio_analyzer/analyzer_engine.py
shiranr: Well, truthfully, looking at the code, you can just do range... 
something like:
validate(list_of_texts)
that will do range of the list_of_texts and won't return anything as the _validate_type is not really filtering anything and it's a little confusing to do it with:
list_of_texts = (text for text in list_of_texts if self._validate_type(text))
As it seems as though it is actually filtering something but it's not it's abruptly raise an exception

presidio-analyzer/presidio_analyzer/analyzer_engine.py
```diff
@@ -223,6 +227,116 @@ def analyze(
...
+        """
+
+        if "context" in kwargs:
+            context = kwargs["context"]
+            del kwargs["context"]
+        else:
+            context = []
+
+        if not keys_to_skip:
+            keys_to_skip = []
```
shiranr, line 298: Why isn't this the default value the method receives?

omri374: You mean why we don't pass keys_to_skip = [] as default? https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

presidio-analyzer/presidio_analyzer/analyzer_engine.py
omri374: `list_of_texts` is not really a list, it's an iterator.

presidio-analyzer/presidio_analyzer/analyzer_engine.py
omri374: How would you suggest to pass those arguments? The other option is to introduce all the arguments in `.analyze`, also those introduced in the future. `kwargs` allows us to be dynamic, but if it makes things more difficult maybe we should consider

presidio-analyzer/presidio_analyzer/analyzer_engine.py
omri374: I'll think of other ways :)

shiranr: ohhh another option is to do it as part of going through the iterator in analyze_dict method which is perfectly fine to do validation on each and throwing an exception if needed.

presidio-analyzer/presidio_analyzer/analyzer_engine.py
shiranr: One option is to create a class for it with the parameter and a dict in addition to the permanent params you know must be there.

presidio-analyzer/presidio_analyzer/analyzer_engine.py
shiranr: Booooo

presidio-analyzer/presidio_analyzer/analyzer_engine.py
```diff
@@ -223,6 +227,116 @@ def analyze(
...
+        for key, value in input_dict.items():
+            if key in keys_to_skip:
+                yield DictAnalyzerResult(key=key, value=value, recognizer_results=[])
+                continue  # skip this key as requested
+
+            # Add the key as an additional context
+            specific_context = context[:]
+            specific_context.append(key)
+
+            if not value:
```
shiranr, line 309: This can be if not value or key in keys_to_skip ... 
instead of having it above..

presidio-analyzer/presidio_analyzer/analyzer_engine.py
shiranr: The other option is just to make sure we document well the arguments which go into the kwargs as part of the methods doc string

omri374: I will refactor to make the `AnalyzerEngine` injected with its own parameters, instead of asking the user to pass those through `analyze_dict`

omri374: Fixed

omri374: Fixed

presidio-analyzer/presidio_analyzer/analyzer_engine.py
omri374: Fixed, moved to an external method which doesn't return True or does filtering

presidio-analyzer/presidio_analyzer/analyzer_engine.py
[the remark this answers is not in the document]
omri374: As discussed, this isn't something the user is likely to pass through a REST API. How would you suggest to document it aside from a docstring update?

presidio-analyzer/presidio_analyzer/analyzer_engine.py
omri374: Fixed

presidio-analyzer/presidio_analyzer/analyzer_engine.py
omri374: Thanks for the brainstorming session!

presidio-analyzer/presidio_analyzer/analyzer_engine.py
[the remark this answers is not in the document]
balteravishay: right, since the docs are auto generated from the docstring i guess this is fine!

balteravishay (dismissed):
Great stuff!

balteravishay (approved):

shiranr (approved):

PR #884, opened 2022-06-27
Bug/missing data on batch

omri374:
Bugfix on recent change in batch analyzer. Yielding the val even if it's empty.

If there is a missing value, we would still want Presidio to return an output for it. Otherwise, we'll have a mismatch between input and output. For example, if we pass a list of four and one of the values is None, we'd like to get an output of size four, where the output for the None is empty.

**Important**:
Note that we would still have an issue with the anonymizer in this case, as it is throwing an exception for empty values. This would be fixed in a later stage.


presidio-analyzer/presidio_analyzer/batch_analyzer_engine.py
```diff
@@ -127,9 +127,7 @@ def analyze_dict(
     @staticmethod
     def _validate_types(value_iterator: Iterable[Any]) -> Iterator[Any]:
         for val in value_iterator:
-            if not val:
-                continue
-            if not type(val) in (int, float, bool, str):
+            if val and not type(val) in (int, float, bool, str):
```
shiranr, line 130: there is a yield val at the bottom, are you sure you want to do it if val is empty?

omri374: Yes, that's the issue with the previous logic. If there is no val, we still want to return a value for it. Otherwise we'll have a mismatch between input and output. For example, if we pass a list of 4 and one of the values is None, we'd like to get an output of size 4, where the output for the None is empty.

shiranr: Perhaps add this to the commit description :)

shiranr (approved):

presidio-analyzer/presidio_analyzer/batch_analyzer_engine.py
omri374: Good idea :)

omri374: Done.

evgri243 (approved):

Issue #887, opened 2022-06-29
Transformers based NLP engine


omri374:
## Background
Presidio currently leverages spaCy for NER. It is possible to [switch to a stanza model](https://microsoft.github.io/presidio/tutorial/05_languages/), or to create additional NER recognizers using 3rd party packages such as Flair and Transformers (see examples [here](https://github.com/microsoft/presidio/blob/main/docs/samples/python/flair_recognizer.py) and [here](https://github.com/microsoft/presidio/blob/main/docs/samples/python/transformers_recognizer.py)).
However, it is not possible to completely replace a spacy/stanza model with a transformers model, except for when using the `trf` models coming from spaCy, such as `en_core_web_trf`. 

## Proposed solution

Create a new type of `NlpEngine` called `TransformersNlpEngine` which runs a transformers NER model instead of spaCy's NER. The user could decide which transformers model to plug in (either pretrained or custom). 

Since Presidio leverages spaCy's pipeline for different features such as lemmas and tokens, which aren't supported by a transformers NER model, we propose to create a [spaCy pipeline](https://spacy.io/usage/processing-pipelines) without NER, and add a new component which runs a transformers NER model.

Instead of this pipeline which we currently have:
![image](https://user-images.githubusercontent.com/3776619/176368408-d513a6d7-7f0e-4ffa-b4b6-7a598f30025f.png)

We would have this pipeline:
![image](https://user-images.githubusercontent.com/3776619/176369943-94080c2d-1442-49c7-93cf-41db08fa4a82.png)

This pipeline would not be trainable, but one can train a transformers model and inject it into this pipeline

matthewchung74:
Hi @erann1987 , is there any documentation on how to add a transformer model, as mentioned here? https://github.com/microsoft/presidio/blob/37f74e8e880cb1bdf3f5224a05eaa9b63df02d31/presidio-analyzer/presidio_analyzer/nlp_engine/transformers_nlp_engine.py#L89

PR #912, opened 2022-09-10
Install transformers model into the docker image

omri374:
## Change Description

Extended `install_nlp_models.py` to include transformers models based on config
Added a new Dockerfile which is tailored for transformers models

>Note: as part of this PR I had to refresh all Pipfile.lock files, which resulted in changes to tests due to changes in spaCy outputs and other dependencies

Main changes to look at:
1. `install_nlp_models.py`
2. `Dockerfile.transformers`
3. CI/CD changes


presidio-analyzer/Dockerfile.transformers
```diff
@@ -0,0 +1,22 @@
...
+ARG NAME
+ARG NLP_CONF_FILE=conf/transformers.yaml
+ENV PIPENV_VENV_IN_PROJECT=1
+ENV PIP_NO_CACHE_DIR=1
+WORKDIR /usr/bin/${NAME}
+
+COPY ./Pipfile* /usr/bin/${NAME}/
+RUN pip install pipenv \
+  && pipenv sync
+RUN pipenv install torch transformers huggingface_hub --skip-lock
```
omri374, line 12: Added specific installation for torch, transformers and huggingface_hub

omri374, line 12: This is the main change between the transformers version and the previous one. Any ideas on how we can have those under one Dockerfile?

tamirkamara: Do you mean a single file that can produce 2 images?

tamirkamara: So a build arg that would conditionally run a line? Some like this?
https://github.com/microsoft/AzureTRE/blob/main/.devcontainer/Dockerfile#L45

omri374: Yes. Thanks for the reference

balteravishay: Personally I think that is not a best practice, since the produced docker image is not consistent.

tamirkamara: @balteravishay then any build arg will produce an inconsistent image...

tamirkamara: Another, more elaborate way would be multistage with "conditional" stages.

balteravishay: I'm not sure multistage with "conditional" stages is supported without docker buildkit.

tamirkamara: And that's not something you want to use?

tamirkamara: At and case it would be an arg of some sort that would change how the image is built. So at the base of it, an input can produce different images but I think it's pretty common scenario.

omri374: What's the verdict? Having one Dockerfile that could produce images? or having two `Dockerfile` and `Dockerfile.transformers` files?

SharonHart: Its a good practice to not use conditions in the dockerfile (use one dockerfile) since you are not using the layer caching option. 
What you can do have a base dockerfile and extend it to create flavours of it

balteravishay: talked with @tamirkamara and we can proceed with having a single dockerfile, and using buildkit.

omri374: Thanks. That's quite new to me. WDYT about the existing solution (two dockerfiles) as a first step, and introducing buildkit as a second? Not optimal, only temporary.

e2e-tests/tests/test_analyzer.py
```diff
@@ -8,7 +8,7 @@
 def test_given_a_correct_analyze_input_then_return_full_response():
     request_body = """
     {
-        "text": "John Smith drivers license is AC432223",
+        "text": "John Smith drivers license is AC532223",
```
omri374, line 11: Updates to the spacy model required a change in tests

presidio-image-redactor/tests/integration/test_image_analyzer_engine_integration.py
```diff
@@ -20,12 +20,21 @@ def test_given_image_without_text_then_no_entities_recognized(image_analyzer_eng
...
+    # fmt: off
     return [
-        ImageRecognizerResult("PERSON", 31, 44, 0.85, 472, 20, 91, 31),
-        ImageRecognizerResult("PERSON", 31, 44, 0.85, 576, 20, 147, 31),
-        ImageRecognizerResult("DOMAIN_NAME", 303, 320, 1.0, 28, 299, 438, 38),
-        ImageRecognizerResult("PHONE_NUMBER", 332, 346, 0.75, 666, 298, 88, 40),
-        ImageRecognizerResult("PHONE_NUMBER", 332, 346, 0.75, 769, 301, 169, 29),
-        ImageRecognizerResult("EMAIL_ADDRESS", 772, 794, 1.0, 27, 912, 458, 39),
-        ImageRecognizerResult("DOMAIN_NAME", 781, 794, 1.0, 27, 912, 458, 39),
+        ImageRecognizerResult(entity_type="PERSON", start=31, end=44,
```
omri374, line 25: Updates to tests due to the change in the spacy model

.pipelines/templates/e2e-tests.yml
```diff
@@ -29,7 +29,7 @@ steps:
           python -m pip install --upgrade pip
           pip install -r requirements.txt
           pip install pytest-azurepipelines
-          python -m spacy download en_core_web_sm
+          python -m spacy download en_core_web_lg
```
navalev, line 32: the usage of th large package is on purpuse?

omri374: Yes, to simulate the real instance

navalev (approved):

presidio-analyzer/tests/conftest.py
```diff
@@ -52,6 +52,15 @@ def nlp_engines(request, nlp_engine_provider) -> Dict[str, NlpEngine]:
     for name, engine_cls in nlp_engines.items():
         if name == "spacy" and not request.config.getoption("--runfast"):
             available_engines[f"{name}_en"] = engine_cls({"en": "en_core_web_lg"})
+        elif name == "transformers" and not request.config.getoption("--runfast"):
+            available_engines[f"{name}_en"] = engine_cls(
+                {
+                    "en": {
+                        "spacy": "en_core_web_lg",
```
SharonHart, line 59: Shouldn't it be a spacy transformer model as well? `en_core_web_trf`?

SharonHart (approved):
Left a question

presidio-analyzer/tests/conftest.py
omri374: The `trf` models are standard spaCy models, so the `huggingface` option is not required for them. This PR would handle non-spaCy transformers models.

omri374: Also, the spaCy model here is only used as a pipeline, not as a NER model. See here: https://github.com/microsoft/presidio/issues/887

Issue #956, opened 2022-11-23
Improve performance in recognizer with similar regexes


aperezfals:
There are some implementations of `PatternRecognizer` that export several regexes in the patterns array because the small difference between the regexes influences the score. An example of this is `UsItinRecognizer`. 

```python
PATTERNS = [
    Pattern(
        "Itin (very weak)",
        r"\b9\d{2}[- ](5\d|6[0-5]|7\d|8[0-8]|9([0-2]|[4-9]))\d{4}\b|\b9\d{2}(5\d|6[0-5]|7\d|8[0-8]|9([0-2]|[4-9]))[- ]\d{4}\b",  # noqa: E501
        0.05,
    ),
    Pattern(
        "Itin (weak)",
        r"\b9\d{2}(5\d|6[0-5]|7\d|8[0-8]|9([0-2]|[4-9]))\d{4}\b",  # noqa: E501
        0.3,
    ),
    Pattern(
        "Itin (medium)",
        r"\b9\d{2}[- ](5\d|6[0-5]|7\d|8[0-8]|9([0-2]|[4-9]))[- ]\d{4}\b",  # noqa: E501
        0.5,
    ),
]
```

The difference between those 3 Patterns is the separators `-` and space in an ITIN, matching the patterns `9XX-XX-XXxX`, `9XX-XXXXXX` , `9XXXX-XXXX` or `9XXXXXXXX`. The 3 regexes are necessary for the current implementation because they represent different pattern names and scores.

Having multiple regexes is not optimal, since every regex needs to match every text that needs to be analyzed. Reducing the number of regexes while keeping different pattern names and scores would improve the Presidio performance.

**Proposal**

- Add a `get_pattern_from_match(pattern: Pattern, regex_match) -> Pattern` method to `PatternRecognizer`
- This method will return by default in `PatternRecognizer` the same pattern passed in parameters
- It can be overridden by specific recognizers, like `UsItinRecognizer`
- Some recognizers now can reduce the number of regexes and implement `get_pattern_from_match`. Let's take a look at `UsItinRecognizer`:
  - The 3 regexes can be reduced to one, that matches the separators `[- ]` in regex named groups
  - `r"\b9\d{2}(?P<firstSeparator>[- ])?(5\d|6[0-5]|7\d|8[0-8]|9([0-2]|[4-9]))(?P<secondSeparator>[- ])?\d{4}\b"`
[...]

PR #959, opened 2022-12-01
Get patterns from regex match in ITIN recognizer

aperezfals:
## Change Description

Adds a new method `get_pattern_from_match(self, pattern: Pattern, match: Match) -> Pattern` to `PatternRecognizer` that returns a new `Pattern` based on the regex match info.

The method `get_pattern_from_match` method returns the same detected pattern in `PatternRecognizer` to not affect current recognizer implementations. In `UsItinRecognizer` is overridden, reducing the regex used from 3 to only 1. The regex in `UsItinRecognizer` use now named groups for the separators `-` and ` `. The implementation of `get_pattern_from_match` in `UsItinRecognizer` uses the named groups to return different pattern names and scores.

## Issue reference
https://github.com/microsoft/presidio/issues/956

Future PRs will reduce the number of regex in the recognizers list defined in the issue.

## Checklist

- [ ] Add default implementation of `get_pattern_from_match(self, pattern: Pattern, match: Match) -> Pattern` in  `PatternRecognizer`
- [ ] Call `get_pattern_from_match` in `PatternRecognizer`
- [ ] Reduce the number of regexes in `UsItinRecognizer`
- [ ] Override `get_pattern_from_match` in `UsItinRecognizer`
- [ ] Add more unit tests for space separators


presidio-analyzer/presidio_analyzer/predefined_recognizers/us_itin_recognizer.py
```diff
@@ -48,3 +38,25 @@ def __init__(
...
+            return Pattern(
+                "Itin (weak)",
+                pattern.regex,
+                0.3
+            )
+
+        return Pattern(
+            "Itin (very weak)",
+            pattern.regex,
+            0.05
```
omri374, line 61: This is a great addition but could be confusing from a flow perspective, as we alrady define the score on the list of patterns, or we accept patterns from the user whose score would be overwritten in this method. 

Any ideas on how to make the flow simpler while still keeping the functionality? Maybe have a confidence boost param, similar to what we have with context scores, which would be added to the original score? 
Furthermore, we should also think about the affect it has on the analysis explanation.

presidio-analyzer/presidio_analyzer/pattern_recognizer.py
```diff
@@ -230,6 +233,18 @@ def __analyze_patterns(
         results = EntityRecognizer.remove_duplicates(results)
         return results
 
+    def get_pattern_from_match(
+        self, pattern: Pattern, match: re.Match
+    ) -> Pattern:
+        """
+        Return a new Pattern based on the matched regex info e.g., the named groups in the regex.
+
```
omri374, line 241: Can we add "by default, would return the input pattern.", or something similar?

omri374 (commented):
Thanks! This is a great addition. Added some comments with questions

presidio-analyzer/presidio_analyzer/predefined_recognizers/us_itin_recognizer.py
aperezfals: You are right, if the user needs to add custom patterns, it was not simple. If we pass to the pattern a function to improve it (return a new pattern) based on the regex match, then any user can actually use this functionality. We can think about it like if the regex has a default name and score, that can change based on the match.

This right now doesn't affect the analysis explanation, since this only return the same patterns as before, just in a different way. This is done before the validate_result step.

aperezfals: The PR now is updated with this change: moving the improve pattern function to the pattern class

SharonHart: Another option might be to enhance `pattern_recognizer.py` to have a scoring enhancement / decrease / increase methods that are altering scores as part of postprocessing. Similarly to the existing binary `validate_result` and `invalidate_result` methods. The calculation is totally custom and can be with regex or plain string operation. Might be less optimal time-wise since here you already found all the capture groups, but a more generic solution that can use different approach besides regex for altering the score

omri374: I agree with @SharonHart. I like this approach but there's a tradeoff between complexity and performance improvement. Sharon's approach is simpler in my opinion, and can generalize to other improvements/score changes that are not based on capture groups.

aperezfals: @SharonHart if we implement a method for score enhancement at the recognizer level, we should keep in mind the following
1. If the score enhancement works for some patterns in the recognizer, but not for others, it should discriminate using the pattern that matched the text. This makes any score enhancement method more complex. Keeping the score enhancement method on Pattern level will avoid that. This actually decreases the complexity, and of course, improves the performance.
  - Lets take as example the current `date_recognizer`. There are some patterns that recognize dates in the form `mm/yyyy` and others in the form `dd/mm`. Lets think that we want to improve the score by checking that the year is not less than 1600 (just as example) and also the day should be in a number range acording to the month.  If we want to implement this 2 enhancements at recognizer level, we need to put if statements on this enhace method to check what pattern detected the date, in order to apply 1 enhancement or the other.
 ```python
def enhance_score(...):
   if patter_detected_is_dd_mm:
      enhance_score_by_day_check()
   elif patter_detected_is_mm_yyyy:
      enhance_score_by_year_check()
```

2. In order to implement a score enhancement method, the user will need to inherit from the recognizer class and customize the presidio registry to replace the recognizer that he is using. Keeping the score enhancement at Pattern level will avoid that. It is only a matter of passing a score enhancement method to his own pattern list. This actually decreases the complexity.

3. To keep a score enhancement method on the Pattern level, will NOT oblige the user at all to base the enhancement on the regex match info (like matched groups). He can do there any kind of string treatment or analysis of any kind, by keeping control of exactly which regex matched the text, becuase is implemented on the Pattern level. It will make easier also to give an analysis explanation.

4. There are score enhancements that should be applied to the all recognizer, no matter which Pattern detects the text. This is the case of the current validate and invalidate methods. In the end, those 2 are specific cases of enhancements, that return maximum and minimum scores. That is why, it should be good to have also a score enhancement at the recognizer level, that returns not only minimun and maximum scores, but custom ones. A default implementation of this enchance method it could be to check for pattern enhancement availabe in the current match. As @aflores-qb said, this is a good mix between both approaches, that keeps a clean flow of how to implement a pattern recognizer. 

**Based on this, I would think in the following approach:**
- Implement an `enhance_score(pattern: Pattern, match: Match, analysis_explanation: AnalysisExplanation)`  method on `PatternRecognizer`. This should modify the analysis_explanation score and add custom explanation lines. The default implementation is to check if the Pattern that matched the regex can be enhanced. This could be inherited in order to give custom enhancements on recognizer level.
- For now, lets keep the current `validate_result` and `invalidate_result`. In future PRs we can move the implementation of those in presidio default recognizers, to the new enhance method. When that is done, we can start a deprecation process for them. To keep them toguether with an enhance score method could be missleading.
- Add an enhance method on the Pattern class. A Pattern will have a property `enhance_score_func: Optional[Callable[(Pattern, Match, AnalysisExplanation), None]`. This method will be called by the default implementation of `enhance_score` on `PatternRecognizer`. This method if passed to the pattern, will modify the analysis explanation based on the specific  pattern that matched.

omri374: Hi @aperezfals, apologies for the delay! We will be happy to continue the discussion after the holidays.

SharonHart: I'm good with the suggestion 👍

omri374: Me too! Thanks for proposing this.

omri374:
@aperezfals, thanks for the update of the PR! I might be missing something, but could you elaborate on why this change cannot be added to the existing `PatternRecognizer` and is not compatible with the existing flow?

omri374:
Thanks @aperezfals. Some considerations I'm thinking of:
1. We should handle backward compatibility. If a user has a custom recognizer with an implementation of `validate`, and we decide to deprecate it, their code would not be called.
2. To me, `validate` and `invalidate` are a deterministic method that should return True or False. Checksum is one example of validation. This is different from improvement which is a way to enhance the score (like context enhancement). Having all three might be confusing, but I do see them serving different purposes. Happy to get your thoughts and @SharonHart's too.

omri374 (commented):
Adding my review on the code. Hopefully we can have this resolved (together with the previous comment on adding vs. modifying) and released in the next version!

iukea1:
This would be nice

Issue #970, opened 2022-12-13
Add evaluation code for DICOM image de-identification


niwilso:
**Is your feature request related to a problem? Please describe.**
When experimenting with the `DicomImageRedactorEngine`, it would be great to be able to automatically evaluate the performance. Metrics such as precision and recall (with the identified PHI) could be very helpful.

**Describe the solution you'd like**
Add new functionality to `DicomImageRedactorEngine` or create a new class to host all code necessary to automate evaluation of DICOM image de-identification.

A template for the ground-truth file should also be provided.

**Describe alternatives you've considered**
Evaluation can be done manually, but it is tedious and not scalable.

niwilso:
Currently working on addressing this. Will create a PR when ready.

niwilso:
Need to merge in a PR to address a limitation with `ImagePiiVerifyEngine` before creating PR for `DicomImagePiiVerifyEngine`.

PR #979, opened 2022-12-16
Add evaluation code for DICOM de-identification

niwilso:
## Change Description

This Pull Request introduces an evaluation framework for the DICOM de-identification feature in the `presidio-image-redactor` module.

### Major Changes
* Adding the `DicomImagePiiVerifyEngine` class and tests
* Adding example notebook on running evaluation
* Adding documentation on evaluation methodology (including ground truth template)

## Issue reference

This PR fixes issue #970 

## Checklist


presidio-image-redactor/tests/integration/test_dicom_image_pii_verify_engine_integration.py
```diff
@@ -0,0 +1,132 @@
...
+    Args:
+        mock_engine (DicomImagePiiVerifyEngine): Instantiated engine.
+        mock_instance (pydicom.dataset.FileDataset): Loaded DICOM.
+        mock_results (dict): Dictionary with loaded results.
+    """
+    # Assign
+    expected_ocr_results = mock_results["ocr_results_formatted"]
+    expected_analyzer_results = mock_results["analyzer_results"]
+    expected_ocr_results_labels = []
+    for i in expected_ocr_results:
```
SharonHart, line 69: for item in expected_ocr_results:

SharonHart: or even better `expected_ocr_results_labels = [item["label"] for item, in expected_ocr_results]`

presidio-image-redactor/tests/integration/test_dicom_image_pii_verify_engine_integration.py
```diff
@@ -0,0 +1,132 @@
...
+    DicomImagePiiVerifyEngine,
+)
+import pytest
+
+SCRIPT_DIR = os.path.dirname(__file__)
+PADDING_WIDTH = 25
+
+
+@pytest.fixture(scope="module")
+def mock_engine():
```
SharonHart, line 22: That way, it's instantiated once for all test cases right? So one can affect the other? (although methods are immutable which is great)
I would just instantiate it once in each test on_test = DicomImagePiiVerifyEngine()

presidio-image-redactor/tests/test_dicom_image_pii_verify_engine.py
```diff
@@ -0,0 +1,929 @@
...
+@pytest.fixture(scope="module")
+def mock_gt_single(mock_results: dict):
+    """Ground truth for a single instance"""
+    gt = mock_results["ground_truth"]
+    return gt
+
+
+# ------------------------------------------------------
+# DicomImagePiiVerifyEngine.__init__()
+# ------------------------------------------------------
```
SharonHart, line 57: implied from the test name / Act line, don't see a reason for a comment

presidio-image-redactor/tests/test_dicom_image_pii_verify_engine.py
```diff
@@ -0,0 +1,929 @@
...
+    assert mock_rescale_array.call_count == 1
+    assert mock_save_pixel_array.call_count == 1
+    assert mock_image_open.call_count == 1
+    assert mock_add_padding.call_count == 1
+    assert mock_get_metadata.call_count == 1
+    assert mock_make_phi_list.call_count == 1
+    assert mock_patternrecognizer.call_count == 1
+    assert mock_perform_ocr.call_count == 1
+    assert mock_analyze.call_count == 1
+    assert mock_verify.call_count == 1
```
SharonHart, line 147: What is it testing? and how will you enforce future addition of new method calls? Worth adding that one is called with the others return value?

presidio-image-redactor/presidio_image_redactor/dicom_image_pii_verify_engine.py
```diff
@@ -0,0 +1,336 @@
...
+                        other_width = results[other]["width"]
+                        other_height = results[other]["height"]
+
+                        match_left = abs(i_left - other_left) <= dup_pix_tolerance
+                        match_top = abs(i_top - other_top) <= dup_pix_tolerance
+                        match_width = abs(i_width - other_width) <= dup_pix_tolerance
+                        match_height = abs(i_height - other_height) <= dup_pix_tolerance
+                        matching = [match_left, match_top, match_width, match_height]
+
+                        if False not in matching:
```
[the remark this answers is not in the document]
niwilso: This is checking that all the items in `matching` are True. I will update this to `if all(matching)`.

presidio-image-redactor/presidio_image_redactor/dicom_image_pii_verify_engine.py
```diff
@@ -0,0 +1,336 @@
...
+        analyzer_results = self.image_analyzer_engine.analyze(
+            image, ad_hoc_recognizers=[deny_list_recognizer], **kwargs
+        )
+
+        # Get image with verification boxes
+        verify_image = self.verify(
+            image, ad_hoc_recognizers=[deny_list_recognizer], **kwargs
+        )
+
+        return verify_image, ocr_results, analyzer_results
```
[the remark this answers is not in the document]
niwilso: Good catch. Thank you

presidio-image-redactor/presidio_image_redactor/dicom_image_pii_verify_engine.py
```diff
@@ -0,0 +1,336 @@
...
+        # Remove duplicates
+        for dup in dups:
+            results_no_dups.pop(dup)
+
+        return results_no_dups
+
+    @staticmethod
+    def _match_with_source(
+        all_pos: dict, phi_source_dict: dict, detected_phi: dict, tolerance: int = 50
+    ) -> dict:
```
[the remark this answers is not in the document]
niwilso: Thank you for catching this. Changing to `Tuple[dict, bool]` to reflect the output.

presidio-image-redactor/presidio_image_redactor/dicom_image_pii_verify_engine.py
```diff
@@ -0,0 +1,336 @@
...
+            tolerance (int): Tolerance for exact coordinates and size data.
+
+        Return:
+            all_pos (dict): Results dict with all positives, labeled.
+        """
+        all_pos = {}
+
+        # Cycle through each positive (TP or FP)
+        for i in analyzer_results:
+            analyzer_result = analyzer_results[i]
```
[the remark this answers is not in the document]
niwilso: Doing so just returns the key. For example
```python
tmp = {"1": {"2": [1,2,3], "3": [4,5,6]}}
for analyzer_result in tmp:
    print(analyzer_result)
```
Prints out 1 and not the contents of 1.

presidio-image-redactor/tests/integration/test_dicom_image_pii_verify_engine_integration.py
niwilso: Doing so just returns the key for item. For example
```python
tmp = {"1": {"2": [1,2,3], "3": [4,5,6]}}
for item in tmp:
    print(item)
```
Prints out 1 and not the contents of 1.

presidio-image-redactor/tests/integration/test_dicom_image_pii_verify_engine_integration.py
niwilso: As you mentioned, the methods are immutable so I believe having the one object should not be an issue.

I used this approach of instantiating once and passing `mock_engine` to each of the tests to follow what we did for the `DicomImageRedactorEngine` integration test.

Let's ask for another opinion. @omri374 , thoughts?

presidio-image-redactor/tests/test_dicom_image_pii_verify_engine.py
niwilso: I just put these in to help me quickly visually identify where tests start and end (easier to look at). It's more helpful for when a single method has both happy path and exception tests.

Since this is a test file, is it alright if we keep these in?

presidio-image-redactor/tests/test_dicom_image_pii_verify_engine.py
niwilso: To be honest, this unit test does not test much given that `.verify_dicom_instance()` mostly consists of just calling other functions. I made this to help with code coverage.

The true testing of the functionality of `.verify_dicom_instance()` as a whole occurs in the integration test.

presidio-image-redactor/tests/test_dicom_image_pii_verify_engine.py
```diff
@@ -0,0 +1,929 @@
...
+    """Test happy path for DicomImagePiiVerifyEngine._calculate_recall
+
+    Args:
+        mock_engine (DicomImagePiiVerifyEngine): Instantiated engine.
+        gt_labels (dict): Ground truth labels for single instance.
+        all_pos (dict): All positives (matched detected PHI).
+        expected_result (float): Expected precision.
+    """
+    # Act
+    test_recall = mock_engine._calculate_recall(gt_labels, all_pos)
```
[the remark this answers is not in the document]
niwilso: Yes, I test private methods as well considering they are code I write which could have mistakes by accident if not tested.

The test for `_calculate_recall` is the last test in this file.

docs/samples/python/sample_data/ground_truth.json
```diff
@@ -0,0 +1,143 @@
+{
+    "sample_data/0_ORIGINAL.dcm": {
+        "0": {
```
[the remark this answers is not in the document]
niwilso: Good point. I honestly just forgot that using a list of objects was an option.

I'll see if I can update the code to use lists of objects instead of 100% dictionaries. This may result in some changes in the `DicomImageRedactorEngine` files as well.

presidio-image-redactor/presidio_image_redactor/dicom_image_pii_verify_engine.py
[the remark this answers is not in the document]
niwilso: Disregard. I have switched over to using lists instead of a dictionary so this now works.

docs/samples/python/sample_data/ground_truth.json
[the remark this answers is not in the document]
niwilso: Everything updated! Now using lists of dictionaries instead of keeping manual string indexes in a larger dictionary.

presidio-image-redactor/tests/integration/test_dicom_image_pii_verify_engine_integration.py
niwilso: Disregard. I have switched over to using lists instead of a dictionary so this now works.

presidio-image-redactor/presidio_image_redactor/dicom_image_pii_verify_engine.py
```diff
@@ -0,0 +1,336 @@
...
+        """
+        dups = []
+        results_no_dups = results.copy()
+
+        # Check for duplicates
+        for i in results:
+            i_left = results[i]["left"]
+            i_top = results[i]["top"]
+            i_width = results[i]["width"]
+            i_height = results[i]["height"]
```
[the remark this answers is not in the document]
niwilso: I am not familiar with how to use dunders in this case. Would you mind sharing an example and/or resource?

presidio-image-redactor/presidio_image_redactor/dicom_image_pii_verify_engine.py
```diff
@@ -0,0 +1,336 @@
...
+                        other_left = results[other]["left"]
+                        other_top = results[other]["top"]
+                        other_width = results[other]["width"]
+                        other_height = results[other]["height"]
+
+                        match_left = abs(i_left - other_left) <= dup_pix_tolerance
+                        match_top = abs(i_top - other_top) <= dup_pix_tolerance
+                        match_width = abs(i_width - other_width) <= dup_pix_tolerance
+                        match_height = abs(i_height - other_height) <= dup_pix_tolerance
+                        matching = [match_left, match_top, match_width, match_height]
```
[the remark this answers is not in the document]
niwilso: If one box is contained within the other, then the smaller one does not need to exist. I'll add this capability tomorrow.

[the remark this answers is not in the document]
niwilso: Also, I am not sure if I understand your second comment. Would you mind elaborating?

presidio-image-redactor/presidio_image_redactor/dicom_image_pii_verify_engine.py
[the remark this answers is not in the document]
SharonHart: In case we want to check for BBs equality or that one is contained within the other, we can implement the `__equals__` and `__container__` methods respectfully.

presidio-image-redactor/presidio_image_redactor/dicom_image_pii_verify_engine.py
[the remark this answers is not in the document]
SharonHart: Regarding to the second comment, the terminology and functionality is not necessarily related to DICOM or the verification engine, but to the more generic case, thus I would place it in a different module. My second ask is how do we remove duplicates in the non-DICOM case (the existing code). Can we use the same logic if its implemented? Or improve it if not, by extracting into a common module method?

presidio-image-redactor/tests/test_dicom_image_pii_verify_engine.py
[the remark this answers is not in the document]
SharonHart: Personally, I am not testing private methods. The reasoning behind it is that only public methods are accessible to a client and act as the contract with it. Altering them is considered as a breaking change, but the inner implementation and the use of private methods can change without affecting the functionality. IMO it should be tested the same, meaning changing privates should not affect or break tests (clients) but changing public should. Natively, is some languages, privates are not even accessible and can't be mocked without special libraries (Java needs reflection for example)

presidio-image-redactor/presidio_image_redactor/dicom_image_pii_verify_engine.py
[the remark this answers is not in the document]
SharonHart: Right, my bad. In case we could not change the payload, like if it would come from an API, a possible solution would be  using `for analyzer_result in analyzer_results.values():`
Ultimately not using the keys at all would be a 'smell' hinting we can change the structure (as you did)

presidio-image-redactor/tests/integration/test_dicom_image_pii_verify_engine_integration.py
SharonHart: This might be fragile as we can't ensure that the next test won't be mutable, and be a hell to debug :)
https://stackoverflow.com/questions/36579835/shared-context-between-net-unit-test-classes
You can also change the scope of the fixture for a quicker fix

presidio-image-redactor/tests/test_dicom_image_pii_verify_engine.py
SharonHart: Its fine either way, again just a preference. Just want to share my thought process:
I personally like to follow the Clean Code principals in which:
1. Comment, almost of all kinds, are considered bad, as they are not maintained and tend to not be reviewed, so with time they become more misleading than good. The alternatives offer more by count and more descriptive variables and method names. Testing a happy path and negative / exception tests can be incorporated in the test name.
2. Treat Test Code as Production Code - that along with the first principle applies comments to tests as well.

SharonHart (dismissed):
Left some follow up comments but overall looks great!

presidio-image-redactor/tests/test_dicom_image_pii_verify_engine.py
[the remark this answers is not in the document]
niwilso: I see your point, thank you for explaining! However, I still feel that having these tests are really useful.

I have been using a test-driven development approach for writing these private methods. Having the unit tests for private methods was critical in saving me time when I was updating the logic of existing code to use lists instead of dictionaries in this PR.

To your point about developers needing to alter private methods without causing breaking changes, I feel the developers should have a good enough understanding of any methods they are altering to be able to update the associated unit tests as necessary.

Also, although these methods do have the underscore prefix, they are not truly private methods.

presidio-image-redactor/presidio_image_redactor/dicom_image_pii_verify_engine.py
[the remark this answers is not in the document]
niwilso: Thank you for clarifying!

1. Yes, it makes sense then to pull this out into a utility accessible by different modules. Do you have a recommended location?
2. This should be doable for the non-DICOM case as well, but I would need to move `._get_bboxes_from_ocr_results()` and `_get_bboxes_from_analyzer_results()` as well into the new utility location.

presidio-image-redactor/presidio_image_redactor/dicom_image_pii_verify_engine.py
```diff
@@ -0,0 +1,310 @@
...
+    def _calculate_precision(gt: List[dict], all_pos: List[dict]) -> float:
+        """Calculate precision.
+
+        :param gt: List of ground truth labels.
+        :param all_pos: All Detected PHI (mapped back to have actual PHI text).
+        :return: Precision value.
+        """
+        # Find True Positive (TP) and precision
+        tp = [i for i in all_pos if i in gt]
+        precision = len(tp) / len(all_pos)
```
omri374, line 294: Can `len(all_pos)` be 0? if yes, consider adding an epsilon

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
```diff
@@ -567,27 +567,28 @@ def _make_phi_list(
...
         :return: Bounding box info organized.
         """
-        bboxes_dict = {}
-        for i in range(0, len(analyzer_results)):
+        bboxes = []
+        for i in range(len(analyzer_results)):
             result = analyzer_results[i].to_dict()
 
-            bboxes_dict[str(i)] = {
+            bbox_item = {
```
omri374, line 581: isn't `result` already a dict? why do we need to copy it to a new one?

presidio-image-redactor/tests/integration/test_dicom_image_pii_verify_engine_integration.py
omri374: IMHO this is fine. While we cannot ensure that it would be immutable, this is definitely the design we have for the different engines in Presidio (all of them are stateless). 

If a test requires some change to the default, we can have them instantiated within the test instead of using the fixture.

presidio-image-redactor/tests/integration/test_dicom_image_pii_verify_engine_integration.py
```diff
@@ -0,0 +1,130 @@
...
+    filepath = f"{SCRIPT_DIR}/resources/0_ORIGINAL.dcm"
+
+    # Act
+    instance = pydicom.dcmread(filepath)
+
+    return instance
+
+
+@pytest.fixture(scope="module")
+def mock_results():
```
omri374, line 45: If you use the same fixtures in multiple tests/files, consider creating them in the `conftest.py` file with a broader scope to avoid code duplication and to simplify the flow

presidio-image-redactor/tests/test_dicom_image_pii_verify_engine.py
omri374: I think that moving the mock functions to `conftest.py` would simplify the structure of this file and make those comments unnecessary.

presidio-image-redactor/tests/test_dicom_image_pii_verify_engine.py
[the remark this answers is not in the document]
omri374: I don't have a strong preference here :) @niwilso if we could think of test cases to test private methods through public methods, and make sure they work as expected, that would be the best, but it's not always straightforward. Another option is to consider making those public. Maybe a user would want access to precision/recall methods?

omri374 (commented):
Thank you @niwilso! Added some comments, overall, as always, looks great.

presidio-image-redactor/presidio_image_redactor/dicom_image_pii_verify_engine.py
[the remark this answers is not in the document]
niwilso: The left, top, width, and height attributes are available in the returned list.

I have decided to use a dictionary approach to reduce the lines of code for now. I will revisit the use of dunders in the future PR for moving the bounding box code to a common module.

presidio-image-redactor/presidio_image_redactor/dicom_image_pii_verify_engine.py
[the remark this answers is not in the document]
niwilso: Will address moving this code to a common module in a separate PR.

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
niwilso: I create a new one such that we only keep what is necessary for the downstream bounding box operations, and so that we don't alter the existing results in case the user wants to explore `analyzer_results` on their own separately.

presidio-image-redactor/tests/test_dicom_image_pii_verify_engine.py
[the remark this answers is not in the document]
niwilso: Ah good point. I'll make the precision and recall calculation methods public since there are cases where the user may want to call them.

I'll revisit testing private methods through public methods in the future PR for refactoring the bounding box code into a common module.

presidio-image-redactor/tests/integration/test_dicom_image_pii_verify_engine_integration.py
niwilso: Since this fixture is not used across different files, I will keep this in here. However, I will keep this in mind for future tests.

niwilso: I do see that `mock_instance` is used across files so I will move that to `conftest.py`.

SharonHart (dismissed):

presidio-image-redactor/tests/conftest.py
```diff
@@ -45,3 +50,26 @@ def get_image_recognizerresult():
 @pytest.fixture(scope="module")
 def image_analyzer_engine():
     return ImageAnalyzerEngine()
+
+
+@pytest.fixture(scope="module")
+def mock_instance():
```
omri374, line 56: nit: `mock_instance` is a very generic name, and this package does different things. Consider changing to `get_mock_dicom_instance` or something similar. We can also address this in a future PR.

presidio-image-redactor/tests/conftest.py
```diff
@@ -45,3 +50,26 @@ def get_image_recognizerresult():
...
+    filepath = f"{SCRIPT_DIR}/test_data/0_ORIGINAL.dcm"
+
+    # Act
+    instance = pydicom.dcmread(filepath)
+
+    return instance
+
+
+@pytest.fixture(scope="module")
+def mock_results():
```
omri374, line 68: Same comment as above

presidio-image-redactor/presidio_image_redactor/dicom_image_pii_verify_engine.py
```diff
@@ -0,0 +1,316 @@
...
+        :param gt: List of ground truth labels.
+        :param all_pos: All Detected PHI (mapped back to have actual PHI text).
+        :return: Recall value.
+        """
+        # Find True Positive (TP) and precision
+        tp = [i for i in all_pos if i in gt]
+        try:
+            recall = len(tp) / len(gt)
+        except ZeroDivisionError:
+            recall = 1
```
omri374, line 314: I would go with 0 as default :) but it's hard to say how this would be used by others.

omri374 (dismissed):
I left a couple of low priority comments. Thank you! Great work.

omri374 (approved):

SharonHart (approved):

Issue #998, opened 2023-01-04
Enable OCR thresholding for image redaction


niwilso:
**Is your feature request related to a problem? Please describe.**
When redacting using the `ImageRedactorEngine` or the `DicomImageRedactorEngine`, analysis is run on all detected text, regardless of if we are confident that the text actually exists or not. This can lead to undesired False Positive redacting boxes being drawn onto images.

**Describe the solution you'd like**
Update the `ImageAnalyzerEngine.analyze()` method such that the user can pass in an `ocr_threshold`. This `ocr_threshold` can then be used in a new staticmethod to filter out the `ocr_result` before `text` is defined.

**Describe alternatives you've considered**
I believe adding an optional argument with a default value into `.analyze()` will work best since the `**kwargs` are already specific to the `self.AnalyzerEngine.analyze()` call.

niwilso:
I am working on a False Positive investigation and have developed code for this as a result. Will create a Pull Request for this once the currently open PRs related to DICOM image redaction are merged.

PR #1001, opened 2023-01-04
Enable thresholding of OCR results

niwilso:
## Change Description

This PR adds the ability to threshold the OCR results (using the returned confidence values) before they are passed to the analyzer to help reduce False Positives. This is particularly useful in cases where we run redaction on a large number of images, regardless of if the images have text present on them or not.

### Summary of Changes
* `ImageAnalyzerEngine` has a new static method to filter OCR results based on the returned confidence scores
* Users can now pass a dictionary of OCR parameters (e.g., `ocr_kwargs={"ocr_threshold": 50}`) to `ImageAnalyzerEngine` via `ImageRedactorEngine`, `ImagePiiVerifyEngine`, `DicomImageRedactorEngine`, and `DicomImagePiiVerifyEngine` 
* Minor documentation updates

## Issue reference

This PR fixes issue #998 

## Checklist


presidio-image-redactor/tests/test_image_analyzer_engine.py
```diff
@@ -113,3 +113,38 @@ def test_given_dif_len_entities_then_map_analyzer_returns_correct_outputand_len(
...
+        "Cromley.",
+    ]
+    ocr_result["left"] = [143, 143, 322, 530, 634, 827, 896, 141, 141]
+    ocr_result["top"] = [64, 67, 67, 76, 64, 64, 64, 134, 134]
+    ocr_result["width"] = [936, 160, 191, 87, 172, 51, 183, 801, 190]
+    ocr_result["height"] = [50, 47, 37, 28, 40, 50, 40, 50, 50]
+    ocr_result["conf"] = [-1, 99.5, 92.3, 42.7, 66.1, 51.2, 79.7, 64.0, 70.3]
+
+    # Act
+    test_filtered = image_analyzer_engine.threshold_ocr_result(
```
omri374, line 145: Consider adding an integration test as well. As discussed, I'm not sure this code would not work (haven't tested it though) because the `ocr_threshold` parameter would throw an error in `AnalyzerEngine.analyze`).

presidio-image-redactor/presidio_image_redactor/image_analyzer_engine.py
```diff
@@ -23,7 +27,7 @@ def __init__(self, analyzer_engine: AnalyzerEngine = None, ocr: OCR = None):
             ocr = TesseractOCR()
         self.ocr = ocr
 
-    def analyze(self, image: object, **kwargs) -> List[ImageRecognizerResult]:
+    def analyze(self, image: Optional[object], **kwargs) -> List[ImageRecognizerResult]:
```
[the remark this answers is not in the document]
niwilso: Thank you for writing out the potential alternatives.

The current approach is Approach 2 you described above, where `ocr_threshold` is removed from `**kwargs` prior to being passed into `AnalyzerEngine.analyze` (see lines 41-45).

If we want to go back to explicitly defining `ocr_threshold`, then I can revert this PR to [c35ffb1e86659c5bc9cd44fab70fc85c1ba34b3d](https://github.com/microsoft/presidio/tree/c35ffb1e86659c5bc9cd44fab70fc85c1ba34b3d)

presidio-image-redactor/tests/test_image_analyzer_engine.py
niwilso: It does work given the removal of `ocr_threshold` from `**kwargs` as defined in `ImageAnalyzerEngine.analyze` (the integration tests run successfully and I have tested separately as well), but let's see the outcome of the earlier comment before adjusting this.

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
```diff
@@ -127,9 +129,9 @@ def redact_from_directory(
         self,
         input_dicom_path: str,
         output_dir: str,
-        padding_width: int = 25,
-        crop_ratio: float = 0.75,
-        fill: str = "contrast",
+        padding_width: Optional[int] = 25,
```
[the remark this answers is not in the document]
niwilso: No, this was a misunderstanding on my end. I'll remove `Optional[]` for any of the arguments that cannot have a valid `None` type.

Thank you for the resource!

presidio-image-redactor/presidio_image_redactor/image_analyzer_engine.py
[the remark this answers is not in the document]
niwilso: The downside of going back to the previous commit is that more files are changed, but if we feel that brings clarity, then it could be worth it.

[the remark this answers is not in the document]
niwilso: As discussed prior, the only place where `ocr_threshold` is really applied is directly in `ImageAnalyzerEngine.analyze`, which is called by `ImageAnalyzerEngine`, `ImagePiiVerifyEngine`, `DicomImageAnalyzerEngine`, and `DicomPiiVerifyEngine`. This is why this current version of the PR uses the `**kwargs` approach to reduce repeatition.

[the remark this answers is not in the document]
niwilso: And apologies I missed the last line of your original comment. I'll schedule some time for us to discuss. I agree it may be easier to chat in real-time about the pros and cons of the different approaches.

presidio-image-redactor/tests/test_image_analyzer_engine.py
omri374: Great, I missed that part. Thanks!

presidio-image-redactor/presidio_image_redactor/image_analyzer_engine.py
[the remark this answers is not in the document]
niwilso: @SharonHart - FYI, we have decided to go forward with an approach where OCR parameters are specified in the Optional argument `ocr_kwargs`, keeping it separate from the general `**kwargs` used by the text AnalyzerEngine.

presidio-image-redactor/presidio_image_redactor/image_analyzer_engine.py
```diff
@@ -23,15 +27,26 @@ def __init__(self, analyzer_engine: AnalyzerEngine = None, ocr: OCR = None):
             ocr = TesseractOCR()
         self.ocr = ocr
 
-    def analyze(self, image: object, **kwargs) -> List[ImageRecognizerResult]:
+    def analyze(
+        self, image: object, ocr_kwargs: Optional[dict] = None, **kwargs
```
omri374, line 31: consider renaming `kwargs` to `text_kwargs`. This would be backward compatible as we call it with **.

omri374 (dismissed):
Thanks! great work

presidio-image-redactor/presidio_image_redactor/image_analyzer_engine.py
niwilso: Agreed, that does make things clearer. I've added another commit renaming `kwargs` to `text_analyzer_kwargs` in `ImageAnalyzerEngine` and all the classes that pass kwargs to it.

omri374 (approved):

presidio-image-redactor/presidio_image_redactor/image_analyzer_engine.py
```diff
@@ -23,25 +27,62 @@ def __init__(self, analyzer_engine: AnalyzerEngine = None, ocr: OCR = None):
...
-        :return: list of the extract entities with image bounding boxes
+        :return: List of the extract entities with image bounding boxes.
         """
-        ocr_result = self.ocr.perform_ocr(image)
+        # Perform OCR
+        perform_ocr_kwargs, ocr_threshold = self._parse_ocr_kwargs(ocr_kwargs)
+        ocr_result = self.ocr.perform_ocr(image, **perform_ocr_kwargs)
+
+        # Apply OCR confidence threshold if it is passed in
+        if ocr_threshold is not None:
```
SharonHart, line 47: if ocr_threshold:

presidio-image-redactor/presidio_image_redactor/image_analyzer_engine.py
```diff
@@ -23,25 +27,62 @@ def __init__(self, analyzer_engine: AnalyzerEngine = None, ocr: OCR = None):
...
+        for i, val in enumerate(ocr_result["conf"]):
+            if float(val) >= ocr_threshold:
+                idx.append(i)
+
+        # Only retain high confidence items
+        filtered_ocr_result = {}
+        for key in list(ocr_result.keys()):
+            filtered_ocr_result[key] = [ocr_result[key][i] for i in idx]
+
+        return filtered_ocr_result
```
SharonHart, line 84: nit: Because of the not-so-straightforward structure we get the results in, would be much easier to query the confidence while mapping into bounding boxes (`map_analyzer_results_to_bounding_boxes`). When iterating ocr results - keep processing into BBs only if above threshold.

SharonHart (dismissed):
Two minor nit suggestions, looks good!

presidio-image-redactor/presidio_image_redactor/image_analyzer_engine.py
niwilso: That is a good point, however, `map_analyzer_results_to_bounding_boxes` takes in `text` as an argument. I placed the ocr thresholding inside `analyze` before `get_text_from_ocr_dict` is called such that we only have to filter out in one location rather than two (e.g., `ocr_results` and `text`).

SharonHart: got it, makes sense

SharonHart (approved):

omri374 (approved):

PR #1009, opened 2023-01-17
Image Redactor - REST API to support web applications payload

SharonHart:
## Change Description

Support sending image as encoded byte array, to align with `input` based web application components.

## Checklist

### DIFF
--- a/e2e-tests/common/methods.py
+++ b/e2e-tests/common/methods.py
@@ -1,3 +1,5 @@
+import base64
+import json
 import os
 
 import requests
@@ -45,12 +47,16 @@ def analyzer_supported_entities(data):
     return response.status_code, response.content
 
 
-def redact(file, color_fill=None):
-    multipart_form_data = __get_multipart_form_data(file)
+def redact(file, color_fill=None, json_payload=False):
     payload = __get_redact_payload(color_fill)
-    response = requests.post(
-        f"{IMAGE_REDACTOR_BASE_URL}/redact", files=multipart_form_data, data=payload
-    )
+    if json_payload:
+        json_string = '{"image": "' + base64.b64encode(file.read()).decode("utf-8") + '"}'
+        response = requests.post(
+            f"{IMAGE_REDACTOR_BASE_URL}/redact", json=json.loads(json_string), headers=DEFAULT_HEADERS
+        )
+    else:
+        response = requests.post(
+        f"{IMAGE_REDACTOR_BASE_URL}/redact", files=__get_multipart_form_data(file), data=payload)
     return response
 
 
--- a/presidio-image-redactor/app.py
+++ b/presidio-image-redactor/app.py
@@ -52,16 +52,20 @@ class Server:
             if request.get_json(silent=True) and "image" in request.json:
                 im = Image.open(BytesIO(base64.b64decode(request.json.get("image"))))
                 analyzer_entities = request.json.get('analyzer_entities')
-                redacted_image = self.engine.redact(im, color_fill, entities=analyzer_entities)
+                redacted_image = self.engine.redact(im, color_fill,
+                                                    entities=analyzer_entities)
+                img_byte_arr = image_to_byte_array(redacted_image, im.format)
+                return Response(base64.b64encode(img_byte_arr),
+                                mimetype="application/octet-stream")
+
             elif request.files and "image" in request.files:
                 im = Image.open(request.files.get("image"))
                 redacted_image = self.engine.redact(im, color_fill)
+                img_byte_arr = image_to_byte_array(redacted_image, im.format)
+                return Response(img_byte_arr,
+                                mimetype="application/octet-stream")
             else:
                 raise InvalidParamException("Invalid parameter, please add image data")
-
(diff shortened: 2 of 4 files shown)


presidio-image-redactor/tests/test_dicom_image_redactor_engine.py
```diff
@@ -90,7 +90,7 @@ def test_get_all_dcm_files_happy_path(
     print(expected_list)
 
     # Assert
-    assert test_files == expected_list
```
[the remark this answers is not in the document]
niwilso: Thank you for the heads up! I'll look into this.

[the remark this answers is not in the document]
SharonHart: I think that the change here solved it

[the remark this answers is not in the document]
niwilso: Thank you for fixing this! I'll include this in my current PR ( #1001 ) to help fix the build pipeline issue I'm currently facing as well.

omri374 (approved):
Looks great. Thanks!

e2e-tests/common/methods.py
```diff
@@ -45,12 +47,16 @@ def analyzer_supported_entities(data):
     return response.status_code, response.content
 
 
-def redact(file, color_fill=None):
-    multipart_form_data = __get_multipart_form_data(file)
+def redact(file, color_fill=None, json_payload=False):
```
niwilso, line 50: nit: I suggest adding type hinting and docstrings.

e2e-tests/common/methods.py
```diff
@@ -45,12 +47,16 @@ def analyzer_supported_entities(data):
...
     payload = __get_redact_payload(color_fill)
-    response = requests.post(
-        f"{IMAGE_REDACTOR_BASE_URL}/redact", files=multipart_form_data, data=payload
-    )
+    if json_payload:
+        json_string = '{"image": "' + base64.b64encode(file.read()).decode("utf-8") + '"}'
+        response = requests.post(
+            f"{IMAGE_REDACTOR_BASE_URL}/redact", json=json.loads(json_string), headers=DEFAULT_HEADERS
+        )
+    else:
```
niwilso, line 57: nit: Might be worth adding a check / raise exception here (and exception test) if the user passes in the image in the json but forgets to set `json_payload=True`.

presidio-image-redactor/app.py
```diff
@@ -47,15 +49,23 @@ def redact():
...
-            image_file = request.files.get("image")
-            if not image_file:
-                raise InvalidParamException("Invalid parameter, please add image data")
-            im = Image.open(image_file)
-
-            redacted_image = self.engine.redact(im, color_fill)
+            if request.get_json(silent=True) and "image" in request.json:
+                im = Image.open(BytesIO(base64.b64decode(request.json.get("image"))))
+                analyzer_entities = request.json.get('analyzer_entities')
+                redacted_image = self.engine.redact(im, color_fill,
```
niwilso, line 55: Please update to allow user to pass in the optional `ocr_kwargs` argument and optional `**text_analyzer_kwargs` (see latest version of [presidio-image-redactor/presidio_image_redactor/image_redactor_engine.py](https://github.com/microsoft/presidio/blob/main/presidio-image-redactor/presidio_image_redactor/image_redactor_engine.py)

presidio-image-redactor/app.py
```diff
@@ -47,15 +49,23 @@ def redact():
...
+                                                    entities=analyzer_entities)
+                img_byte_arr = image_to_byte_array(redacted_image, im.format)
+                return Response(base64.b64encode(img_byte_arr),
+                                mimetype="application/octet-stream")
 
-            img_byte_arr = image_to_byte_array(redacted_image, im.format)
-            return Response(img_byte_arr, mimetype="application/octet-stream")
+            elif request.files and "image" in request.files:
+                im = Image.open(request.files.get("image"))
+                redacted_image = self.engine.redact(im, color_fill)
```
niwilso, line 63: Same comment as above re: `ocr_kwargs` and `**text_analyzer_kwargs`.

niwilso (commented):
Overall looks great! Left some comments for things to consider and one required change.

e2e-tests/common/methods.py
SharonHart: That's exclusive for the e2e framework, and to use the same method

presidio-image-redactor/app.py
SharonHart: each arg needs to be parsed separately by the API, like the entities, so I don't think we can do it globally

presidio-image-redactor/app.py
SharonHart: See below

e2e-tests/common/methods.py
SharonHart: As this the just the test framework, I think we previously decided to not use both (aligned with the other tests)

balteravishay (approved):

Issue #996, opened 2023-01-03
Move bounding box comparison logic to common module


niwilso:
**Is your feature request related to a problem? Please describe.**
Currently, the logic for comparing ground truth to returned bounding boxes (i.e., label and coordinates) is contained in the `DicomImagePiiVerifyEngine` and its associated code. The bounding box logic could be broadly useful for evaluating the performance of any Presidio module.

**Describe the solution you'd like**
We should move the existing logic for bounding box operations to a common module such that it is easily accessible by all Presidio modules and is not tied specifically to the `DicomImagePiiVerifyEngine`.

**Describe alternatives you've considered**
We *can* keep the logic tied to `DicomImagePiiVerifyEngine` and import from there, but it would not be intuitive.

**Additional context**
We may also want to add functionality to check for if one bounding box is contained within another. If a box is contained within another, keep the larger box.

PR #1011, opened 2023-01-18
Consolidating general bounding box operations

niwilso:
## Change Description

This PR moves existing bounding box code spread across different classes in the presidio-image-redactor module into one class. This new `BboxProcessor` class can also be used independently during evaluation, as specified in the updated documentation.

### Summary of Changes
* Existing, general bounding box operations have been consolidated into `BboxProcessor`
* Existing tests for general bounding box operations have been moved into `presidio-image-redactor/tests/test_bbox.py` (no new tests have been added)
* `ImageRedactorEngine`, `DicomImageRedactorEngine`, and `DicomImagePiiVerifyEngine` and their tests have been updated to call bounding box operations from `BboxProcessor`
* Updated evaluation documentation to call bounding box operations from inherited `BboxProcessor` methods

## Issue reference

This PR fixes issue #996 

## Checklist

### DIFF
--- a/docs/image-redactor/evaluating_dicom_redaction.md
+++ b/docs/image-redactor/evaluating_dicom_redaction.md
@@ -73,11 +73,10 @@ The `DicomImagePiiVerifyEngine` class can be used to assist in ground truth labe
 
 ```python
 import pydicom
-from presidio_image_redactor import DicomImagePiiVerifyEngine, BboxProcessor
+from presidio_image_redactor import DicomImagePiiVerifyEngine
 
-# Initialize engines
+# Initialize engine
 dicom_engine = DicomImagePiiVerifyEngine()
-bbox_processor = BboxProcessor()
 
 # Choose your file to create ground truth for
 filename = "path/to/your/file.dcm"
@@ -88,8 +87,8 @@ padding_width = 25
 verification_image, ocr_results, analyzer_results = dicom_engine.verify_dicom_instance(instance, padding_width)
 
 # Format results for more direct comparison
-ocr_results_formatted = bbox_processor.get_bboxes_from_ocr_results(ocr_results)
-analyzer_results_formatted = bbox_processor.get_bboxes_from_analyzer_results(analyzer_results)
+ocr_results_formatted = dicom_engine.bbox_processor.get_bboxes_from_ocr_results(ocr_results)
+analyzer_results_formatted = dicom_engine.bbox_processor.get_bboxes_from_analyzer_results(analyzer_results)
 ```
 
 By looking at the output of `verify_dicom_instance`, we can create a ground truth labels json.
--- a/docs/image-redactor/evaluating_dicom_redaction.md
+++ b/docs/image-redactor/evaluating_dicom_redaction.md
@@ -73,7 +73,11 @@ The `DicomImagePiiVerifyEngine` class can be used to assist in ground truth labe
 
 ```python
 import pydicom
-from presidio_image_redactor import DicomImagePiiVerifyEngine
+from presidio_image_redactor import DicomImagePiiVerifyEngine, BboxProcessor
+
+# Initialize engines
+dicom_engine = DicomImagePiiVerifyEngine()
+bbox_processor = BboxProcessor()
 
 # Choose your file to create ground truth for
 filename = "path/to/your/file.dcm"
@@ -84,8 +88,8 @@ padding_width = 25
 verification_image, ocr_results, analyzer_results = dicom_engine.verify_dicom_instance(instance, padding_width)
 
 # Format results for more direct comparison
-ocr_results_formatted = dicom_engine._get_bboxes_from_ocr_results(ocr_results)
-analyzer_results_formatted = dicom_engine._get_bboxes_from_analyzer_results(analyzer_results)
+ocr_results_formatted = bbox_processor.get_bboxes_from_ocr_results(ocr_results)
+analyzer_results_formatted = bbox_processor.get_bboxes_from_analyzer_results(analyzer_results)
 ```
 
 By looking at the output of `verify_dicom_instance`, we can create a ground truth labels json.
--- a/presidio-image-redactor/presidio_image_redactor/bbox.py
+++ b/presidio-image-redactor/presidio_image_redactor/bbox.py
@@ -51,23 +51,20 @@ class BboxProcessor:
 
         return bboxes
 
-    def format_bboxes_for_dicom(
-        self, analyzer_results: list, padding_width: int
+    def remove_bbox_padding(
+        self, analyzer_bboxes: List[dict], padding_width: int
     ) -> List[dict]:
-        """Format the bounding boxes to write directly back to DICOM pixel data.
+        """Remove added padding in bounding box coordinates.
 
-        :param analyzer_results: The analyzer results.
+        :param analyzer_bboxes: The bounding boxes from analyzer results.
         :param padding_width: Pixel width used for padding (0 if no padding).
 
         :return: Bounding box information per word.
         """
         if padding_width < 0:
-            raise ValueError("Padding width must be a positive number.")
+            raise ValueError("Padding width must be a non-negative integer.")
 
-        # Write bounding box info to json files for now
-        bboxes = self.get_bboxes_from_analyzer_results(analyzer_results)
-
-        # remove padding from all bounding boxes
+        # Remove padding from all bounding boxes
         bboxes = [
             {
                 "top": max(0, bbox["top"] - padding_width),
@@ -75,7 +72,7 @@ class BboxProcessor:
                 "width": bbox["width"],
                 "height": bbox["height"],
             }
-            for bbox in bboxes
+            for bbox in analyzer_bboxes
         ]
 
         return bboxes
--- a/presidio-image-redactor/presidio_image_redactor/bbox.py
+++ b/presidio-image-redactor/presidio_image_redactor/bbox.py
@@ -15,7 +15,6 @@ class BboxProcessor:
         :return: Bounding box information per word.
         """
         bboxes = []
-        item_count = 0
         for i in range(len(ocr_results["text"])):
             detected_text = ocr_results["text"][i]
             if detected_text:
@@ -28,7 +27,6 @@ class BboxProcessor:
                     "label": detected_text,
                 }
                 bboxes.append(bbox)
-                item_count += 1
 
         return bboxes
 
--- a/presidio-image-redactor/presidio_image_redactor/bbox.py
+++ b/presidio-image-redactor/presidio_image_redactor/bbox.py
@@ -1,10 +1,14 @@
-from typing import List, Tuple
+from presidio_image_redactor.entities import ImageRecognizerResult
+from typing import List, Tuple, Dict, Union
 
 
 class BboxProcessor:
     """Common module for general bounding box operators."""
 
-    def get_bboxes_from_ocr_results(self, ocr_results: dict) -> List[dict]:
+    def get_bboxes_from_ocr_results(
+        self,
+        ocr_results: Dict[str, List[Union[int, str]]],
+    ) -> List[Dict[str, Union[int, float, str]]]:
         """Get bounding boxes on padded image for all detected words from ocr_results.
 
         :param ocr_results: Raw results from OCR.
@@ -28,7 +32,10 @@ class BboxProcessor:
 
         return bboxes
 
-    def get_bboxes_from_analyzer_results(self, analyzer_results: list) -> List[dict]:
+    def get_bboxes_from_analyzer_results(
+        self,
+        analyzer_results: List[ImageRecognizerResult],
+    ) -> List[Dict[str, Union[str, float, int]]]:
         """Organize bounding box info from analyzer results.
 
         :param analyzer_results: Results from using ImageAnalyzerEngine.
@@ -52,8 +59,10 @@ class BboxProcessor:
         return bboxes
 
     def remove_bbox_padding(
-        self, analyzer_bboxes: List[dict], padding_width: int
-    ) -> List[dict]:
+        self,
+        analyzer_bboxes: List[Dict[str, Union[str, float, int]]],
+        padding_width: int,
+    ) -> List[Dict[str, int]]:
         """Remove added padding in bounding box coordinates.
 
         :param analyzer_bboxes: The bounding boxes from analyzer results.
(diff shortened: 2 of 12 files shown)


niwilso:
All tests pass locally, but there seems to be an import error when running the build pipeline. Pipeline failure can be seen [here](https://dev.azure.com/csedevil/Presidio/_build/results?buildId=14973&view=logs&j=334821e9-0aec-524d-aadd-ec8ab81296e4&t=7e3d8d17-f121-508d-a19c-0b4e9ae8c2b0).

@omri374 , @SharonHart - Do either of you know why `presidio-image-redactor/__init__.py` would fail to load `bbox.BboxProcessor`?

SharonHart:
> All tests pass locally, but there seems to be an import error when running the build pipeline. Pipeline failure can be seen [here](https://dev.azure.com/csedevil/Presidio/_build/results?buildId=14973&view=logs&j=334821e9-0aec-524d-aadd-ec8ab81296e4&t=7e3d8d17-f121-508d-a19c-0b4e9ae8c2b0).
> 
> @omri374 , @SharonHart - Do either of you know why `presidio-image-redactor/presidio_image_redactor/__init__.py` would fail to load `bbox.BboxProcessor`?

I don't see an BboxProcessor file, missed committing it?

presidio-image-redactor/presidio_image_redactor/bbox.py
```diff
@@ -0,0 +1,130 @@
+from typing import List, Tuple
+
+
+class BboxProcessor:
+    """Common module for general bounding box operators."""
+
+    @staticmethod
+    def get_bboxes_from_ocr_results(ocr_results: dict) -> List[dict]:
```
omri374, line 8: There is an assumption here that the user uses tesseract, as `ocr_results` are the tesseract output. If we turn `get_bboxes_from_ocr_results` into a non static method, a user could potentially implement a new `BboxProcessor` and overwrite this method with logic to extract boxes for another OCR engine. WDYT?

presidio-image-redactor/presidio_image_redactor/bbox.py
```diff
@@ -0,0 +1,130 @@
...
+                    "conf": float(ocr_results["conf"][i]),
+                    "label": detected_text,
+                }
+                bboxes.append(bbox)
+                item_count += 1
+
+        return bboxes
+
+    @staticmethod
+    def get_bboxes_from_analyzer_results(analyzer_results: list) -> List[dict]:
```
omri374, line 33: Consider adding explicit input types. Not sure if it is 
```python
analyzer_results: List[List[RecognizerResult]]
```
or
```python
List[List[ImageRecognizerResult]]
```

presidio-image-redactor/presidio_image_redactor/bbox.py
```diff
@@ -0,0 +1,130 @@
...
+                "top": result["top"],
+                "width": result["width"],
+                "height": result["height"],
+            }
+            bboxes.append(bbox_item)
+
+        return bboxes
+
+    @classmethod
+    def format_bboxes_for_dicom(
```
omri374, line 57: Is this only for dicom? if we call it `remove_padding_from_results` or something similar, then we could use it also for non-dicom cases, right?

presidio-image-redactor/presidio_image_redactor/bbox.py
```diff
@@ -0,0 +1,130 @@
...
+                "height": bbox["height"],
+            }
+            for bbox in bboxes
+        ]
+
+        return bboxes
+
+    @staticmethod
+    def match_with_source(
+        all_pos: List[dict],
```
omri374, line 88: I would suggest writing more explicit type hints here and for `pii_source_dict`.

docs/image-redactor/evaluating_dicom_redaction.md
```diff
@@ -73,7 +73,11 @@ The `DicomImagePiiVerifyEngine` class can be used to assist in ground truth labe
 
 ```python
 import pydicom
-from presidio_image_redactor import DicomImagePiiVerifyEngine
+from presidio_image_redactor import DicomImagePiiVerifyEngine, BboxProcessor
+
+# Initialize engines
+dicom_engine = DicomImagePiiVerifyEngine()
+bbox_processor = BboxProcessor()
```
omri374, line 80: Does the user need to interact directly with this object? If it is a field in `dicom_engine` then we could call its methods through the dicom_engine, e.g. `dicom_engine.bbox_processor.get_bboxes_from_ocr_results` or expose a method e.g. `dicom_engine.get_bboxes_from_ocer_results` that would call the `bbox_processor`'s method. This could simplify the API for the user.

omri374:
Thanks @niwilso! This is a great improvement for the codebase. 

As this package evolves and matures, I think we should consider architecting it in a way that would support further customization and analysis and wanted to get your and @SharonHart's thoughts. My initial suggestion is to have a base class for image analysis and redaction, and an inheriting class for dicom image analysis and redaction.
Each class would initially have three methods:
- redact
- validate
- analyze

The redact and validate would be the same as the `ImageRedactorEngine` and `ImageVerifyEngine`. The `analyze` method would return all bounding boxes with their text analysis results. This would facilitate further debugging.

This is definitely not for this PR, and might not be doable or easy, so please provide your feedback and we can also have a quick call to further discuss this if needed.

niwilso:
> Thanks @niwilso! This is a great improvement for the codebase.
> 
> As this package evolves and matures, I think we should consider architecting it in a way that would support further customization and analysis and wanted to get your and @SharonHart's thoughts. My initial suggestion is to have a base class for image analysis and redaction, and an inheriting class for dicom image analysis and redaction. Each class would initially have three methods:
> 
> * redact
> * validate
> * analyze
> 
> The redact and validate would be the same as the `ImageRedactorEngine` and `ImageVerifyEngine`. The `analyze` method would return all bounding boxes with their text analysis results. This would facilitate further debugging, and can also plug into improvements in the demo website that Sharon is currently working on (in future releases)
> 
> This is definitely not for this PR, and might not be doable or easy, so please provide your feedback and we can also have a quick call to further discuss this if needed.

Yes, I think it makes sense to consolidate as you suggested. I think it would also be great to have "evaluate" as one of the required public methods (already exists for the DICOM engines but would need to be added for the standard image engine).

My only concerns would be around backwards compatibility and having really large classes.

I'm happy to discuss this over a call as well. I've gone ahead and made an issue for this to handle in the future: https://github.com/microsoft/presidio/issues/1012

presidio-image-redactor/presidio_image_redactor/bbox.py
```diff
@@ -0,0 +1,130 @@
...
+                "left": result["left"],
+                "top": result["top"],
+                "width": result["width"],
+                "height": result["height"],
+            }
+            bboxes.append(bbox_item)
+
+        return bboxes
+
+    @classmethod
```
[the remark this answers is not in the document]
niwilso: I have this as a `@classmethod` because it calls another method within the class, `cls.get_bboxes_from_analyzer_results`.

presidio-image-redactor/presidio_image_redactor/bbox.py
niwilso: Good point, I'll remove the `@staticmethod` decorator here.

Though, I do want to point out that the expected format of `ocr_results` here matches what is produced within `ImageAnalyzerEngine.analyze`. Perhaps we should add some sort of schema check inside the `ocr` class that ensures the output of `ocr.perform_ocr` matches the output format from Tesseract?

docs/image-redactor/evaluating_dicom_redaction.md
niwilso: Good point. There isn't really a benefit from using the `BboxProcessor` class directly from a user standpoint. In the future, we may add code within the broader image (verification) class to assist with evaluation.

I'll change this to calling `dicom_engine.bbox_processor.get_bboxes_from_ocr_results` for now and then will update with any future changes to evaluation assistance code.

presidio-image-redactor/presidio_image_redactor/bbox.py
niwilso: Created an issue to address the OCR output schema checking in the future: https://github.com/microsoft/presidio/issues/1013

omri374: Thanks, that's a great suggestion.

omri374 (approved):

presidio-image-redactor/presidio_image_redactor/bbox.py
```diff
@@ -0,0 +1,135 @@
...
+        self,
+        ocr_results: Dict[str, List[Union[int, str]]],
+    ) -> List[Dict[str, Union[int, float, str]]]:
+        """Get bounding boxes on padded image for all detected words from ocr_results.
+
+        :param ocr_results: Raw results from OCR.
+        :return: Bounding box information per word.
+        """
+        bboxes = []
+        item_count = 0
```
SharonHart, line 18: not used?

presidio-image-redactor/presidio_image_redactor/bbox.py
```diff
@@ -0,0 +1,135 @@
...
+        for i in range(len(ocr_results["text"])):
+            detected_text = ocr_results["text"][i]
+            if detected_text:
+                bbox = {
+                    "left": ocr_results["left"][i],
+                    "top": ocr_results["top"][i],
+                    "width": ocr_results["width"][i],
+                    "height": ocr_results["height"][i],
+                    "conf": float(ocr_results["conf"][i]),
+                    "label": detected_text,
```
SharonHart, line 28: I would consider consolidating those attributes in a Bbox class

presidio-image-redactor/presidio_image_redactor/bbox.py
```diff
@@ -0,0 +1,135 @@
...
+        for i in range(len(analyzer_results)):
+            result = analyzer_results[i].to_dict()
+
+            bbox_item = {
+                "entity_type": result["entity_type"],
+                "score": result["score"],
+                "left": result["left"],
+                "top": result["top"],
+                "width": result["width"],
+                "height": result["height"],
```
SharonHart, line 55: Along with the suggested Bbox class change, this can be a one liner with spread operator

SharonHart (dismissed):
Looks good

SharonHart:
> Thanks @niwilso! This is a great improvement for the codebase.
> 
> As this package evolves and matures, I think we should consider architecting it in a way that would support further customization and analysis and wanted to get your and @SharonHart's thoughts. My initial suggestion is to have a base class for image analysis and redaction, and an inheriting class for dicom image analysis and redaction. Each class would initially have three methods:
> 
> * redact
> * validate
> * analyze
> 
> The redact and validate would be the same as the `ImageRedactorEngine` and `ImageVerifyEngine`. The `analyze` method would return all bounding boxes with their text analysis results. This would facilitate further debugging, and can also plug into improvements in the demo website that Sharon is currently working on (in future releases)
> 
> This is definitely not for this PR, and might not be doable or easy, so please provide your feedback and we can also have a quick call to further discuss this if needed.

I like this approach

presidio-image-redactor/presidio_image_redactor/bbox.py
niwilso: Besides reducing some lines of code (lines 23-28 and 50-55), is there a functional reason to create an additional class? I feel that this would be extra work that would also need extra tests.

SharonHart: the spread operator / ocr_result to bbox implemented once

niwilso: Could you provide a code example? I don't think I'm visualizing the correct implementation in terms of how the bbox class would be structured and implemented.

SharonHart: Assuming Bbox is defined as dict or pydantic or dataclass:
```python
class Bbox(dict):
    left: int
    top: int
    width: int
    height: int
    text: str
    conf: float
    label: str
    entity_type: str

```

We can transform `get_bboxes_from_analyzer_results` to a one-liner:
```python
    def get_bboxes_from_analyzer_results(
        self,
        analyzer_results: List[ImageRecognizerResult],
    ) -> List[Dict[str, Union[str, float, int]]]:
        """Organize bounding box info from analyzer results.

        :param analyzer_results: Results from using ImageAnalyzerEngine.

        :return: Bounding box info organized.
        """
        return [Bbox(**result.to_dict()) for result in analyzer_results]

```

And Bbox an be accessed with a more straightforward access `bbox.left `instead of `bbox["left"]`

niwilso: Got it! Thank you for explicitly writing this out. This helps a lot. I'll make the changes.

niwilso: @SharonHart - I've spent this morning making the changes in my local branch and have discovered that introducing the `Bbox` class introduces more complications than I believe are necessary.

## Complications
1. The `get_bboxes_from_analyzer_results` method cannot be reduced down to a single line because this method is also used to filter out irrelevant properties and only carry over relevant information into the bounding boxes. The method would need to look like the following if we introduce the `Bbox` class.

```python
    def get_bboxes_from_analyzer_results(
        self,
        analyzer_results: List[ImageRecognizerResult],
    ) -> List[Bbox]:
        """Organize bounding box info from analyzer results.

        :param analyzer_results: Results from using ImageAnalyzerEngine.

        :return: Bounding box info organized.
        """
        bboxes = []
        exclude = ["start", "end", "analysis_explanation", "recognition_metadata"]
        for result in analyzer_results:
            values = result.to_dict()

            # Remove irrelevant fields
            for key in exclude:
                values.pop(key, None)

            # Add bbox with relevant fields
            bboxes.append(Bbox(**values))

        return bboxes
```

2. Users provide the ground truth as a dictionary either defined in their notebook or loaded from a json file during evaluation. Loading ground truth from JSON files is convenient and works easily with dictionaries. If we use the `Bbox` class for OCR results and analyzer results, but then need to compare those against the ground truth (like in `BboxProcessor.match_with_source`), there is a mix of types. Keeping everything as dictionaries is easier and more straightforward in my opinion. Forcing users to format their ground truth to fit the `Bbox` class would be an unnecessary extra step and prevent easy use with JSON files.
4. There would be no methods or any operations needed for the `Bbox` class itself, which I feel supports the case to not make a class for it in the first place.

## Suggestion
Given the above, I suggest we keep the code as-is with the latest commit. If that is alright with you, please re-approve and I will resolve the related comments.

SharonHart (approved):

presidio-image-redactor/presidio_image_redactor/bbox.py
niwilso: Leaving as-is and marking as resolved (see above response from me about the `Bbox` class).

omri374 (approved):

PR #1023, opened 2023-01-27
Adding a sample for identifying PII in a PDF

andrewcmsft:
## Change Description

Adds a new python notebook sample showing how to identify and annotate PII in a PDF. Makes use of two PDF packages which both have permissive licenses.

I've updated the sample index, but if there anywhere else I need to update, please let me know in the PR comments.

## Issue reference

No specific issue, this was in response to a discussion with Omri.

## Checklist

- [ ] My code includes unit tests (N/A - sample only.)
- [ ] All unit tests and lint checks pass locally (N/A - sample only.)


docs/samples/index.md
```diff
@@ -16,6 +16,7 @@
 | Usage       | Python                                | [Using Transformers as an external PII model](python/transformers_recognizer/index.md)                                               |
 | Usage       | Python Notebook                       | [Anonymizing known values](python/Anonymizing%20known%20values.ipynb)
 | Usage       | Python Notebook                       | [Redacting text PII from DICOM images](python/example_dicom_image_redactor.ipynb)
+| Usage       | Python Notebook                       | [Annotating PII in a PDF](python/example_pdf_annotation.ipynb)
```
omri374, line 19: Please add to the Python TOC as well: https://github.com/microsoft/presidio/blob/main/docs/samples/python/index.md

docs/samples/python/example_pdf_annotation.ipynb
```diff
@@ -0,0 +1,254 @@
...
+    "This sample takes a PDF as an input, extracts the text, identifies PII using Presidio and annotates the PII using highlight annotations.\n",
+    "\n",
+    "## Prerequisites\n",
+    "Before getting started, make sure the following packages are installed. For detailed documentation, see the [installation docs](https://microsoft.github.io/presidio/installation).\n",
+    "\n",
+    "Install from PyPI:\n",
+    "```bash\n",
+    "pip install presidio_analyzer\n",
+    "pip install presidio_anonymizer\n",
+    "pip install presidio_image_redactor\n",
```
omri374, line 19: Is the Presidio Image Redactor used here?

omri374 (dismissed):
Thank you @andrewcmsft, this is a great sample! very easy to go through and well documented.

docs/samples/python/example_pdf_annotation.ipynb
```diff
@@ -0,0 +1,254 @@
...
+    "We finally iterate through all the analyzed bounding boxes and create highlight annotations for all of them."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "pdf = Pdf.open(\"./sample_data/sample.pdf\")\n",
```
omri374, line 156: Is it possible to read the pdf once and use the same object throughout the sample notebook?

docs/samples/python/example_pdf_annotation.ipynb
andrewcmsft: Good spot! It's a copy and paste error from another sample.

docs/samples/python/example_pdf_annotation.ipynb
andrewcmsft: So we choose to use two PDF libraries here, the first is pdfminer, which we use solely for text extraction, we do so as it's sole function is to extract data (text) from a PDF accurately. In the case of this sample, the text extraction needs to be accurate so we've got the best chance of finding PII. For example, some libraries 'break' the text when it's extracted, leading to extra or missing spaces in text. Unfortunately text extraction from a PDF is not a trivial thing to get right, but pdfminer seems to do the best job of it.

The second library we use is pikepdf. We choose this one for updating and writing out as pdfminder doesn't have writing capabilities, and of the other PDF libraries, this one is meant to produce good, valid PDF. It also supports standards such as PDF/A, which is used quite a lot in government and organisations where archiving is important.

So long story short, we open the file twice, once for each library. :)

docs/samples/index.md
andrewcmsft: All sorted, thanks!

docs/samples/python/example_pdf_annotation.ipynb
omri374: I see, thanks!

omri374 (approved):

SharonHart (approved):

Issue #1035, opened 2023-02-17
Adding QR codes support in the ImageRedactorEngine


vpvpvpvp:
**Is your feature request related to a problem? Please describe.**
QR codes usually contain PII entities. But ocr engines like tesseract can't recognize QR codes on images.

**Describe the solution you'd like**
Integrate into Presidio Image Redactor the ability to recognize QR codes and analyze their contents. And if the QR code contains PII, then redact it. One way to implement this functionality is to add a new `QRImageAnalyzerEngine` in addition to the default `ImageAnalyzerEngine`. And then passing `QRImageAnalyzerEngine` as the `image_analyzer_engine` parameter to `ImageRedactorEngine` and to `ImagePiiVerifyEngine`.

**Additional context**
An example of the input image (first image) and the desired output (second image):

![qr_multi-min](https://user-images.githubusercontent.com/55928886/219716569-6f13256e-023b-4ceb-8946-bd3bbd98beaf.png)

![qr_multi_redacted-min](https://user-images.githubusercontent.com/55928886/219716510-b3afed99-f24a-4106-87fe-9b307965a94d.png)

omri374:
Thanks! we'll review the PR shortly.

PR #1036, opened 2023-02-17
Adding QR codes support in the ImageRedactorEngine

vpvpvpvp:
## Change Description

This PR adds to the Presidio Image Redactor the ability to analyze the content of QR codes on the image.

qr-image-redactor-design

## Summary of Changes

* Added abstract class `QRRecognizer` for QR code recognizers
* Added concrete `OpenCVQRRecongnizer` which uses OpenCV to recognize QR codes
* Added `QRImageAnalyzerEngine` which uses `QRRecognizer` for QR code recognition and `AnalyzerEngine` to analyze its contents for PII entities
* Modified `ImagePiiVerifyEngine` and `ImageRedactorEngine` to allow using `QRImageAnalyzerEngine` as an alternative to `ImageAnalyzerEngine`

## Issue reference

This PR fixes issue #1035

## Checklist


vpvpvpvp:

SharonHart:
@vpvpvpvp
Seems like the unit tests are failing in the CI, what version of Tesseract OCR did you use to generate the baseline images for the tests?

vpvpvpvp:
> @vpvpvpvp Seems like the unit tests are failing in the CI, what version of Tesseract OCR did you use to generate the baseline images for the tests?

Tesseract OCR - 5.2.0
pytesseract - 0.3.10
OS - MacOS Ventura 13.2

Indeed, I noticed that in different test environments, the results of ImagePiiVerifyEngine may differ in some pixels. For example, below in the first image is the result on Mac, next on Ubuntu and their difference. At the same time, both the recognized text itself and the box coordinates are the same. 
mac_ubuntu_diff

omri374:
Hi @vpvpvpvp, before going deeper into the code, what are your thoughts of having the QR code analyzer working potentially in parallel to OCR? something like that:
```mermaid
stateDiagram-v2
    read_image
    read_image --> extract_ocr_text
    read_image --> extract_qr_text
    extract_ocr_text --> presidio_analyzer
    extract_qr_text --> presidio_analyzer
    presidio_analyzer --> redact_image
    redact_image --> return_image
```

Then we could always extend it to more types of detectors in the future, similar to the text analyzer architecture, e.g.:
```mermaid
stateDiagram-v2
    read_image
    read_image --> extract_ocr_text
    read_image --> extract_qr_text
    read_image --> extract_faces
    read_image --> extract_license_plates
    extract_ocr_text --> presidio_analyzer
    extract_qr_text --> presidio_analyzer
    presidio_analyzer --> redact_image
    extract_faces --> redact_image
    extract_license_plates --> redact_image
    redact_image --> return_image
```

One way to achieve this is to have `QRImageAnalyzerEngine` extend `ImageAnalyzerEngine`, and then we could later create a composable `ImageAnalyzerEngine` which holds multiple image analyzers.
WDYT?

vpvpvpvp:
Hi @omri374, that sounds great! In the current PR you can choose between `QRImageAnalyzerEngine` and `ImageAnalyzerEngine`, but it would be great to be able to run them and other analyzers in parallel. At first, I wanted to extend `ImageAnalyzerEngine` a bit, so that it would also accept `QRRecognizer` as a parameter in addition to OCR. Something like this:

```
class ImageAnalyzerEngine:
    """ImageAnalyzerEngine class.

    :param analyzer_engine: The Presidio AnalyzerEngine instance
        to be used to detect PII in text
    :param ocr: the OCR object to be used to detect text in images.
    :param qr: the QRRecognizer object to detect and decode text in QR codes
    """
    def __init__(
        self,
        analyzer_engine: Optional[AnalyzerEngine] = None,
        ocr: Optional[OCR] = None,
        qr: Optional[QRRecognizer] = None,
    ):
```

And then, in the `analyze` function, extract the text and its coordinates first with `self.ocr` and then with `self.qr`. But then I decided not to change `ImageAnalyzerEngine` this time, to make less edits to the original source code.

vpvpvpvp:
> > > @vpvpvpvp Seems like the unit tests are failing in the CI, what version of Tesseract OCR did you use to generate the baseline images for the tests?
> > 
> > 
> > Tesseract OCR - 5.2.0 pytesseract - 0.3.10 OS - MacOS Ventura 13.2
> > Indeed, I noticed that in different test environments, the results of ImagePiiVerifyEngine may differ in some pixels. For example, below in the first image is the result on Mac, next on Ubuntu and their difference. At the same time, both the recognized text itself and the box coordinates are the same. mac_ubuntu_diff
> 
> I would suggest to use the original image as baseline (not a screenshot of it or of the screen). If its still failing, lets see how to add thresholding to the comparison

Updated the test images, locally the tests passed.

SharonHart:
@vpvpvpvp Seems to work now, but failing on another test that should be resolved in https://github.com/microsoft/presidio/pull/1032 , try to rebase after once merged

presidio-image-redactor/Pipfile
```diff
@@ -12,6 +12,8 @@ pydicom = ">=2.3.0"
 pypng = ">=0.20220715.0"
 matplotlib = "==3.6.2"
 typing-extensions = "*"
+opencv-python = ">=4.5.0"
```
omri374, line 15: opencv-python needs to be added to `setup.py` too

vpvpvpvp: Updated `setup.py` and `Pipefile.lock`

omri374: Thanks!

presidio-image-redactor/setup.py
```diff
@@ -12,6 +12,8 @@
     "pydantic==1.7.4",
     "pydicom>=2.3.0",
     "pypng>=0.20220715.0",
+    "opencv-python = >=4.5.0",
```
omri374, line 15:
```suggestion
    "opencv-python>=4.5.0",
```

SharonHart:
@vpvpvpvp you have a green build :)
Will try to review the code later today

presidio-image-redactor/tests/integration/test_qr_image_analyzer_engine_integration.py
```diff
@@ -0,0 +1,39 @@
+from presidio_image_redactor.entities import ImageRecognizerResult
+from tests.integration.methods import get_resource_image
+
+
+def test_given_qr_image_then_text_entities_are_recognized_correctly(
```
SharonHart, line 5: very nice scenarios

presidio-image-redactor/presidio_image_redactor/qr_recognizer.py
```diff
@@ -0,0 +1,144 @@
...
+
+        if ret:
+            decoded = self._decode(image, points)
+
+            for text, p in zip(decoded, points):
+                (x, y, w, h) = cv2.boundingRect(p)
+
+                recognized.append(
+                    QRRecognizerResult(
+                        text=text, bbox=[x, y, w, h], polygon=[*p.flatten(), *p[0]]
```
SharonHart, line 106: passed list but declared tuple

presidio-image-redactor/presidio_image_redactor/qr_recognizer.py
```diff
@@ -0,0 +1,144 @@
...
+                    )
+                )
```
SharonHart, line 108: Prefer immutability.
```suggestion
            for text, p in zip(decoded, points):
                (x, y, w, h) = cv2.boundingRect(p)

                recognized.append(
                    QRRecognizerResult(
                        text=text, bbox=[x, y, w, h], polygon=[*p.flatten(), *p[0]]
                    )
                )
```
recognized = [QRRecognizerResult(text=text, bbox=cv2.boundingRect(point), polygon=[*point.flatten(), *point[0]]) for text, point in zip(decoded, points)]

presidio-image-redactor/presidio_image_redactor/qr_image_analyzer_engine.py
```diff
@@ -0,0 +1,62 @@
+from typing import List, Optional
+
+from presidio_analyzer import AnalyzerEngine
+
+from presidio_image_redactor.entities import ImageRecognizerResult
+from presidio_image_redactor.qr_recognizer import QRRecognizer
+from presidio_image_redactor.qr_recognizer import OpenCVQRRecongnizer
+
+
+class QRImageAnalyzerEngine:
```
omri374, line 10: Can this class be inherited from `ImageAnalyzerEngine`? Just a question, to see if we can simplify the design instead of extending it to a new set of independent classes.

presidio-image-redactor/presidio_image_redactor/image_redactor_engine.py
```diff
@@ -11,7 +15,10 @@ class ImageRedactorEngine:
     :param image_analyzer_engine: Engine which performs OCR + PII detection.
     """
 
-    def __init__(self, image_analyzer_engine: ImageAnalyzerEngine = None):
+    def __init__(
+        self,
+        image_analyzer_engine: Union[ImageAnalyzerEngine, QRImageAnalyzerEngine] = None,
```
omri374, line 20: If `QRImageAnalyzerEngine` inherits from `ImageAnalyzerEngine`, then this class could be independent of the QR implementation

omri374 (commented):
Thanks again for the contribution! Left some points for discussion, hopefully we can simplify the design and decouple the QR code analysis from downstream classes.

presidio-image-redactor/presidio_image_redactor/image_redactor_engine.py
```diff
@@ -42,9 +49,12 @@ def redact(
 
         image = ImageChops.duplicate(image)
 
-        bboxes = self.image_analyzer_engine.analyze(
-            image, ocr_kwargs, **text_analyzer_kwargs
-        )
+        if isinstance(self.image_analyzer_engine, QRImageAnalyzerEngine):
```
SharonHart, line 52: @omri374 @vpvpvpvp Any idea on making it more open-close? 
Maybe a single ImageAnalyzerEngine that we inherit from with optional **ocr_kwargs?

SharonHart (approved):
Looks great, left some minor comments

presidio-image-redactor/presidio_image_redactor/qr_image_analyzer_engine.py
SharonHart: Was thinking exactly the same, see below.

presidio-image-redactor/tests/integration/test_qr_image_analyzer_engine_integration.py
vpvpvpvp: Should I add more complex/realistic scenarios?

presidio-image-redactor/presidio_image_redactor/qr_recognizer.py
vpvpvpvp: I will add these changes, thanks for the suggestion.

presidio-image-redactor/presidio_image_redactor/qr_recognizer.py
```diff
@@ -0,0 +1,144 @@
+from abc import ABC, abstractmethod
+from typing import Tuple, List, Optional
+import cv2
+import numpy as np
+
+
+class QRRecognizerResult:
```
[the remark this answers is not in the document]
vpvpvpvp: I'm not so sure about that. `QRRecognizerResult` is needed to represent the results of QR code recognition (bboxes and raw text without PII analysis). In this sense, `QRRecognizerResult` is closer to the dictionary returned by the `perform_ocr` method of the `TesseractOCR(OCR)` class. At the same time, `ImageRecognizerResult` already includes the results of text analysis by the presidio_analyzer.

presidio-image-redactor/presidio_image_redactor/qr_image_analyzer_engine.py
vpvpvpvp: Yes, it can be inherited from `ImageAnalyzerEngine`. My concern is that in this case, `QRImageAnalyzerEngine` will also inherit the logic of working with ocr tools not related to QR code recognition.

presidio-image-redactor/presidio_image_redactor/image_redactor_engine.py
vpvpvpvp: In case of direct inheritance of `QRImageAnalyzerEngine` from `ImageAnalyzerEngine`, it would only need to add ocr_kwars to the analyze method of `QRImageAnalyzerEngine`. This is probably the easiest way.

Potentially, it seems like the most optimal implementation when `ImageAnalyzerEngine` is used for orchestrating different recognizers (ocr recognizer, QR recognizer, etc.). In the vein of what was suggested earlier https://github.com/microsoft/presidio/pull/1036#issuecomment-1437483048.

presidio-image-redactor/presidio_image_redactor/qr_recognizer.py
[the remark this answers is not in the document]
omri374: I see, thanks for the clarification

presidio-image-redactor/presidio_image_redactor/qr_image_analyzer_engine.py
omri374: Yes, that's my concern too. As the package is still in beta, we should (carefully) consider breaking backward compatibility. We'll do some thinking on this and get back to you. We can also have a quick design session together over video if you're interested.

vpvpvpvp: Yes, that sounds interesting. If you have time, we could do that.

omri374: Sure. To avoid putting personal emails on GH, could you please email presidio@microsoft.com and we'll continue the discussion over email?

pabloromeo (approved):

PR #1062, opened 2023-04-26
Added "keep" anonymizer

paulo-raca:
## Change Description

This adds `keep` anonymizer. This is a no-op and doesn't modify the input string.

Having it allows us to keep track of all PII data in the anonymizer output, even when we want to preserve some types of PII.

### Example
I'm making a demo that highlights all PIIs and anonymizes only phone numbers.

When using `anonymizers: {"DEFAULT":{"type":"keep"}, "PHONE_NUMBER": {"type":"replace"}}`

> I'm John Smith and my phone is +1234567890

Becomes

> I'm `John Smith` and my phone is `<PHONE_NUMBER>`

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
+### Added
+#### Anonymizer
+* Added `keep`, an no-op anonymizer that allows preserving some types of PII while keeping track of its position in anonymized output.
 
 ## [2.2.32] - 25.01.2023
 ### Changed
--- a/docs/anonymizer/index.md
+++ b/docs/anonymizer/index.md
@@ -190,7 +190,7 @@ with some other value by applying a certain operator (e.g. replace, mask, redact
 
 ## Built-in operators
 
-Operator type | Operator name | Description | Parameters |
+| Operator type | Operator name | Description | Parameters |
 | --- | --- | --- | --- |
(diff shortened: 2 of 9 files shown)


omri374:
Thanks for this great contribution! I've thing I'm curious about is what happens when we have overlaps. Currently the package resolves overlaps using a set of rules. It is possible that the output would not be the same as the input in case of overlaps. Have you looked into this by any chance? Consider adding an integration test for this if it is really an issue.

paulo-raca:
> One thing I'm curious about is what happens when we have overlaps. 

I haven't looked into it yet, but I think having a no-op anonymizer instead of ignoring the entity type would help keeping th behavior consistent.

> Consider adding an integration test for this if it is really an issue

I'll look into this.

paulo-raca:
@omri374, I've added e2e tests and the summary is that `I'm George Washington Square Park` becomes `I'm George WashingtonWashington Square Park`.

It is a bit weird, but it is consistent with the Partial intersection rules described in the docs, so I'm good with it :+1: 

I also installed pre-commit and fixed the linting errors

omri374 (dismissed):
LGTM, Thanks!

paulo-raca:

dorlugasigal:
this looks great @paulo-raca
can you modify the tests name to have a `given_when_then` or `given_then` test names
and an additional sample notebook under `docs/sample/python`?

paulo-raca:
>can you modify the tests name to have a `given_when_then` or `given_then` test names

Whoops, looks a copy-pasting mistake :sweat_smile: 

> and an additional sample notebook under `docs/sample/python`?

Added, let me know if it is enough :+1:

omri374 (approved):

dorlugasigal (approved):
lgtm

PR #1080, opened 2023-06-01
Lock with 3.8, drop building on 3.7 for 3.11

SharonHart:
* Lock pipenv with python 3.8 as 3.7 is reaching EOL, and installing from a locked 3.7 file is no longer working.
* Run build on 3.11 instead of 3.7

(at the time of this PR, installing from pypi still works on >=3.7.2)


SharonHart:
@omri374 please rename yourself to 384

PR #1106, opened 2023-07-04
Anonymize with instances

### LINKED ISSUE #1102: Instance preserving anonymization

When I use a REPLACE anonymizer, all entities of the same `entity_type` are collapsed to the same `<ENTITY_TYPE>` token. For example "Alice and Bob and Alice" is replaced to "<NAME> and <NAME> and <NAME>" and it is not possible to determine whether these tokens refer to the same or different people.

The ability to preserve instance identity. For example, if we have a `preserve_instance: True` param to the `REPLACE` operator, then we would like the result:

"<NAME 0> and <NAME 1> and <NAME 0>"

It may be useful to annotate instance identity within the `PIIEntity` or `RecognizerResult` classes themselves, for example for later use in deanonymization.

**Additional context**

We have this running on a fork; happy to submit a PR if there's interest.

feynmanliang:
## Change Description

Adds instance identifiers to `replace` anonymizer entities.

NOTE: we overrode JSON serialization to exclude the instance identifiers so that existing tests could pass with minimal changes; let me know if we'd rather update the assertions themselves (all serialized JSONs for `PIIEntity` in tests will need to be changed to include this new `instance_id` field).

## Issue reference

Fixes #1102

## Checklist

### DIFF
--- a/presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
+++ b/presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
@@ -80,6 +80,15 @@ class AnonymizerEngine(EngineBase):
 
         operators = self.__check_or_add_default_operator(operators)
 
+        instances = []
+        for result in analyzer_results:
+            instance = text[result.start:result.end]
+            if instance in instances:
+                result.instance_id = instances.index(instance)
+            else:
+                instances.append(instance)
+                result.instance_id = len(instances) - 1
+
         return self._operate(text, analyzer_results, operators, OperatorType.Anonymize)
 
     def _remove_conflicts_and_get_text_manipulation_data(
--- a/presidio-anonymizer/presidio_anonymizer/core/engine_base.py
+++ b/presidio-anonymizer/presidio_anonymizer/core/engine_base.py
@@ -65,7 +65,8 @@ class EngineBase(ABC):
                 operator.entity_type,
                 changed_text,
                 operator_metadata.operator_name,
-                operator.instance_id
+                operator.instance_id,
+                text_to_operate_on
             )
             engine_result.add_item(result_item)
 
--- a/presidio-anonymizer/presidio_anonymizer/core/engine_base.py
+++ b/presidio-anonymizer/presidio_anonymizer/core/engine_base.py
@@ -65,6 +65,7 @@ class EngineBase(ABC):
                 operator.entity_type,
                 changed_text,
                 operator_metadata.operator_name,
+                operator.instance_id
             )
             engine_result.add_item(result_item)
 
@@ -88,6 +89,7 @@ class EngineBase(ABC):
         operator.validate(params=operator_metadata.params)
         params = operator_metadata.params
         params["entity_type"] = entity_type
+        params["instance_id"] = text_metadata.instance_id
         self.logger.debug(f"operating on {entity_type} with {operator}")
         operated_on_text = operator.operate(params=params, text=text_to_operate_on)
         return operated_on_text
--- a/presidio-anonymizer/presidio_anonymizer/entities/engine/pii_entity.py
+++ b/presidio-anonymizer/presidio_anonymizer/entities/engine/pii_entity.py
@@ -14,10 +14,11 @@ class PIIEntity(ABC):
 
     logger = logging.getLogger("presidio-anonymizer")
 
-    def __init__(self, start: int, end: int, entity_type: str):
+    def __init__(self, start: int, end: int, entity_type: str, instance_id: int = None):
         self.start = start
         self.end = end
         self.entity_type = entity_type
+        self.instance_id = instance_id
         self.__validate_fields()
 
     def __repr__(self):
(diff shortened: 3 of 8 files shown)


omri374:
Hi @feynmmanliang, I like this implementation a lot, thanks for creating this! My only comment is focused on allowing more arbitrary pseudonoymization in the future, e.g., instead of `David -> PERSON_1`, we could have `David -> Mark` or `David -> D`.
While allowing such capability isn't necessary for this PR, can we think of a way to do this in the future without overcomplicating the API or changin it in the future?

feynmanliang:
Thanks for the comments; I saw your remark about this in the issue and have been thinking about it as well. My first thought was to maintain an entity type -> pseudonym generator map like https://github.com/microsoft/presidio/blob/a1c8a23e206f1a5671f164758a4d14b769901fd4/presidio-analyzer/tests/test_context_support.py#L25-L36, but since my requirements here are to maintain instance identity there's going to have to be some state like `instances` in this PR around.

An alternative would be to offer a format-preserving encryption option (eg https://support.fortanix.com/hc/en-us/articles/15658066081044-Security-Objects-Tokenization-Quickstart), which does not require a token mapping table (ie "vaultless tokenization"). This would allow more arbitrary pseudonymization (via specifying the FPE cipher alphabet) but adds much more API surface area.

SharonHart:
What if we think about it as a separate new anonymization operator? That way it will be easier to keep the substitution generic, like Omri suggested.
The downside of this is that you no longer have the id as part of the API, **BUT** the {ENTITY}_{INDEX} is unique enough and can be used as id to find similar values. (PERSON_0 can't be with the same value as LOCATION_0)
We can also make the index generation as a function, potentially generating uuid in the future for example, 
and then replace the list with a dictionary for keeping track of the indices, and placing it in the operator class?

feynmanliang:
We need the token mappings as part of the API to do detokenization; is there another way we can return the plaintext to entity+id mappings to a client?

SharonHart:
Maybe I missed, how is it returned in the current implementation? 
by accessing engine.instances?

feynmanliang:
https://github.com/microsoft/presidio/pull/1106/files#diff-51935f832d6830c31b45e997277afbd0214b10ba6bb9400f83fbc5ff154a50cbL23

omri374:
What if we do this like the encryption/decryption operators work? We can have a new operator (say called `Pseudonymize`) which would receive a PII entity and replace it with a value (new if it's not seen yet or existing if the value exists in the mapping). Then, we can have a `Depseudonymize` operator, similar to `Deanonymize` which would take a pseudonymized text, and revert the process. 

Another thing to consider is the persistence of the mapping. In the current implementation, if we run the process twice, we would get different mappings.

omri374:
What if we do this like the encryption/decryption operators work? We can have a new operator (say called `Pseudonymize`) which would receive a PII entity and replace it with a value (new if it's not seen yet or existing if the value exists in the mapping). Then, we can have a `Depseudonymize` operator, similar to `Deanonymize` which would take a pseudonymized text, and revert the process. 

Another thing to consider is the persistency of the mapping.

feynmanliang:
I believe the current deanonymize API does not fit our use case because (1)
we are not detokenizing the same text sample as what's being tokenized and
(2) we are doing this in a streaming use case where BPE tokens of words are
arriving in chunks and buffering the full response before detokenizing has
unacceptable latency.

feynmanliang:
Happy to hop on a VC and discuss the use case in more detail if it is
easier, but AFAIK right now making the plaintext to replacement text
mapping available would be the easiest way to limit the scope here.

omri374:
Hi @feynmanliang, could you please send an email to presidio@microsoft.com, and we'll take it from there? We'd be more than happy to continue the discussion on this and find an approach that would fit all cases.

stnslwk:
I noticed that there has been some discussion on this issue, but it seems to have gone quiet recently. Are there any updates on this?

feynmanliang:
@stnslwk Apologies for the silence. Yes, I plan to get back to this issue after BlackHat/Defcon. Did you have some requirements or comments to share?

maks-operlejn-ds:
Hi @feynmanliang @omri374, do you plan to return to this topic soon? I am currently working on a module for LangChain that anonymizes (or rather pseudonymizes) data - a key thing when playing with LLMs through the API. I am actually using Microsoft Presidio, which seemed like a great tool for this (kudos!), but I lack the separation into instances.

@feynmanliang, if you don't have time for this feature, I can carry on this matter further. I think it would be better to do this in the Presidio repository rather than implement it in LangChain, as this would require some weird workarounds :smile: 

If you are interested, I am attaching some links to the langchain documentation with this feature:
https://python.langchain.com/docs/guides/privacy/presidio_data_anonymization/
https://python.langchain.com/docs/guides/privacy/presidio_data_anonymization/reversible
https://github.com/langchain-ai/langchain/blob/master/libs/experimental/langchain_experimental/data_anonymizer/presidio.py

feynmanliang:
Yeah please go ahead and take over the issue :)

omri374:
Hi @maks-operlejn-ds, great work on the Langchain plugin! We'd be happy to get your feedback on the discussion. Instance anonymization in a simple (non-parallel) flow, can be achieved by replacing values with instance ids, similar to how you replace values with faker values. This dictionary can be passed to the operator to re-use entity values already seen, or append a new one if missing. The reason we still don't have it in Presidio, is because it doesn't scale to parallel processing flows, as maintaining this mapping is challenging and platform dependent.

PR #1110, opened 2023-07-05
DICOM redactor improvement: Reorder kwargs

niwilso:
## Change Description

This PR introduces small kwargs reordering to enable an upcoming PR. The upcoming PR will extend the allow list capabilities to the DICOM redactor engine.

## Issue reference

This PR fixes issue #XX

## Checklist


ayabel (approved):

presidio-image-redactor/presidio_image_redactor/dicom_image_pii_verify_engine.py
```diff
@@ -91,8 +91,8 @@ def verify_dicom_instance(
         ocr_results = self.ocr_engine.perform_ocr(image)
         analyzer_results = self.image_analyzer_engine.analyze(
             image,
-            ad_hoc_recognizers=[deny_list_recognizer],
             ocr_kwargs=ocr_kwargs,
+            ad_hoc_recognizers=[deny_list_recognizer],
```
omri374, line 95: Changing the order of parameters is a breaking change. Is there an alternative to have this order changed?

niwilso: Unfortunately, this is required to enable the upcoming allow list approach with the DICOM image redactor. This is due to how additional parameters may be passed into both the image analyzer and the OCR engine.

When we do the Presidio image redactor module refactor, we will want to streamline this.

omri374: Thanks for the additional context

PR #1111, opened 2023-07-05
DICOM redactor improvement: Enable return of redacted bboxes

niwilso:
## Change Description

This PR introduces the ability to return the bounding boxes used in redaction. This may be helpful in cases where users want to evaluate or modify the redacted regions.

## Issue reference

This PR fixes issue #XX

## Checklist


ayabel (approved):

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
```diff
@@ -30,9 +31,10 @@ def redact(
         fill: str = "contrast",
         padding_width: int = 25,
         crop_ratio: float = 0.75,
+        return_bboxes: bool = False,
         ocr_kwargs: Optional[dict] = None,
         **text_analyzer_kwargs,
-    ):
+    ) -> Union[pydicom.dataset.FileDataset, Tuple[pydicom.dataset.FileDataset, dict]]:
```
omri374, line 37: Have you considered the option of having two methods, one for redaction and one which would return bounding boxes? Is there a use case to get both? If yes, could we have a third method which would return both?

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
```diff
@@ -741,6 +769,7 @@ def _redact_single_dicom_image(
         padding_width: int,
         overwrite: bool,
         dst_parent_dir: str,
+        return_bboxes: bool,
```
omri374, line 772: This is called `return` but it saves it to file. Can we consider renaming it to save?

omri374 (commented):
Cool! I left a comment on potentially breaking it into two methods. I would be happy to get your insights into this as I might not have all the context.

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
niwilso: Technically, users can return bbox information by using the verification engine rather than the redactor engine. However, there is currently no way to return the bbox information while running the redactor.

The reason why it is an optional return in the `redact()` method rather than a separate method is as follows:
- Users have expressed there is high value in returning the bbox information each time they run the redactor.
- With the above information, I believe it would be best to return bbox information by default alongside the redacted instance/image, but I did not want to change the number of variables being returned which may cause issues. This is something we can update/address in the broader Presidio image redactor module refactor.
- Running a separate method to return the bboxes would duplicate computation, unless we saved the bboxes into the redactor object and had users pull the information from there.

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
niwilso: That is a good point. The reason why I used "return_bboxes" here is to have consistency between the argument names in `.redact()`, `.redact_from_file()`, and `.redact_from_directory()`.

Do you suggest we prioritize consistency in the argument names (since these methods are highly similar), or should we adjust argument names as necessary in this group of methods?

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
omri374: Thanks. If we create a separate method that does exactly what this method does- returns both bboxes and redacted image, and then have two methods calling it: one for redact only, the other for bboxes only, we would not have to change the original redact API, but would allow the extended customizability you are after. Would that work?

```python
def get_redact_and_bbox(...):
   ...
   return redacted, bbox
   
def redact(...):
   redacted, _ = self.get_redact_and_bbox(...)
   return redacted
def get_bboxes(...):
   _, bboxes = self.get_redact_and_bbox(...)
   return bboxes
```

I agree that bboxes could be retrieved by the verification engine, but if we can keep API changes minimal, that would probably be better, especially as we are planning on changing the API once again soon.
WDYT?

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
niwilso: No problem, I've gone ahead and changed it. Thank you for the suggestion!

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
niwilso: Thank you for providing a code example! I see what you mean.

My only remaining concern is the inconsistency in how users access the bboxes used in redaction. With this approach, users would call a method other than `redact()` to return the redacted image and the associated bounding boxes, but would instead use an argument to effectively to do the same with the `redact_from_file()` and `redact_from_directory()` methods. These methods were designed to be as similar as possible for ease of use.

I understand the concern about changing the output of the `redact()` method, but I do feel it is clearer to include `return_bboxes` as an argument and then, in the future, ensure that any class with `redact()` always returns both the image and bboxes.

The code introduced in this PR does not break older versions since the default value for `return_bboxes` is `False`, meaning that only the redacted image will be returned unless specified otherwise.

SharonHart: I think that I agree with Omri, especially with the future refactoring idea of using interfaces vs inheritance, in which the redact method will have a single method signature, which will make it harder to refactor in the future. 
If I recall, we did say that we would split the redaction from the discovery, just like in the text analyzer and anonymizer.

niwilso: Understood. I'll implement the suggested approach then. Thank you both for the input!

niwilso: @omri374 , @SharonHart - I have pushed a new commit which separates out `redact()` and `redact_and_return_bbox()`.

I did not create an isolated `get_bbox()` method because I do not see a use case where the user would not just use `_, bboxes = engine.redact_and_return_bbox()` when using the redactor engine. If users wanted just bounding boxes without a redacted image, they can use the verifier instead.

omri374: Thanks @niwilso. I think the return type should be updated too.

niwilso: Great catch, @omri374 ! I have removed the Union typing. It now just returns the tuple.

omri374 (approved):

ayabel (dismissed):

omri374 (approved):

omri374 (approved):

ayabel (approved):

PR #1113, opened 2023-07-06
DICOM redactor improvement: Enable selection of redact approach

niwilso:
## Change Description

In the current state, the DICOM image redactor automatically uses the metadata to create a custom recognizer to assist in analyzing words to redact. However, this may not always be the best approach.

This PR introduces changes which allow the user to toggle whether they want to use the DICOM metadata and also allows them to easily pass in `ad_hoc_recognizers` to use for analysis.

## Issue reference

This PR fixes issue #XX

## Checklist


presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
```diff
@@ -733,12 +737,74 @@ def _add_redact_box(
...
+        redact_approach: Union[str, PatternRecognizer],
+        ocr_kwargs: Optional[dict],
+        **text_analyzer_kwargs
+    ) -> List[ImageRecognizerResult]:
+        """Analyze image with selected redaction approach.
+
+        :param image: DICOM pixel data as PIL image.
+        :param instance: DICOM instance (with metadata).
+        :param redact_approach: What approach to use when redacting
+        ("default", "metadata", "allow", or a PatternRecognizer object).
```
omri374, line 753: Please see my comment on this approach further down. In case we decide to go with this approach, could you please provide more information on each redact approach in the docstring? (Here and in the main function's docstring).

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
```diff
@@ -733,12 +737,74 @@ def _add_redact_box(
...
+                )
+                analyzer_results = self.image_analyzer_engine.analyze(
+                    image,
+                    ocr_kwargs=ocr_kwargs,
+                    ad_hoc_recognizers=[deny_list_recognizer],
+                    **text_analyzer_kwargs,
+                )
+            else:
+                raise ValueError("Please enter valid string or PatternRecognizer object for redact_approach")  # noqa: E501
+        elif type(redact_approach) == PatternRecognizer:
```
omri374, line 788: What is the benefit of this approach vs. passing an AnalyzerEngine with added recognizers? It could be that a user would want to pass in multiple deny-lists, each for different entity types, or with different confidence scores. While this can be a useful shortcut, IMHO it makes the interface more complex, and the underlying logic would not be clear to a user who might be confused by the Analyzer object and the PatternRecognizer object.

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
```diff
@@ -43,6 +45,8 @@ def redact(
         :param padding_width: Padding width to use when running OCR.
         :param crop_ratio: Portion of image to consider when selecting
         most common pixel value as the background color value.
+        :param redact_approach: What approach to use when redacting
```
SharonHart, line 48: its a bit weird that the default is not 'default'. Maybe choose different terminology

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
```diff
@@ -733,12 +737,74 @@ def _add_redact_box(
...
+        ("default", "metadata", "allow", or a PatternRecognizer object).
+        :param ocr_kwargs: Additional params for OCR methods.
+        :param text_analyzer_kwargs: Additional values for the analyze method
+        in AnalyzerEngine (e.g., allow_list).
+
+        :return: Analyzer results.
+        """
+        # Detect PII
+        if type(redact_approach) == str:
```
SharonHart, line 761: I would consider using interfaces for a more opn-close design. next approach will just accumulate here and further lengthen the method (which is already complex)

SharonHart: Also what if I want to use multiple approaches like before - metadata and text analyzer?

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
niwilso: Thank you for bringing this up.

To make this as clear, consistent, and convenient as possible for the users, I would like your input on the following questions. Please see the full `_get_analyzer_results()` copied below for a better understanding of how I plan for the method to look after adding in more options (note: only a portion was included in this PR to keep the PR small).

1. Users can still pass in their own customized AnalyzerEngine when instantiating their DICOM image redactor engine. To prevent confusion, **should I remove the "elif" completely** and just have the "default" approach as the default? This would still use the user's custom AnalyzerEngine. I would make note of this in the documentation.
2. I believe there is value in having a few built-in options for approach since the "metadata" and "allow" approaches are common scenarios. Besides including more detail in the docstrings and updating the documentation (coming in a separate PR once we have more approaches implemented), is there anything else I need to consider?

```python
def _get_analyzer_results(
            self,
            image: PIL.PngImagePlugin.PngImageFile,
            instance: pydicom.dataset.FileDataset,
            redact_approach: Union[str, PatternRecognizer],
            ocr_kwargs: Optional[dict],
            **text_analyzer_kwargs
        ) -> List[ImageRecognizerResult]:
        """Analyze image with selected redaction approach.
        :param image: DICOM pixel data as PIL image.
        :param instance: DICOM instance (with metadata).
        :param redact_approach: What approach to use when redacting ("default", "metadata", "allow", or a PatternRecognizer object).
        :param ocr_kwargs: Additional params for OCR methods.
        :param text_analyzer_kwargs: Additional values for the analyze method
        in AnalyzerEngine (e.g., allow_list).
        :return: Analyzer results.
        """
        # Detect PII
        if type(redact_approach) == str:
            if redact_approach.lower() == "default":
                # Use default redactor
                analyzer_results = self.image_analyzer_engine.analyze(
                    image,
                    ocr_kwargs=ocr_kwargs,
                    **text_analyzer_kwargs,
                )
            elif redact_approach.lower() == "metadata":
                # Create custom recognizer using DICOM metadata
                original_metadata, is_name, is_patient = self._get_text_metadata(instance)
                phi_list = self._make_phi_list(original_metadata, is_name, is_patient)
                deny_list_recognizer = PatternRecognizer(
                    supported_entity="PERSON", deny_list=phi_list
                )
                analyzer_results = self.image_analyzer_engine.analyze(
                    image,
                    ocr_kwargs=ocr_kwargs,
                    ad_hoc_recognizers=[deny_list_recognizer],
                    **text_analyzer_kwargs,
                )
            elif redact_approach.lower() in ["allow", "allow_list", "allow list"]:
                # Mark all text as PII by default
                pattern_all_text = Pattern(name="any_text", regex=r"(?s).*", score=0.5)
                custom_recognizer = PatternRecognizer(
                    supported_entity="TEXT",
                    patterns=[pattern_all_text],
                )
                # Pass in allow_list through text_analyzer_kwargs
                analyzer_results = self.image_analyzer_engine.analyze(
                    image,
                    ocr_kwargs=ocr_kwargs,
                    ad_hoc_recognizers=[custom_recognizer],
                    **text_analyzer_kwargs,
                )
            else:
                raise ValueError("Please enter valid string or PatternRecognizer object for redact_approach")
        elif type(redact_approach)==PatternRecognizer:
            # Use passed in recognizer
            analyzer_results = self.image_analyzer_engine.analyze(
                image,
                ocr_kwargs=ocr_kwargs,
                ad_hoc_recognizers=[redact_approach],
                **text_analyzer_kwargs,
            )
        else:
            raise ValueError("Please enter valid string or PatternRecognizer object for redact_approach")

        return analyzer_results
```

omri374: Thanks for providing more context. In my perspective, while this approach makes it easier to get started with several hypotheses, it might create issues for more advanced or customized scenarios. Here are some points I thought are worth raising:
- By reading the docstring, it is not clear what would happen on each case (default, metadata, PatternRecognizer). We can add more information to the docstring on each approach, but it feels like we would be replicating Presidio's documentation in this specific docstring, to some extent, and might confuse users.
- It might cause further confusion if the user passes an `AnalyzerEngine` object when instantiating this object, but the analyzer instance gets ignored because the redaction approach here is not relying on the `AnalyzerEngine` object. 
- Calling the Analyzer Engine with a pattern recognizer brings more value than just calling the pattern recognizer on its own, for example when it comes to context enhancement. Passing just the `PatternRecognizer` would not benefit from those improvements, and future ones too.

Another small issue (which can be easily fixed) is the `type(redact_approach)==PatternRecognizer` which wouldn't work if the user passed an object which inherits from `PatternRecognizer`. In that case, using `isinstance` is better.

Overall, the approach we've been taking with Presidio is to allow the user to customize the objects as much as needed. While this approach would make it easier to get started, it might prevent a user not knowing the codebase well enough to get stuck on future customizations, or to find it difficult to switch from one approach to another.
I'd be happy to further discuss this (here or in a call) in case I'm missing anything.

niwilso: Got it, that makes sense. As for suggestion # 3, I will introduce that in a separate PR as there are still some code changes I need to make before the allow list approach can be called upon.

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
niwilso: I have made changes to match Omri's suggestion of using a `use_metadata=` bool arg instead and allowing for `ad_hoc_recognizers` to be passed in and used along with metadata redaction.

Resolving this comment as the original feedback here is no longer relevant with the code change.

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
niwilso: And I'm planning on adding example notebooks in a later PR.

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
```diff
@@ -44,7 +48,11 @@ def redact_and_return_bbox(
         :param padding_width: Padding width to use when running OCR.
         :param crop_ratio: Portion of image to consider when selecting
         most common pixel value as the background color value.
+        :param use_metadata: Whether to redact text in the image that
+        are present in the metadata.
         :param ocr_kwargs: Additional params for OCR methods.
+        :param ad_hoc_recognizers: List of PatternRecognizer objects to use
```
omri374, line 54: Thanks! This is a great approach IMHO.

presidio-image-redactor/tests/test_dicom_image_redactor_engine.py
```diff
@@ -1157,6 +1158,126 @@ def test_add_redact_box_happy_path(
...
+    mock_engine: DicomImageRedactorEngine,
+    ad_hoc_recognizers: Optional[List[PatternRecognizer]],
+):
+    """Test error handling of DicomImageRedactorEngine _get_analyzer_results()
+
+    Args:
+        mock_engine (DicomImageRedactorEngine): DicomImageRedactorEngine object.
+        ad_hoc_recognizers(None or list): Ad-hoc recognizers to use.
+    """
+    with pytest.raises(Exception):
```
omri374, line 1268: Consider checking for `TypeError` and not any exception

omri374 (approved):
Thanks! LGTM

ayabel (approved):

presidio-image-redactor/tests/test_dicom_image_redactor_engine.py
niwilso: Will create a separate small PR to make this change such that we can get this current PR merged in.

Issue #1131, opened 2023-07-28
Not be able to use a random key in encrypt and decrypt operators


zizhong:
**Describe the bug**
Currently we cannot use a random key in encrypt and decrypt operators. The interface expects a str but a random key could be bytes that cannot be decoded easily to a str.
```
>>> import os
>>> key = os.urandom(8)
>>> key_str = key.decode('utf-8')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x91 in position 1: invalid start byte
```

**Expected behavior**
encrypt and decrypt operators can take any bytes as the key.

PR #1147, opened 2023-08-21
Anonymizer - Pass bytes key to Encrypt / Decrypt

SharonHart:
## Change Description

In addition to string typed key, add the ability to pass a bytes key (for as-a-package use)
## Issue reference

This PR fixes issue #1131 

## Checklist


presidio-anonymizer/tests/integration/test_deanonymize_engine.py
```diff
@@ -20,7 +20,7 @@ def test_given_operator_decrypt_with_valid_params_then_decrypt_text_successfully
     decryption = engine.deanonymize(
         text,
         encryption_results,
-        {"DEFAULT": OperatorConfig(Decrypt.NAME, {"key": "WmZq4t7w!z%C&F)J"})},
+        {"DEFAULT": OperatorConfig(Decrypt.NAME, {"key": b'WmZq4t7w!z%C&F)J'})},
```
niwilso, line 23: **nit**: Should the single quotes be double quotes to be consistent with other quotes throughout the file?

niwilso (approved):

presidio-anonymizer/tests/integration/test_deanonymize_engine.py
SharonHart: maybe, I am used to seeing byte string represented with single quotes, not sure what is the convention there, but all bytestrings in the repo are with single quotes so it is consistent

omri374, line 23: Does this require a change [in this notebook](https://github.com/microsoft/presidio/blob/main/docs/samples/python/encrypt_decrypt.ipynb) and in the [tutorial](https://github.com/microsoft/presidio/blob/main/docs/tutorial/12_encryption.md]?

presidio-anonymizer/tests/services/test_aes_cypher.py
```diff
@@ -14,6 +16,7 @@
         (b'1111111111111111', "面汤"),  # Chinese text
```
omri374, line 16: If a string is a valid input which would get encoded as part of the process, should we have tests checking the validity of a string input as well?

presidio-anonymizer/tests/integration/test_deanonymize_engine.py
SharonHart: No, its another option, kept one test with that option, one with the string option

presidio-anonymizer/tests/services/test_aes_cypher.py
SharonHart: At this level only bytes are passed, added both string and bytes option in tests for upper level

omri374 (approved):

PR #1150, opened 2023-08-21
Improve process names method in DICOM image redactor

niwilso:
## Change Description

Adding slight improvements to the `_process_names()` method in the `DicomImageRedactorEngine`. The new `augment_word` method can also be called by the user for the allow list approach.

## Issue reference

This PR fixes issue #XX

## Checklist


presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
```diff
@@ -634,7 +634,49 @@ def _get_text_metadata(
...
+    def augment_word(word: str) -> list:
+        """Apply multiple types of casing to the provided string.
+
+        :param words: String containing the word or term of interest.
+
+        :return: List of the same string with different casings and spacing.
+        """
+        word_list = []
+        if word != "":
+            word_list = []
```
SharonHart, line 646: redundant

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
```diff
@@ -634,7 +634,49 @@ def _get_text_metadata(
...
+
+            # Replacing separator character with space, if any
+            text_1 = word.replace("^", " ")
+            text_1 = text_1.replace("-", " ")
```
SharonHart, line 650: nit: I would give more descriptive names

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
```diff
@@ -634,7 +634,49 @@ def _get_text_metadata(
...
+            text_4 = text_1.title()
+
+            # Append iterations
+            word_list.append(text_1)
+            word_list.append(text_2)
+            word_list.append(text_3)
+            word_list.append(text_4)
+
+            # Adding each term as a separate item in the list
+            word_list = word_list + text_1.split(" ")
```
SharonHart, line 668: word_list += [text_1.split(" "), text_2.split(" "), text_3.split(" "), text_4.split(" ")]

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
```diff
@@ -634,7 +634,49 @@ def _get_text_metadata(
...
+        :param words: String containing the word or term of interest.
+
+        :return: List of the same string with different casings and spacing.
+        """
+        word_list = []
+        if word != "":
+            word_list = []
+
+            # Replacing separator character with space, if any
+            text_1 = word.replace("^", " ")
```
SharonHart, line 649: I would use more descriptive names

presidio-image-redactor/tests/test_dicom_image_redactor_engine.py
```diff
@@ -620,6 +620,60 @@ def test_get_text_metadata_happy_path(
...
 
+# ------------------------------------------------------
+# DicomImageRedactorEngine.augment_word()
+# ------------------------------------------------------
+@pytest.mark.parametrize(
+    "word, expected_list",
+    [
+        ("", []),
+        (" ", ["", " "]),
+        ("JOHN^DOE",[
```
SharonHart, line 631: there was something about '-' as well

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
```diff
@@ -634,7 +634,49 @@ def _get_text_metadata(
...
+
+            # Capitalize all characters in string
+            text_2 = text_1.upper()
+
+            # Lowercase all characters in string
+            text_3 = text_1.lower()
+
+            # Capitalize first letter in each part of string
+            text_4 = text_1.title()
+
```
SharonHart, line 660: would it make sense to match this from the other side?
The client of word_list would match lower() to lower()?

SharonHart (approved):

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
niwilso: I included `.lower()` and the other casing transformations inside this method such that it could be used by (1) the DICOM image redactor engine when metadata is used and (2) users who wish to call the method directly to augment any list of words they may have (e.g., allow list).

SharonHart: In that case, it might even be better to not reduce it to lowecase.
So that in (2), users that care about casing don't lose that information and can decide if they want to exact match or do .lower()
the DICOM image redactor engine can do .lower() itself on the input vs each element in the returned list.  
Or am I missing something?
Anyway, up to you

niwilso: Good point. I have added the bool argument `case_sensitive` such that users can specify as needed.

SharonHart (dismissed):

omri374 (approved):

Issue #1083, opened 2023-06-01
Using Presidio with Huggingface support


Matei9721:
Hi, currently I am using presidio with Spacy and Stanza by creating an `nlp_engine` using `NlpEngineProvider` and passing it the correct model in the config. I was planning on adding support for HuggingFace transformer models, but I was a bit confused by the fact that there are 2 ways of doing this:

1. Using a `TransformerRecognizer`
2. Using `TransformerNlpEngine`

As far as I understand, if you use the recognizer then you apply the recognizer on top of the usual e.g. Spacy NER pipeline so you will get results from both Spacy and HuggingFace model. On the other hand, using the TransformerNlpEngine substitutes the Spacy NER module in the pipeline.

In this example: https://microsoft.github.io/presidio/samples/python/transformers_recognizer/ it is shown how to use the TransformersRecognizer with a specific configuration given as an example in configuration.py where you can do the `MODEL_TO_PRESIDIO_MAPPING`. If you are to use the `TransformerNlpEngine`, how are you supposed to do the mapping between model types and presidio types similar to the ones done in `TransformerRecognizer`?

Is my understanding above right and if yes, is there a way to create an `AnalyzerEngine `with a `TransformerNlpEngine `with the same configuration as a `TransformerRecognizer`?

Thanks for the help!

Matei9721:
Actually, after checking the source code more, it's actually not clear to me how one is supposed to use the `TransformerNlpEngine`. What is the `TransformersComponent` class used for in this case?

Using the `TransformerRecognizer` seems easier as there are more code examples, but is it advised to use it over `TransformerNlpEngine`?

omri374:
Hi @Matei9721, thanks for your feedback! I can understand why this causes confusion. We initially wanted to support Huggingface the same way we support Stanza, but bumped into some issues. In the future, the plan is to integrate the new [spacy-huggingface-pipelines](https://github.com/explosion/spacy-huggingface-pipelines) package for a more seamless integration.

[...]

PR #1159, opened 2023-08-28
Integrating spacy-huggingface-pipelines and refactoring NlpEngine logic

omri374:
## Change Description

This PR improves the handling of transformers models using the `NlpEngine` flow. Note that there are two ways to introduce NER models into presidio: `NlpEngine` and a standalone recognizer. See more info on #1083. 

Main changes:
1. The entire handling of models, entities, mapping and alignment of model spans to Presidio's spans is done as part of the `NlpEngine` config (either `SpacyNlpEngine` or `TransformersNlpEngine`). In the future, we could also add `FlairNlpEngine` or other NER packages.
2. Introduced a new `NerModelConfiguration` dataclass which holds the user configuration. Configuration can be set in a `conf` file, for example: 
image
with most of the configuration coming from here: https://github.com/explosion/spacy-huggingface-pipelines#token-classification
3. Simplified the logic in `SpacyRecognizer` which now only gets the entities out of `NlpArtifacts` and returns them
4. Integrated `spacy-huggingface-pipelines` to improve the handling of transformer models inside spaCy pipelines.

Flow before:
```mermaid
sequenceDiagram
    AnalyzerEngine->>SpacyNlpEngine: Call engine.process_text(text) <br>to get model results
    SpacyNlpEngine->>Spacy: call spaCy NER model
    Spacy->>SpacyNlpEngine: return PII entities
    SpacyNlpEngine->>AnalyzerEngine: Pass NlpArtifacts<BR>(Entities, lemmas, tokens etc.)
    Note over AnalyzerEngine: Call all recognizers
    AnalyzerEngine->>SpacyRecognizer: Pass NlpArtifacts
    Note over SpacyRecognizer: Extract PII entities out of NlpArtifacts, <BR>map to Presidio entities, <BR>filter requested entities, <BR> add confidence scores
    SpacyRecognizer->>AnalyzerEngine: Return List[RecognizerResult]

```

Flow after:
```mermaid
sequenceDiagram
    AnalyzerEngine->>SpacyNlpEngine: Call engine.process_text(text) <br>to get model results
    SpacyNlpEngine->>Spacy: call spaCy NER model
    Spacy->>SpacyNlpEngine: return PII entities
    Note over SpacyNlpEngine: Map entity names to Presidio's, <BR>update scores, <BR>remove unwanted entities <BR> based on NerModelConfiguration
    SpacyNlpEngine->>AnalyzerEngine: Pass NlpArtifacts<BR>(Entities, lemmas, tokens, scores etc.)
    Note over AnalyzerEngine: Call all recognizers
    AnalyzerEngine->>SpacyRecognizer: Pass NlpArtifacts
    Note over SpacyRecognizer: Extract PII entities out of NlpArtifacts
    SpacyRecognizer->>AnalyzerEngine: Return List[RecognizerResult]

```

## Issue reference

This PR fixes issue #1083 

## Checklist

- [X] My code includes unit tests: partially/wip
- [ ] My PR contains documentation updates / additions if required: wip


presidio-analyzer/presidio_analyzer/nlp_engine/transformers_nlp_engine.py
```diff
@@ -2,153 +2,168 @@
...
+
+    def process_text(self, text: str, language: str) -> NlpArtifacts:
+        """Execute the SpaCy NLP pipeline on the given text and language."""
+        if not self.nlp:
+            raise ValueError("NLP engine is not loaded. Consider calling .load()")
+
+        doc = self.nlp[language](text)
+        return self._doc_to_nlp_artifact(doc, language)
+
+    def _get_entities(self, doc: Doc) -> SpanGroup:
```
omri374, line 123: This part does the filtering, mapping, scoring etc. that was previously in SpacyRecognizer/TransformersRecognizer.

omri374:
Hi @LSD-98, thanks for much for reviewing this and trying it out!
By design, Presidio would not be installed with transformers, to reduce the installation complexity. Eventually when this is released to PyPI, you would be able to install it using this command: `pip install presidio-analyzer["transformers"]`.

Before released, you can install it locally by cloning the repo and calling `pip install -e ".[transformers]"`
Please try it out and let me know? You can also explicitly install the two dependencies: `spacy-huggingface-pipelines` and `transformers`.

LSD-98:
> Hi @LSD-98, thanks for much for reviewing this and trying it out! By design, Presidio would not be installed with transformers, to reduce the installation complexity. Eventually when this is released to PyPI, you would be able to install it using this command: `pip install presidio-analyzer["transformers"]`.
> 
> Before released, you can install it locally by cloning the repo and calling `pip install -e ".[transformers]"` Please try it out and let me know? You can also explicitly install the two dependencies: `spacy-huggingface-pipelines` and `transformers`.

Thank you for your quick reply. I had the ```transformers``` package but not the ```spacy-huggingface-pipelines``` one :) Working perfectly now !

presidio-analyzer/setup.cfg
```diff
@@ -7,4 +7,4 @@ exclude =
     dist,
     tests
 docstring-convention = numpy
-extend-ignore = E203 D100 D202 ANN101 ANN102 ANN204 ANN203
\ No newline at end of file
+extend-ignore = E203 D100 D202 ANN101 ANN102 ANN204 ANN203 TC
```
SharonHart, line 10: What's TC?

presidio-analyzer/tests/conf/default.yaml
```diff
@@ -4,3 +4,16 @@ models:
     lang_code: en
     model_name: en_core_web_lg
 
+ner_model_configuration:
+  model_to_presidio_entity_mapping:
+    PER: PERSON
+    PERSON: PERSON
+    NORP: NRP
+    FAC: FACILITY
```
SharonHart, line 12: Do we have a FACILITY entitiy in presidio?

presidio-analyzer/tests/test_ner_model_configuration.py
```diff
@@ -0,0 +1,57 @@
...
+    [
+        ("stride", []),
+        ("stride", "X"),
+        ("stride", None),
+        ("alignment_mode", 5),
+        ("alignment_mode", None),
+        ("low_confidence_score_multiplier", "X"),
+    ],
+)
+def test_from_dict_wrong_types(ner_model_configuration_dict, key, value):
```
SharonHart, line 52: We should look into incorporating pydantic at some point, for automatic yaml loading, type validation, and at some point it migrating to FastAPI that includes it

SharonHart (approved):

LSD-98:
Hello @omri374,

By running some test i got a warning from ```spacy-huggingface-pipeline```: 
```UserWarning: Skipping annotation, {'entity_group': 'LOC', 'score': 0.46050456, 'word': 'ian', 'start': 33, 'end': 36} is overlapping or can't be aligned for doc 'Je suis Tom et je travaille à Ardian' ```

In the end, it detected the right entities to anonymize but wondered whether it could lead to a "false negative" (i.e. skipping an entity that should be detected). I saw you also had the same issue [here](https://github.com/explosion/spaCy/discussions/12998). I did not understand the technical points of the answer so I wanted to check if you got a solution for this.

Many thanks in advance!

PS : please tell me if I should post here or on the discussion tab.

File: presidio-analyzer/tests/conf/default.yaml (hunk 3 above)
Comment (omri374) [reply presidio-analyzer/tests/conf/default.yaml:12]: Not right now, and it hardly appears as an output of Spacy. One option would be to add it to the supported entities list, the other is to map it to location. WDYT?

File: presidio-analyzer/tests/test_ner_model_configuration.py (hunk 4 above)
Comment (omri374) [reply presidio-analyzer/tests/test_ner_model_configuration.py:52]: Totally agree

**Comment by omri374** [comment]
@LSD-98 thanks for raising this. This is caused by a mismatch between wordpiece tokenization (in transformers) and the spaCy tokenizer. It is defined as "expand" as usually wordpiece tokens are a subset of a spaCy token, but there could be cases where this might result in a false negative or false positive (most likely false positive, as the wordpiece token would be expanded to more than the PII itself). I don't see an immediate workaround for this, so I guess we'd have to live with it. If someone is making sure there are no alignment errors, they could still add a transformers recognizer as an independent recognizer and not as part of the NLPEngine mechanism.

File: presidio-analyzer/tests/conf/default.yaml (hunk 3 above)
Comment (SharonHart) [reply presidio-analyzer/tests/conf/default.yaml:12]: Map it to location IMO

**Review by SharonHart — DISMISSED** [review]

File: presidio-analyzer/tests/conf/default.yaml (hunk 3 above)
Comment (omri374) [reply presidio-analyzer/tests/conf/default.yaml]: Done

**Review by navalev — APPROVED** [review]

File: presidio-analyzer/Pipfile [hunk 5]
```diff
@@ -4,16 +4,20 @@ verify_ssl = true
...
 [packages]
-spacy = ">=3.4.4,<4.0.0"
+spacy = ">=3.4.4, <4.0.0"
 regex = "*"
 tldextract = "*"
 flask = ">=1.1"
 pyyaml = "*"
 phonenumbers = ">=8.12,<9.0.0"
 typing-extensions = "*"
+spacy-huggingface-pipelines = "*"
```
niwilso, line 14: Do we need to add this package to the root-level NOTICE file?

niwilso: Same with stanza

presidio-analyzer/conf/transformers.yaml
```diff
@@ -4,4 +4,41 @@ models:
...
-      transformers: elastic/distilbert-base-uncased-finetuned-conll03-english
+      transformers: StanfordAIMI/stanford-deidentifier-base
+
+ner_model_configuration:
+  labels_to_ignore:
+  - O
+  aggregation_strategy: simple # "simple", "first", "average", "max"
+  stride: 16  # If stride >= 0, process long texts in
+              # overlapping windows of the model max
+             # length. The value is the length of the
```
niwilso, line 15: **nit**: Spacing for start of comment is different for above lines and below lines.

presidio-analyzer/presidio_analyzer/nlp_engine/ner_model_configuration.py
```diff
@@ -0,0 +1,121 @@
...
+            )
+
+    @staticmethod
+    def __validate_type(config_dict: Dict, key: str, field_type: Type) -> None:
+        if key in config_dict:
+            if not isinstance(config_dict[key], field_type):
+                raise ValueError(f"{key} must be of type {field_type}")
+
+    @classmethod
+    def from_dict(cls, nlp_engine_configuration: Dict) -> "NerModelConfiguration":
```
niwilso, line 104: Just to confirm, is the output type hint supposed to be the current string "NerModelConfiguration"?

presidio-analyzer/presidio_analyzer/nlp_engine/nlp_artifacts.py
```diff
@@ -18,15 +26,17 @@ def __init__(
         tokens: Doc,
         tokens_indices: List[int],
         lemmas: List[str],
-        nlp_engine,  # noqa ANN001
+        nlp_engine: "NlpEngine",  # noqa F821
```
niwilso, line 29: Also checking that "NlpEngine" here is supposed to be a string

presidio-analyzer/presidio_analyzer/nlp_engine/nlp_artifacts.py
```diff
@@ -18,15 +26,17 @@ def __init__(
...
         language: str,
+        scores: Optional[List[float]] = None,
     ):
         self.entities = entities
         self.tokens = tokens
         self.lemmas = lemmas
         self.tokens_indices = tokens_indices
         self.keywords = self.set_keywords(nlp_engine, lemmas, language)
         self.nlp_engine = nlp_engine
+        self.scores = scores if scores else [0.85] * len(entities)
```
niwilso, line 39: Why auto assign 0.85 by default if no scores are available?

presidio-analyzer/presidio_analyzer/nlp_engine/nlp_engine_provider.py
```diff
@@ -84,11 +84,20 @@ def create_engine(self) -> NlpEngine:
...
+            )
+            if ner_model_configuration:
+                ner_model_configuration = NerModelConfiguration.from_dict(
+                    ner_model_configuration
+                )
+
+            engine = nlp_engine_class(
+                models=nlp_models, ner_model_configuration=ner_model_configuration
+            )
+            engine.load()
```
niwilso, line 100: Would it make sense to include a call to load inside init such that the user does not need to specify engine load at any point?

presidio-analyzer/presidio_analyzer/nlp_engine/spacy_nlp_engine.py
```diff
@@ -76,20 +138,99 @@ def get_nlp(self, language: str) -> Language:
...
+                    f"Add to `NerModelConfiguration.labels_to_ignore` to remove."
+                )
+
+            # Remove presidio entities in the ignore list
+            if ent.label_ in to_ignore:
+                continue
+
+            new_entities.append(ent)
+
+            # Update score if entity is in low score entity names
```
niwilso, line 230: Can the score ever be updated to be higher than the default value?

presidio-analyzer/presidio_analyzer/nlp_engine/transformers_nlp_engine.py
```diff
@@ -1,154 +1,134 @@
...
 
 import spacy
-from spacy.language import Language
 from spacy.tokens import Doc, Span
 
-from presidio_analyzer.nlp_engine import SpacyNlpEngine
-
-
 try:
-    import torch
```
niwilso, line 12: Are we still using PyTorch elsewhere? If not, consider removing from requirements and maybe the notice?

niwilso (changes requested):
Overall looks great! Thank you for providing the swimlane flow diagram.

Just a few points to address before approval.

presidio-analyzer/Pipfile
omri374: Good catch, thanks.

presidio-analyzer/presidio_analyzer/nlp_engine/ner_model_configuration.py
omri374: The output type is `NerModelConfiguration`, but because method is a classmethod for the `NerModelConfiguration` class, we cannot declare it. I'm not sure if there's a different practice to do this.

presidio-analyzer/presidio_analyzer/nlp_engine/transformers_nlp_engine.py
omri374: Good catch

presidio-analyzer/Pipfile
omri374: Done

presidio-analyzer/conf/transformers.yaml
omri374: Done

presidio-analyzer/presidio_analyzer/nlp_engine/nlp_artifacts.py
omri374: Good catch, fixing.

presidio-analyzer/presidio_analyzer/nlp_engine/nlp_engine_provider.py
omri374: This separation is useful if a user wants to create an NlpEngine in a different way, or when they wish to inject a model object instead of always loading it as part of instantiation.  It also reduces the instantiation time (helpful in tests and mocks) at the expense of taking more time on the first run. 

Do you think it should be automatically loaded? I don't have any strong preferences here.

presidio-analyzer/presidio_analyzer/nlp_engine/spacy_nlp_engine.py
omri374: There's no limitation here. Do you think there should be?

presidio-analyzer/presidio_analyzer/nlp_engine/nlp_artifacts.py
omri374: Just checked again, it has to be in quotes, otherwise we have a circular import as the `NlpEngine` class holds `NlpArtifacts`

presidio-analyzer/presidio_analyzer/nlp_engine/nlp_artifacts.py
niwilso: Understood. Is this something that should be revisited from a whole Presidio level? Perhaps I am just lacking context, but returning 0.85 as a fixed default value seems misleading to me.

No need to address in this PR.

presidio-analyzer/presidio_analyzer/nlp_engine/nlp_engine_provider.py
niwilso: Got it, thank you for describing the scenarios where this is helpful. Let's keep it as-is then.

presidio-analyzer/presidio_analyzer/nlp_engine/spacy_nlp_engine.py
niwilso: Looking at how scores are handled in this file, it seems that scores are initially assigned the default value and then only the scores associated with certain entities are updated, namely, by being multiplied by the low confidence score multiplier.

However, from a quick search, I now see that spacy NER might not return confidence scores. To clarify, is this the reason for this logic?

niwilso (approved):

presidio-analyzer/presidio_analyzer/nlp_engine/nlp_artifacts.py
omri374: We can definitely consider changing this. It was like they even before my time 🙂

presidio-analyzer/presidio_analyzer/nlp_engine/spacy_nlp_engine.py
omri374: The intention is to allow users to handle entities that have more errors on their data. Models either return a fixed confidence value (like 0.85) or the softmax output (on transformers based models). Still, there are cases where different entities have higher error rates.

omri374: I would be happy to consider other approaches to this. There is definitely a requirement to handle different entities differently.

presidio-analyzer/presidio_analyzer/nlp_engine/nlp_artifacts.py
niwilso: Create a new issue to revisit this in the future: https://github.com/microsoft/presidio/issues/1190

presidio-analyzer/presidio_analyzer/nlp_engine/spacy_nlp_engine.py
niwilso: Got it. In this case, I am okay with leaving it as-is, but we can revisit in the future as needed.

SharonHart (approved):

PR #1162, opened 2023-08-30
Updating verification engines to include latest updates to redactor engines

niwilso:
## Change Description

Updating both the standard and the DICOM image PII verification engines to allow use of ad-hoc recognizers and the allow list approach, as recently enabled with the redactor engines.

## Issue reference

This PR fixes issue #XX

## Checklist

### DIFF
--- a/presidio-image-redactor/presidio_image_redactor/image_pii_verify_engine.py
+++ b/presidio-image-redactor/presidio_image_redactor/image_pii_verify_engine.py
@@ -21,11 +21,6 @@ def fig2img(fig):
 class ImagePiiVerifyEngine(ImageRedactorEngine):
     """ImagePiiVerifyEngine class only supporting Pii verification currently."""
 
-    def __init__(self, image_analyzer_engine: Optional[ImageAnalyzerEngine] = None):
-        if not image_analyzer_engine:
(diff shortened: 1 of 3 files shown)


presidio-image-redactor/presidio_image_redactor/image_pii_verify_engine.py
```diff
@@ -25,7 +27,11 @@ def __init__(self, image_analyzer_engine: Optional[ImageAnalyzerEngine] = None):
         self.image_analyzer_engine = image_analyzer_engine
```
omri374, line 27: Can we use the super `__init__` instead?

omri374 (dismissed):

presidio-image-redactor/presidio_image_redactor/image_pii_verify_engine.py
niwilso: Yes, will remove in the next commit, verify that no issues arise, and then resolve this thread.

niwilso: @omri374 - Feedback addressed and all tests pass. Resolving this thread.

omri374 (approved):

Issue #1195, opened 2023-10-28
Effective handling of intersecting entities such a way that it won't change the original text. 


VMD7:
**Describe the bug**
I have one unique user journey where I have to mask the PII for one usecase, then again I have to unmasked the PII of same data in another usecase. Its like flowing river where at one village I have to travel under the soil and at another village I have to flow over the soil. The problem I am facing is little unique, now whenever there are two different entities gets detected for some portions of the sentences, and these portions got intersected each other in small portion. Now, when I wanted to get the data back after unmasking the text I am getting changed original text. Due to not handling conflict of these intersections of entities properly. 

**Sample Code**
```
from presidio_analyzer import RecognizerResult, AnalyzerEngine
from presidio_anonymizer import AnonymizerEngine
from presidio_anonymizer.entities import OperatorConfig

text = """Fake card number 4151 3217 6243 3448.com that overlaps with nonexisting URL."""

anomyzerEngine = AnonymizerEngine()
analyzerEngine = AnalyzerEngine()

result = analyzerEngine.analyze(text, language= 'en')

deIdentifiedText = anomyzerEngine.anonymize(
    text=text, analyzer_results=result, operators={"DEFAULT": OperatorConfig("keep")} 
)

print("De-identified text")
print(deIdentifiedText.text)
```

**Results Generated**
`De-identified text`
`Fake card number 4151 3217 6243 34483448.com that overlaps with nonexisting URL.`
-> Here its detected entities list [type: CREDIT_CARD, start: 17, end: 36, score: 1.0, type: URL, start: 32, end: 40, score: 0.5].
-> Now if you see the two entities start and end positions (17, 36) and (32, 40). This edge case making the problem while getting the text original because it is unnecessarily adding extra 3448 number in  the above example. This extra is getting inserted due to common intersection portion of the text from (32, 36) and this changing the original text completely different. 

**Expected behavior**
`De-identified text`
`Fake card number 4151 3217 6243 3448.com that overlaps with nonexisting URL.`
-> It should give the original text as it is without modifying it. and its entities list should look like: [type: CREDIT_CARD, start: 17, end: 36, score: 1.0, type: URL, start: 36, end: 40, score: 0.5]. This will help us to identify the both the entities without loosing them and without modifying the original text. 

**Additional context**
This can be solved by improving the conflict handling function in AnonymizerEngine. I have solved this issue, will raise PR for the same.

PR #1196, opened 2023-10-28
Improved the logic of conflict handling in AnonymizerEngine

VMD7:
Improved the logic of  `_remove_conflicts_and_get_text_manipulation_data` method in `AnonymizerEngine` for a edge case when entities gets intersected then they are changing the original text data due to overlap portion. 

## Change Description
-> In `_remove_conflicts_and_get_text_manipulation_data`  method all the scenarios handled. Only, the issue coming when two different entities detects PII and have some common portion of text in between them. 
-> Now if we try to raise this as conflict in method `_is_result_conflicted_with_other_elements`. Then both entities might get lost, so this might not be the right way to handle it. 
-> So, after getting the list of result of all entities of `_remove_conflicts_and_get_text_manipulation_data` method at end in `_unique_text_metadata_elements` list. Will first sort them ascending order by considering the "start" key. 
-> Then will check is there any conflict is happening or not, if yes then will adjust the start and end positions, without removing them. 

## Issue reference

This PR fixes issue #1195 

## Checklist


VMD7:

VMD7:
Hi @omri374 do I need to change anything in code.
-> The 2/45 test cases are failing due to the change because the test cases you made also changing the original text. So, please check the test cases for the same. 
-> There are some another tests failing they are related with Linting, I will work on it.

omri374:
@VMD7 thanks for this addition. Could you please add a test case which checks the new logic specifically?

VMD7:
Hi @omri374 
-> Added the test case for new logic of conflict management.
-> It passes the all test cases except 1 which is below
```
def test_given_custom_anonymizer_then_we_manage_to_anonymize_successfully():
    engine = AnonymizerEngine()
    text = (
        "Fake card number 4151 3217 6243 3448.com that "
        "overlaps with nonexisting URL."
    )
    analyzer_result = RecognizerResult("CREDIT_CARD", 17, 36, 0.8)
    analyzer_result2 = RecognizerResult("URL", 32, 40, 0.8)
    anonymizer_config = OperatorConfig("custom", {"lambda": lambda x: f"<ENTITY: {x}>"})
    result = engine.anonymize(
        text, [analyzer_result, analyzer_result2], {"DEFAULT": anonymizer_config}
    ).text
    resp = (
        "Fake card number <ENTITY: 4151 3217 6243 3448>"
        "<ENTITY: 3448.com> that overlaps with nonexisting URL."
    )
    assert result == resp
```

Now, if you observe the problem with this test case is that. It gives the overlapped entities. But, It fails to ensure that after unmasking or decrypting the text do we get the original text back. So, my suggestion would be we should modify this test case by below changes in `resp` such as.  The expected `resp` should be:
``` resp = (
        "Fake card number <ENTITY: 4151 3217 6243 3448>"
        "<ENTITY:  .com> that overlaps with nonexisting URL."
    )```
    
I will commit the changes for this as well. Please let me know your thoughts on it.

**Comment by omri374** [comment]
Hi @VMD7. There are two use cases for conflict resolution:
1. Getting the de-anonymized text correctly with no duplications
2. For each entity, regardless if it is conflicting with another or not, we should be able to de-anonymize the anonymized value and get the previous value, the original spans and the anonymized spans.

Is your suggested change solving those two use cases? If it only solves 1, maybe it would be better to tackle it on the text building part, and not the entity conflicts part.

**Comment by omri374** [comment]
Hi @VMD7, thanks for creating this PR! Would you be interested in continuing to work on it? Perhaps we can collaborate?

**Comment by VMD7** [comment]
Hi @omri374,
Thanks for acknowledging. I wanted to work on this thread. But, little bit got stuck in that. Your collaboration definitely help me to implement further.

**Comment by VMD7** [comment]
Hi @omri374 
Yes, this conflict resolution strategy looks nice. I have following doubts for each case, please guide me on this.

**Case A: TEXT** = "text":: 
-> I think we have to modify in the _EngineBase class_  -> __operate method_ . 
-> I have gave a try to directly modify the text without modifying the entites span or entities text. Its working for the all the operators. 
-> The one problem I could see is that in encryption and decryption. In encryption we are getting fine output. In decryption also we are getting perfect output without overlapping the text. But, the text spans are getting changed for decrypt because while encrypting we are giving modified text. So, its expected behaviour. (Please let me know your thoughts on this).

**Case B: ENTITIES = "entities"**::
-> One doubt in this case, if we remove the conflict from entities completely then there won't be any overlapping problem of text. So, indirectly we are completing the third case here as well.
    
**Case C: TEXT_AND_ENTITIES = "text_and_entities"**::
-> Yes the earlier written piece of code works for this scenario. Also, if any additional changes needed please let me know.

**Case D: NONE = "none"**::
-> Are we planning to ignore all the conflicts (# 1156) like the conflicts resolved by some assumptions as well.

Please let me know your thoughts on this.

**Comment by omri374** [comment]
Thanks @VMD7. I suppose we don't have to implement all the possible options. How would you suggest to simplify? Implement A,C,D?

**Comment by VMD7** [comment]
Hi @omri374 
For simplicity can we go with C and D only. Like if user wants to complete remove the conflicts then they will use C case and if user want to go with current conflict management (that contains some assumptions) then they will choose D.  This can be implemented using simple _remove_conflicts_ flag in anonymizer engine.

To implement case A - I am not getting confident because somehow we have to change the entity and their position to avoid the overlap on text side. So, indirectly we are changing the entity positions for text side handling. 

What do you say on this. Should I proceed for the code changes.

**Comment by omri374** [comment]
Yes, let's go with C and D. Should we have this as an enum anyway? If for instance we'd like to add a new strategy for conflict handling, we wouldn't have to change the API. WDYT?

**Comment by VMD7** [comment]
Yes sure using enum it will be better. I will add case C and D with enum and test cases related with that and commit the update code.

**Comment by omri374** [comment]
Awesome, thanks!!

**Comment by VMD7** [comment]
Hi @omri374 
Added all the required changes for conflict resolution strategy. Please test and review it once and let me know if any further improvements needed.

**Comment by VMD7** [comment]
Hi @omri374 
Any comments on these changes.

File: presidio-anonymizer/tests/test_anonymizer_engine.py [hunk 1]
```diff
@@ -255,3 +256,46 @@ def _operate(
...
+        text,
+        [analyzer_result, analyzer_result2],
+        {"DEFAULT": anonymizer_config},
+        conflict_resolution=conflict_strategy
+    ).text
+
+    assert result == text
+
+
+def test_when_none_conflict_selected_then_no_conflict_handled():
```
omri374, line 281: Can you please add another case where conflict_stragy=None (which is the default)?

If the tests are identical, but results are different, consider using pytest.mark.parameterize

omri374 (commented):
Looks great!
Thinking about the default case, where ConflictResolutionStrategy is None, maybe it's better to pass a default value instead of None.

presidio-anonymizer/tests/test_anonymizer_engine.py
VMD7: Hi @omri374 
Added one more test case where conflict_strategy=None. Please let me know if any further changes.

VMD7:
Hi @omri374 
Added one more test case where _conflict_strategy = None._ Pleas check it once and let me know if any further changes required.

For default case - Yes right, instead of using _ConflictResolutionStrategy.NONE_ we can use _ConflictResolutionStrategy.DEFAULT_ as it sounds more relevant. 

One more thought should we make three cases:
**a) _ConflictResolutionStrategy.DEFAULT_**  -> This will be our default strategy (that is current strategy)
**b) _ConflictResolutionStrategy.TEXT_AND_ENTITIES_** -> This will further removes the conflicts from text and entities (The written piece of code already).
**c) _ConflictResolutionStrategy.NONE_** -> This will be our new strategy where we will not handle any type of conflicts will give the results as it that are coming from analyzer engine. 

Or should I simply replace the _ConflictResolutionStrategy.NONE_ by _ConflictResolutionStrategy.DEFAULT_ for now.  Please let me know your thoughts on it.

presidio-anonymizer/tests/test_anonymizer_engine.py
```diff
@@ -255,3 +256,61 @@ def _operate(
...
+    operator_config = OperatorConfig("keep")
+    conflict_strategy = ConflictResolutionStrategy.TEXT_AND_ENTITIES
+    result = engine.anonymize(
+        text,
+        [analyzer_result1, analyzer_result2],
+        {"DEFAULT": operator_config},
+        conflict_resolution=conflict_strategy
+    ).text
+
+    assert result == text
```
omri374, line 278: Should we put the score into consideration as well? Right now the logic would reduce the span of the later entity, regardless of score. Just thinking, maybe we should decide what to trim based on the score. WDYT?

omri374:
Hi, thanks for the updates!
I agree with you that we should have a default here which is identical to the previous logic.
We should make sure the description of each strategy is clear, and I'm not sure DEFAULT and TEXT_AND_ENTITIES would say much to a user interested in understanding which strategy to choose.

Are these good descriptors of the strategies?
a. MERGE_SIMILAR_OR_CONTAINED (default one we have today)
b. REMOVE_INTERSECTIONS (new logic)
c. NONE (do nothing, including a)

We should also mention that b will also trigger a. 
WDYT?

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
```diff
@@ -142,6 +152,26 @@ def _remove_conflicts_and_get_text_manipulation_data(
...
+        # All types of conflicts among entities as well as text.
+        if conflict_resolution == ConflictResolutionStrategy.TEXT_AND_ENTITIES:
+            unique_text_metadata_elements.sort(key=lambda x: x.start)
+
+            for i in range(len(unique_text_metadata_elements)):
+                result = unique_text_metadata_elements[i]
+                j = 0
+                while j < i:
+                    existing_result = unique_text_metadata_elements[j]
+                    if (result.end <= existing_result.start or
```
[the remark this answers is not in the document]
VMD7: Yes we can use this but the problem is I need "=" case as well. In ```RecognizerResult.intersects``` it doesn't cover the ```=``` scenario. It just compares ```>  or  <``` only.

presidio-anonymizer/tests/test_anonymizer_engine.py
VMD7: Yes right so updated the logic by considering the score as well while changing the entity spans. Now it will follow below cases while comparing between two entities.
1) If score are same or first entity score is higher than the other then it will give priority to first one such that it will adjust the spans of second entity only.
2) If second entity score is higher than first one, then it will give priority to second one and adjust the spans of first one only.

VMD7:
Hi @omri374 
Yes, these descriptions looks good. I will update the code by considering these cases. As well as will add some more test cases for updated logic of intersection where we are considering the score as well.

VMD7:
Hi @omri374 
Updated all the required changes. Please review it once and let me know if any improvements needed.

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
```diff
@@ -98,6 +111,10 @@ def _remove_conflicts_and_get_text_manipulation_data(
         2. Have the same indices as other results but with larger score.
         :return: List
         """
+        # Return without modifying analyzer_results.
+        if conflict_resolution == conflict_resolution.NONE:
+            return analyzer_results
+
         tmp_analyzer_results = []
         # This list contains all elements which we need to check a single result
```
omri374, line 119: Should we put this under `ConflightResolutionStrategy.MERGE_SIMILAR_OR_CONTAINED`?

omri374 (commented):
Thanks! looks great. One last comment if I may :)

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
omri374: Another approach would be to remove "None" from being one of the option. Otherwise there's no difference between MERGE_SIMILAR_AND_CONTAINED and None

VMD7: Hi @omri374 
-> Actually could you please explain it little bit more. Like, under ```ConflightResolutionStrategy.MERGE_SIMILAR_OR_CONTAINED``` means inside the code logic of MERGE_SIMILAR_OR_CONTAINED by setting some conditions. If this way then this can be possible but it will unnecessary increase the time complexity for None case. 
-> Are you talking about the "None" case that is handled below:
```
if not conflict_resolution:
            conflict_resolution = ConflictResolutionStrategy.MERGE_SIMILAR_OR_CONTAINED
```
This can be removed we can set None for ConflictResolutionStrategy.NONE case place as below
```
if (not conflict_resolution or conflict_resolution == conflict_resolution.NONE):
            return analyzer_results
```
If its different from what I can understand please elaborate little more. 

-> Although in test cases result looks similar for the options ```ConflictResolutionStrategy.MERGE_SIMILAR_OR_CONTAINED``` & ```ConflictResolutionStrategy.NONE```. But, for multiple entities result will differ. I think removing ```ConflictResolutionStrategy.NONE``` option itself is not best. 

Please let me know your thoughts on it.

presidio-anonymizer/tests/test_anonymizer_engine.py
omri374: Thanks!

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
VMD7: Hi @omri374 
Any comments on above approach.

omri374: Hi @VMD7, apologies for the delay. I'll make sure to review this in the next few days.

omri374: In high level, I see three scenarios:
1. What we have today (which I called `MERGE_SIMILAR_OR_CONTAINED`
2. The new strategy (what we call `REMOVE_INTERSECTIONS`)
3. Do nothing- no conflict handling at all. (what I think of when I see `None`). This might not make sense, and we can decide to skip it. 

My proposal is:
1. Put the existing logic under `MERGE_SIMILAR_OR_CONTAINED`)
2. Change the default to `MERGE_SIMILAR_OR_CONTAINED`)
3. Put the new logic under `REMOVE_INTERSECTIONS`)

I assume that `REMOVE_INTERSECTIONS` would need the current logic to run beforehand, right? If yes, then the if/else should represent this.

Hope this makes sense, apologies if I'm creating confusion here. We can also hop on another call if it helps.

omri374: I'm not sure what you mean by "If this way then this can be possible but it will unnecessary increase the time complexity for None case."

Maybe the position of my comment was misleading. I was putting the comment with regards to the logic in line 122 (`other_elements = analyzer_results.copy()`)

VMD7: Hi @omri374 
Removed the option of NONE from conflict resolution strategy and the default is  same as ```MERGE_SIMILAR_OR_CONTAINED```. Please let me know your thoughts on it.

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
```diff
@@ -28,6 +33,7 @@ def anonymize(
             text: str,
             analyzer_results: List[RecognizerResult],
             operators: Optional[Dict[str, OperatorConfig]] = None,
+            conflict_resolution: Optional[ConflictResolutionStrategy] = None
```
omri374, line 36:
```suggestion
            conflict_resolution: Optional[ConflictResolutionStrategy] = ConflictResolutionStrategy. MERGE_SIMILAR_OR_CONTAINED
```
Thanks. My only suggestion is this. If we keep it as None, it might seem as if there is no conflict resolution happening. WDYT?

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
omri374: Thanks @VMD7! I really appreciate all the work. I added one tiny comment and if you agree we can merge this.

presidio-anonymizer/presidio_anonymizer/anonymizer_engine.py
VMD7: Hi @omri374 
Thanks for the commit. I am extremely sorry for the misunderstanding. Actually, now I got the clear understanding what you were trying to say about third case. For now you can proceed with this or we can connect and discuss the third case as well, also on its documentation and some finding on llm integration.

omri374 (approved):

omri374:
Thank you @VMD7 for this contribution!

VMD7:
Thanks @omri374 for your support and guidance.

PR #1200, opened 2023-10-31
Put org in ignore as it has many FPs

omri374:
Revert the addition of `ORG` entities as they emit too many false positives.

**Note**: The docker based e2e tests are failing as they use the analyzer in PyPI and not the one based on this branch.


presidio-image-redactor/tests/integration/test_image_analyzer_engine_integration.py
```diff
@@ -5,7 +5,7 @@
 def test_given_image_then_text_entities_are_recognized_correctly(image_analyzer_engine):
     # Image with PII entities
     image = get_resource_image("ocr_test.png")
-    analyzer_results = image_analyzer_engine.analyze(image)
+    analyzer_results = image_analyzer_engine.analyze(image, score_threshold=0.4)
```
omri374, line 8: This is due to the change making case insensitive a default choice.

Issue #1254, opened 2024-01-12
Add keep operator as deanonymizer too


dinigo:
My use-case anonymises some data with encryption, and some other with destructive methods (replace, mask).

Then I run de de-anonymisation against all of those cases. Some (`MESSAGE_ID`) need to be decrypted, and some others (`CUSTOM_ID` or even `CUSTOM_DATE` which wasn't obfuscated) need to pass through.

```yaml
analyzer_config:
    entities: [ CUSTOM_ID, MESSAGE_ID, UUID, CUSTOM_DATE ]
    language: en
anonymize_config:
    # if not provided, replace replaces the value with the info_type, <CUSTOM_ID>
    CUSTOM_ID:
        type: replace
    UUID:
        type: kms_encrypt
        key_version_name: projects/my-secret-project/locations/europe-west2/keyRings/dlp/cryptoKeys/presidio-aes-key
    MESSAGE_ID:
        type: kms_encrypt
        key_version_name: projects/my-secret-project/locations/europe-west2/keyRings/dlp/cryptoKeys/presidio-aes-key
    # the rest of the detected fields (CUSTOM_DATE*) will be kept as they are
    DEFAULT:
        type: keep
deanonymize_config:
    # see here we use the `kms_decrypt` operator instead of the `kms_encrypt`
    UUID:
        type: kms_decrypt
        key_version_name: projects/my-secret-project/locations/europe-west2/keyRings/dlp/cryptoKeys/presidio-aes-key
    MESSAGE_ID:
        type: kms_decrypt
        key_version_name: projects/my-secret-project/locations/europe-west2/keyRings/dlp/cryptoKeys/presidio-aes-key
    DEFAULT:
        type: keep
```

However, as `keep` is only registered as anonymiser, I cannot use it. I just copy pasted it to be able to implement the usecase above.

I think this change/fixx might belong to the core, since the use-case of anonymising several and then retrieving only encrypted is a common one.

PR #1255, opened 2024-01-12
Add keep operator as deanonymizer

dinigo:
Implements the `keep` operator for deanonymisation too.

It keeps all the methods from the Keep class and extends only the ones that return the type.

Concrete classes MUST subclass `Operator` type, since operator discovery is done with `Operator.__subclasses__()`

This PR fixes issue #1254

## Checklist

### DIFF
--- a/presidio-anonymizer/presidio_anonymizer/operators/keep.py
+++ b/presidio-anonymizer/presidio_anonymizer/operators/keep.py
@@ -1,14 +1,15 @@
 """Keeps the PII text unmodified."""
+from abc import ABC
 from typing import Dict
 
 from presidio_anonymizer.operators import Operator, OperatorType
 
 
-class Keep(Operator):
-    """No-op anonymizer that keeps the PII text unmodified.
+class BaseKeep(Operator, ABC):
+    """Abstract base class for keep operators.
 
-    This is useful when you don't want to anonymize some types of PII,
-    but wants to keep track of it with the other PIIs.
+    Implements the common functionality of keep operators, leaving the operator
+    type to be implemented by concrete classes
     """
 
     def operate(self, text: str = None, params: Dict = None) -> str:
@@ -19,6 +20,14 @@ class Keep(Operator):
         """Keep does not require any paramters so no validation is needed."""
         pass
 
+
+class Keep(BaseKeep, Operator):
+    """No-op anonymizer that keeps the PII text unmodified.
+
+    This is useful when you don't want to anonymize some types of PII,
+    but wants to keep track of it with the other PIIs.
+    """
+
     def operator_name(self) -> str:
         """Return operator name."""
         return "keep"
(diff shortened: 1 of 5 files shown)


dinigo:
Hello, I'm just creating this PR in case you (de core dev team) find it interesting. If so I'll pay a visit to the testing and other features

dinigo:

presidio-anonymizer/presidio_anonymizer/operators/keep.py
```diff
@@ -1,14 +1,15 @@
 """Keeps the PII text unmodified."""
+from abc import ABC
 from typing import Dict
 
 from presidio_anonymizer.operators import Operator, OperatorType
 
 
-class Keep(Operator):
```
omri374, line 7: I would suggest to keep `Keep` as it might be used by some users already, and this would potentially not be backward compatible

presidio-anonymizer/presidio_anonymizer/operators/keep.py
```diff
@@ -23,6 +24,26 @@ def operator_name(self) -> str:
         """Return operator name."""
         return "keep"
```
omri374, line 25: Can we provide a different name to the keep deanonymizer? It might be confusing having two operators with the same name.

omri374 (changes requested):
Great addition! Left a couple of small comments.

presidio-anonymizer/presidio_anonymizer/operators/keep.py
dinigo: How about?

```suggestion
        return "deanonymize_keep"
```

presidio-anonymizer/presidio_anonymizer/operators/keep.py
dinigo: Sure! Amended, keeping `Keep` and fixed testing

presidio-anonymizer/presidio_anonymizer/operators/keep.py
dinigo: Done! and fixed tests to check for `deanonymize_keep`. If you where thinking of other name just tell. It's a quick change

presidio-anonymizer/tests/integration/test_deanonymize_engine.py
```diff
@@ -20,7 +20,7 @@ def test_given_operator_decrypt_with_valid_params_then_decrypt_text_successfully
     decryption = engine.deanonymize(
         text,
         encryption_results,
-        {"DEFAULT": OperatorConfig(Decrypt.NAME, {"key": b'WmZq4t7w!z%C&F)J'})},
+        {"DEFAULT": OperatorConfig(Decrypt.NAME, {"key": b"WmZq4t7w!z%C&F)J"})},
```
dinigo, line 23: This '' -> "" was changed by the black pre-commit hook

omri374 (approved):

PR #1257, opened 2024-01-21
Add missing dependencies for image-redactor

omri374:
Added two missing dependencies in image-redactor:
- Azure AI form recognizer
- opencv-python (cv2)


SharonHart (approved):

.pipelines/templates/publish-to-pypi.yml
```diff
@@ -13,6 +13,16 @@ steps:
         pip install wheel
         python setup.py bdist_wheel
       workingDirectory: ${{ parameters.WORKING_FOLDER }}
+  - task: Bash@3
+    displayName: 'Install locally and test package''
```
SharonHart, line 17: Isnt it too late at this stage?

SharonHart (approved):

navalev (approved):

.pipelines/templates/publish-to-pypi.yml
omri374: Yes I think so too but don't have a good alternative. Should we rerun all unit tests during CI?

SharonHart: In the e2e suite we install and test using the package, currently without the image redactor. It still doesnt cover everything since you can test a scenario that doesnt use a certian dependency

https://github.com/microsoft/presidio/blob/main/e2e-tests%2Frequirements.txt

omri374: Discussed offline, merging for now to fix the issue, will find a solution to the package testing later.

PR #1280, opened 2024-02-07
initial changes without tests

RoeyBC:
## Change Description

Describe your changes

## Issue reference

This PR fixes issue #XX

## Checklist

- [V] I have reviewed the [contribution guidelines](https://github.com/microsoft/presidio/blob/main/CONTRIBUTING.md)
- [V] I have signed the CLA (if required)
- [V] All unit tests and lint checks pass locally

### DIFF
--- a/presidio-anonymizer/presidio_anonymizer/operators/custom.py
+++ b/presidio-anonymizer/presidio_anonymizer/operators/custom.py
@@ -1,5 +1,6 @@
 """Replaces the PII text with function result."""
 from typing import Dict
+from inspect import signature
 
 from presidio_anonymizer.operators import Operator, OperatorType
 from presidio_anonymizer.entities import InvalidParamException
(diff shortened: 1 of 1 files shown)


presidio-anonymizer/presidio_anonymizer/operators/custom.py
```diff
@@ -36,3 +35,10 @@ def operator_name(self) -> str:
     def operator_type(self) -> OperatorType:
         """Return operator type."""
         return OperatorType.Anonymize
+
+    def _call_lambda(self, text: str = None, params: Dict = None):
+        new_val = params.get(self.LAMBDA)
+        sig = signature(new_val)
+        if len(sig.parameters) == 1:
+            return new_val(text)
+        return new_val(text, params.get("entity_type"))
```
omri374, line 44: Are there cases where we'd want more than just the entity type? It could be worth thinking about other use cases for the custom lambda where the user might need additional metadata.

RoeyBC: How about start location and score? I think it could be a dictionary in case we want to add more stuff in the future.

omri374: Let's discuss offline. I was thinking that maybe a better solution would be to create a new operator. Wdyt?

RoeyBC: Agree! that'll be cleaner :D

PR #1288, opened 2024-02-09
predefined pattern recognizer : IN_VEHICLE_REGISTRATION

devopam:
## Change Description
Addition of a predefined recognizer : IN_VEHICLE_REGISTRATION

Describe your changes
Pattern, Context and Checksum based recognizer for India RTO issued vehicle registration number

## Issue reference

This PR fixes issue #XX

## Checklist


devopam:
Hi @omri374 , please have a look when feasible and let me know your review comments

omri374:
Thanks! I'll review it soon.

presidio-analyzer/presidio_analyzer/predefined_recognizers/in_vehicle_registration_recognizer.py
```diff
@@ -0,0 +1,1582 @@
...
+        # print('Sanitized value:' + sanitized_value)
+        return self.__check_vehicle_registration(sanitized_value)
+
+    def __check_vehicle_registration(self, sanitized_value: str) -> bool:
+        # print('check function called')
+        is_valid_registration = None
+        # logic here
+        an_list = ["01"]
+        ap_list = ["39", "40"]
+        ar_list = [
```
omri374, line 108: Is it possible to create those lists dynamically, and then remove the values that shouldn't be there? It makes the code super long and difficult to go over.

presidio-analyzer/presidio_analyzer/predefined_recognizers/in_vehicle_registration_recognizer.py
```diff
@@ -0,0 +1,1582 @@
...
+        two_factor_registration_prefix.extend(union_territories)
+        two_factor_registration_prefix.extend(states)
+        two_factor_registration_prefix.extend(old_states)
+        two_factor_registration_prefix.extend(old_union_territories)
+        two_factor_registration_prefix.extend(non_standard_state_or_ut)
+        first_two_char = sanitized_value[:2].upper()
+        dist_code: str = ""
+        diplomatic_vehicle_codes = ["CC", "CD", "UN"]
+
+        match first_two_char:
```
omri374, line 1558: This is super cool but unfortunately not supported in python < 3.10, while Presidio still supports 3.8

omri374 (commented):
Thanks! great addition. Left a few comments

presidio-analyzer/presidio_analyzer/predefined_recognizers/in_vehicle_registration_recognizer.py
devopam: oversight error, thanks for pointing it. reverted to compatible syntax

presidio-analyzer/presidio_analyzer/predefined_recognizers/in_vehicle_registration_recognizer.py
devopam: Hi @omri374 ,
Yes, the code was becoming unreadable. So , I moved the metadata to analyzer_utils.py. I deliberated on the design before coding it so and went with generic code + hardcoded data approach instead of the other way round to avoid creating too many exceptions making it unmaintanable in the longer run imho. The root cause is lack of regulation in the defined codes resulting in _gaps_ in the series for most of the cases. Plus, there is always a possibility that an entry is **delisted** in future. We will have a tough time incorporating those changes later on. Hence , this approach. Please let me know your opinion on the same.

devopam:
Hi @omri374 , 
Please review when you get time to do so. I have incorporated changes to reflect on your review feedback to the best intent.
regards
Devopam

presidio-analyzer/presidio_analyzer/analyzer_utils.py
```diff
@@ -73,3 +73,213 @@ def is_verhoeff_number(input_number: int):
         for i in range(len(inverted_number)):
             c = __d__[c][__p__[i % 8][inverted_number[i]]]
         return __inv__[c] == 0
+
+    # fmt: off
+    in_vehicle_foreign_mission_codes = [
```
omri374, line 78: Are these generic or referring just to the in_vehicle_registration? If it's just for this, can we please move it to the recognizer file?

presidio-analyzer/presidio_analyzer/predefined_recognizers/in_vehicle_registration_recognizer.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def validate_result(self, pattern_text: str) -> bool:
+        """Determine absolute value based on calculation."""
+        sanitized_value = Utils.sanitize_value(pattern_text, self.replacement_pairs)
+        # print('Sanitized value:' + sanitized_value)
+        return self.__check_vehicle_registration(sanitized_value)
+
+    def __check_vehicle_registration(self, sanitized_value: str) -> bool:
+        # print('check function called')
+        is_valid_registration = None
```
omri374, line 104:
```suggestion
        is_valid_registration = False
```

presidio-analyzer/presidio_analyzer/predefined_recognizers/in_vehicle_registration_recognizer.py
```diff
@@ -0,0 +1,181 @@
...
+        if first_two_char in two_factor_registration_prefix:
+            if sanitized_value[2].isdigit():
+                if sanitized_value[3].isdigit():
+                    dist_code = sanitized_value[2:4]
+                else:
+                    dist_code = sanitized_value[2:3]
+
+                registration_digits = sanitized_value[-4:]
+                if registration_digits.isnumeric():
+                    if 0 < int(registration_digits) <= 9999:
```
omri374, line 164: Can we check that this is an int before casting? There could be a false positive which would fail the entire process

presidio-analyzer/presidio_analyzer/predefined_recognizers/in_vehicle_registration_recognizer.py
```diff
@@ -0,0 +1,181 @@
...
+        two_factor_registration_prefix.extend(Utils.in_union_territories)
+        two_factor_registration_prefix.extend(Utils.in_states)
+        two_factor_registration_prefix.extend(Utils.in_old_states)
+        two_factor_registration_prefix.extend(Utils.in_old_union_territories)
+        two_factor_registration_prefix.extend(Utils.in_non_standard_state_or_ut)
+        first_two_char = sanitized_value[:2].upper()
+        dist_code: str = ""
+
+        if first_two_char in two_factor_registration_prefix:
+            if sanitized_value[2].isdigit():
```
omri374, line 156: Please check for `len(sanitized_value)`

presidio-analyzer/presidio_analyzer/analyzer_utils.py
devopam: these are indeed specific to Vehicle Registration only and I don't see any future use elsewhere. Moved the logic back to recognizer classfile as advised.

presidio-analyzer/presidio_analyzer/predefined_recognizers/in_vehicle_registration_recognizer.py
devopam: line 163 is checking for that i believe. please have a look once and let me know.

presidio-analyzer/presidio_analyzer/predefined_recognizers/in_vehicle_registration_recognizer.py
devopam: good catch. i missed surely. and this would have caused a failure for invalid values passed. fixed at top level.

presidio-analyzer/presidio_analyzer/predefined_recognizers/in_vehicle_registration_recognizer.py
devopam: deliberately let out . comment added also. this is to ensure that validate_result gives out NULL output for cases that are not falling in the two defined deterministic pattern checks. This is so , because validate_result is designed to give 0/1 as min and max probabilities. I am using validate_result for only two patterns where metadata is available to increase the probability of search while other patterns stay as-is.

devopam:
hi @omri374 ,
incorporated two comments and responded to another two. Please have a look when you get time , and let me know if this makes sense. And thank you for taking time out to go through it thoroughly. Much appreciated.

presidio-analyzer/presidio_analyzer/predefined_recognizers/in_vehicle_registration_recognizer.py
```diff
@@ -0,0 +1,395 @@
...
+            "India Vehicle Registration",
+            r"\b(?!00)\d{2}(A|B|C|D|E|F|H|K|P|R|X)\d{6}[A-Z]{1}\b",
+            0.85,
+        ),
+    ]
+
+    CONTEXT = ["RTO", "vehicle", "plate", "registration"]
+
+    # fmt: off
+    in_vehicle_foreign_mission_codes = [
```
omri374, line 75: Thinking about performance, perhaps it would be better to have those as sets instead of lists. Sets lookup is O(1) while for lists it's O(n)

presidio-analyzer/presidio_analyzer/predefined_recognizers/in_vehicle_registration_recognizer.py
```diff
@@ -0,0 +1,395 @@
...
+        sanitized_value = Utils.sanitize_value(pattern_text, self.replacement_pairs)
+        # print('Sanitized value:' + sanitized_value)
+        return self.__check_vehicle_registration(sanitized_value)
+
+    def __check_vehicle_registration(self, sanitized_value: str) -> bool:
+        # print('check function called')
+        is_valid_registration = None  # deliberately not typecasted or set to bool False
+        # logic here
+        if len(sanitized_value) >= 8:
+            state_rto_district_map = {
```
omri374, line 311: Can this dict be created outside the method? Otherwise it would be created on every call.

presidio-analyzer/presidio_analyzer/predefined_recognizers/in_vehicle_registration_recognizer.py
```diff
@@ -0,0 +1,395 @@
...
+
+    def validate_result(self, pattern_text: str) -> bool:
+        """Determine absolute value based on calculation."""
+        sanitized_value = Utils.sanitize_value(pattern_text, self.replacement_pairs)
+        # print('Sanitized value:' + sanitized_value)
+        return self.__check_vehicle_registration(sanitized_value)
+
+    def __check_vehicle_registration(self, sanitized_value: str) -> bool:
+        # print('check function called')
+        is_valid_registration = None  # deliberately not typecasted or set to bool False
```
omri374, line 308: Could you please explain the comment? If the method signature says it returns bool, then None should not be one of the options.

presidio-analyzer/presidio_analyzer/predefined_recognizers/in_vehicle_registration_recognizer.py
```diff
@@ -0,0 +1,395 @@
...
+                "RJ": self.in_vehicle_dist_rj,
+                "SK": self.in_vehicle_dist_sk,
+                "TN": self.in_vehicle_dist_tn,
+                "TR": self.in_vehicle_dist_tr,
+                "TS": self.in_vehicle_dist_ts,
+                "UK": self.in_vehicle_dist_uk,
+                "UP": self.in_vehicle_dist_up,
+                "WB": self.in_vehicle_dist_wb,
+            }
+            two_factor_registration_prefix = []
```
omri374, line 351: This list could also be crated outside and reused

omri374 (commented):
Thanks @devopam! I've left a few comments mainly around performance.

presidio-analyzer/presidio_analyzer/predefined_recognizers/in_vehicle_registration_recognizer.py
devopam: incorporated. very good suggestion.

presidio-analyzer/presidio_analyzer/predefined_recognizers/in_vehicle_registration_recognizer.py
devopam: incorporated. very good suggestion. didn't think of repeat calls after class invocation.

presidio-analyzer/presidio_analyzer/predefined_recognizers/in_vehicle_registration_recognizer.py
devopam: agreed. it will save time especially for lists > 50 records for sure despite being ordered in values. incorporated.

presidio-analyzer/tests/test_analyzer_utils.py
```diff
@@ -6,8 +6,17 @@
     ["abCba", False, True],
     ["ABBA", False, True],
     ["aBba", True, True],
+    ["NotAPalindrome", True, False],
 ]
 
+in_vehicle_metadata_test_set = [
```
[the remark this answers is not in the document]
devopam: big oops , forgot to delete it from here while cleaning up the calling function.

presidio-analyzer/presidio_analyzer/predefined_recognizers/in_vehicle_registration_recognizer.py
devopam: hi @omri374 , my requirement is that the validate function should kick in for checking 2 of the 9 patterns defined. Now, if it returns a bool : false, then the score is defaulted to 0 (min score) instead of my intended pattern based score as per the logic for this utility function. Hence, I am returning None for the rest of the 7 cases so that I can pick up the pattern based score, instead of 0. Please let me know if there is a better way to implement the same.

presidio-analyzer/presidio_analyzer/predefined_recognizers/in_vehicle_registration_recognizer.py
omri374: Right, sorry I missed that

presidio-analyzer/presidio_analyzer/predefined_recognizers/in_vehicle_registration_recognizer.py
omri374: I see. Ok, let's leave as is. One approach could have been to split to two different recognizers but I think it's probably over engineering.

omri374 (approved):
Thanks for incorporating all the changes! Approved.

Issue #1301, opened 2024-02-16
Anonymizer does not work if not separated by spaces


Siddhartha90:
**Describe the bug**

`test data 630-596-1111`

redacts ok to give back `test data`

but `test data630-596-1111`

does not redact the phone number.

**To Reproduce**

```

text_to_anonymize="test data630-596-1111"
try:
    nlp = spacy.load("en_core_web_lg")
except OSError:
    # Model not found, download it
    spacy.cli.download("en_core_web_lg")
    nlp = spacy.load("en_core_web_lg")

analyzer = AnalyzerEngine()
anonymizer = AnonymizerEngine()

analyzer_results = analyzer.analyze(
    text=text_to_anonymize,
    entities=["PHONE_NUMBER", "EMAIL_ADDRESS", "URL", "LOCATION"],
    language="en",
)
scrubbed_text = anonymizer.anonymize(
    text=text_to_anonymize,
    analyzer_results=analyzer_results,
    operators={"DEFAULT": OperatorConfig("redact", {})},
).text

print(scrubbed_text)

```

**Expected behavior**
Phone number gets scrubbed.

I'm on:

**presidio-analyzer==2.2.351
presidio-anonymizer==2.2.351
spacy==3.7.2
spacy-legacy==3.0.12
spacy-loggers==1.0.5
**

omri374:
Presidio uses the [phonenumbers](https://github.com/daviddrysdale/python-phonenumbers) package to detect phone numbers, and it seems that the package's logic doesn't detect phone numbers without spaces before and after. As @VMD7 suggested on the other issue, you can try using a NER model which was trained on phone numbers and see if you get better results. Alternatively, you can create a custom pattern recognizer to detect phone numbers with no spaces around them.

Siddhartha90:
Thanks @omri374, does your PR above with the leniency param help here as well? I'm not sure if that's a parameter exposed via the library

omri374:
Yes it should, but it's not released yet. Once it is (or if you clone the repo), you should be able to adjust it, for example:
```py
from presidio_analyzer import AnalyzerEngine, RecognizerRegistry

registry = RecognizerRegistry()
registry.load_predefined_recognizers()

phone_recognizer = [rec for rec in registry.recognizers if rec.name == "PhoneRecognizer"][0]
phone_recognizer.leniency = 0 # update phone number leniency

analyzer_engine = AnalyzerEngine(registry=registry)

[...]

PR #1311, opened 2024-02-23
Addition of leniency parameter in predefined PhoneRecognizer

VMD7:
Addition of `leniency` parameter in predefined `PhoneRecognizer`
## Change Description
-> The `PhoneRecognizer` failed to detect phone numbers when they appeared without any space between text and number, as seen in cases like `testdata630-596-1111`.

-> Internally, the `PhoneRecognizer` utilizes the [phonenumbers](https://github.com/daviddrysdale/python-phonenumbers) library for phone number detection. This library offers a parameter known as `leniency`, which determines the flexibility of the phone number format. The leniency parameter ranges from `0 to 3`, with higher values indicating more strict phone number format match.

-> Initially, the leniency factor in `PhoneRecognizer` was set to `1`, causing it to overlook phone numbers lacking spaces between text and digits. To address this issue, we've enhanced PhoneRecognizer to allow users to adjust the leniency parameter according to their requirements, with a default value of `1`.

-> This enhancement resolves various phone number format issues and empowers users to tailor its behavior to their specific needs.

## Issue reference

This PR fixes issue #1301 

## Checklist

### DIFF
--- a/presidio-analyzer/presidio_analyzer/predefined_recognizers/phone_recognizer.py
+++ b/presidio-analyzer/presidio_analyzer/predefined_recognizers/phone_recognizer.py
@@ -18,6 +18,8 @@ class PhoneRecognizer(LocalRecognizer):
     :param context: Base context words for enhancing the assurance scores.
     :param supported_language: Language this recognizer supports
     :param supported_regions: The regions for phone number matching and validation
+    :param leniency: The strictness level of phone number formats.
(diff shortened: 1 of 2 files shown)


omri374 (approved):
Thanks!

omri374:
Could be relevant to revisit #772. The phone number recognizer makes some assumptions, such as the regions it looks for phone numbers for (as it is slower to search all of the regions) and other hyperparameters such as leniency.

VMD7:
Yes sure, will go through it.

Issue #1316, opened 2024-02-28
presidio-structured misidentifies email as URL


ardband:
Presidio-structured incorrectly identifies an email address as a URL within the extracted entities. This can be observed in the following example output:

StructuredAnalysis(entity_mapping={'name': 'PERSON', 'email': 'URL', 'city': 'LOCATION', 'state': 'LOCATION'})

the value in the "email" column ("john.doe@example.com") is mistakenly identified as a URL ("URL") instead of an email address ("EMAIL") during entity extraction.

miltonsim:
I've also encountered this issue.

The issue mainly stems from the `_find_most_common_entity()` method where email addresses in `test_structured.csv` are being incorrectly identified as URLs, albeit with low confidence. It prioritises the entity with the highest count.

Observed behavior: 
- Entity Count: `{'URL': 6, 'EMAIL_ADDRESS': 3}` 
- Confidence Scores: `{'EMAIL_ADDRESS': [1.0, 1.0, 1.0], 'URL': [0.5, 0.5, 0.5, 0.5, 0.5, 0.5]}`

The emails are accurately recognised but are outnumbered by the URL identifications due to their higher frequency, despite lower confidence levels.

I would like to suggest two potential improvements: 
1. **Adapting `_find_most_common_entity()` to Consider Confidence Scores:**  It might be beneficial to adjust the method to account for the actual confidence scores provided by the recognizer results.
2. **Enhancing the URL Recognizer:**  Improving the recognizer's ability to differentiate between URLs and email addresses could help reduce this type of misidentification

I'm keen to contribute to making these improvements and would love to work on refining the logic. Any thoughts or feedback on these suggestions would be greatly appreciated!

omri374:
Thanks for the feedback! the URL recognizer detects parts of emails as well (e.g. microsoft.com is a url inside john.doe@microsoft.com), which makes it detect more URLs than emails. 

I think that a good way forward here would be to allow the user to decide on a strategy for the entity selected. In some cases, we would want the entity with the majority of cases, in others we'd like the one that has the highest confidence, and in others we might want a mix of the two (e.g. most common entity, if confidence > 0.5)

A quick fix could be to update the structured analysis once finalized, in case the column's name is "email" but the detection is actually "URL". 

If you're interested in creating a PR, I'd be happy to review it and discuss.

PR #1319, opened 2024-02-29
feat: Implement user-defined entity selection strategies in Presidio Structured

miltonsim:
## Change Description

I implemented an additional `selection_strategy` option in the `generate_analysis()` function, allowing users to define their preferred strategy for entity selection. Previously, the function only enabled selection based on the most common entity, which could result in scenarios where a small number of high-confidence entities were overshadowed by a larger number of low-confidence entities. Now, users have the flexibility to choose between three strategies: _most common_, _highest_confidence_ or _mixed_.

I would love to hear to hear feedback and suggestions!

## Issue reference

This PR fixes issue #1316 

## Checklist

### DIFF
--- a/presidio-structured/presidio_structured/analysis_builder.py
+++ b/presidio-structured/presidio_structured/analysis_builder.py
@@ -159,6 +159,7 @@ class PandasAnalysisBuilder(TabularAnalysisBuilder):
         df: DataFrame,
         n: Optional[int] = None,
         language: str = "en",
+        selection_strategy: str = "most_common"
     ) -> StructuredAnalysis:
         """
         Generate a configuration from the given tabular data.
@@ -179,7 +180,11 @@ class PandasAnalysisBuilder(TabularAnalysisBuilder):
 
         df = df.sample(n, random_state=123)
 
-        key_recognizer_result_map = self._generate_key_rec_results_map(df, language)
+        key_recognizer_result_map = self._generate_key_rec_results_map(
+            df,
+            language,
+            selection_strategy
+        )
 
         key_entity_map = {
             key: result.entity_type
@@ -190,7 +195,7 @@ class PandasAnalysisBuilder(TabularAnalysisBuilder):
         return StructuredAnalysis(entity_mapping=key_entity_map)
 
     def _generate_key_rec_results_map(
-        self, df: DataFrame, language: str
+        self, df: DataFrame, language: str, selection_strategy: str = "most_common"
     ) -> Dict[str, RecognizerResult]:
         """
         Find the most common entity in a dataframe column.
@@ -204,8 +209,9 @@ class PandasAnalysisBuilder(TabularAnalysisBuilder):
         column_analyzer_results_map = self._batch_analyze_df(df, language)
         key_recognizer_result_map = {}
         for column, analyzer_result in column_analyzer_results_map.items():
-            key_recognizer_result_map[column] = self._find_most_common_entity(
-                analyzer_result
+            key_recognizer_result_map[column] = self._find_entity_based_on_strategy(
+                analyzer_result,
+                selection_strategy
             )
         return key_recognizer_result_map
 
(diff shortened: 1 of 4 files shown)


presidio-structured/presidio_structured/analysis_builder.py
```diff
@@ -159,6 +159,7 @@ def generate_analysis(
         df: DataFrame,
         n: Optional[int] = None,
         language: str = "en",
+        selection_strategy: str = "most_common"
```
omri374, line 162: Please add the selection strategies to the docstring (you have that in the inner function, could be copied here as well)

presidio-structured/presidio_structured/analysis_builder.py
```diff
@@ -230,44 +236,133 @@ def _batch_analyze_df(
...
+        :param flat_results: A list of tuples containing index and RecognizerResult
+        objects from the flattened analysis results.
+        :return: A RecognizerResult object selected based on the mixed strategy.
+        """
+        score_aggregator = self._aggregate_scores(flat_results)
+
+        # Check if the highest score is greater than 0.5 and select accordingly
+        highest_score = max(max(scores) for scores in score_aggregator.values()
+                            if scores)
+        if highest_score > 0.5:
```
omri374, line 336: Can the 0.5 be a configurable parameter?

omri374:
Thanks @mitonsim! Left a few minor comments

omri374:
Please also add a note to the docs about the strategies: https://github.com/microsoft/presidio/blob/main/docs/structured/index.md

presidio-structured/presidio_structured/analysis_builder.py
```diff
@@ -230,44 +236,133 @@ def _batch_analyze_df(
...
-        flat_results = [
-            (cell_idx, res)
-            for cell_idx, cell_results in enumerate(analyzer_results)
-            for res in cell_results
-        ]
+        flat_results = self._flatten_results(analyzer_results)
 
-        # Count the occurrences of each entity type in different cells
-        type_counter = Counter(res.entity_type for cell_idx, res in flat_results)
+        # Select the entity based on the desired strategy
```
omri374, line 260: Please add a validation to the strategy (raise ValueError if the strategy is not supported)

miltonsim:
@omri374, thanks for your feedback! I've incorporated the requested changes.

However, I encountered an issue where my newly added test cases failed in the Azure Pipeline. The assertion assert `key_recognizer_result_map["street"].entity_type == "NON_PII"` is not passing because the 'street' column is incorrectly identified as a location, whereas, in my local environment, it's correctly identified as NON_PII. 

Could you shed some light on this discrepancy and how I can resolve it?

omri374:
@miltonsim, I'm not sure where the discrepancy is coming from. It could be the seed used for sampling, or a different version of the spaCy model. To reduce noise in unit tests, perhaps it would be easier to choose an entity detected using a simpler logic? like a phone number, email, credit card?

omri374:
Hi @miltonsim would you be interested in continuing to work on this? I would suggest to replace the non-deterministic entities with others to make sure tests pass on all environments.

miltonsim:
@omri374 Apologies for the delay. I'll provide an update by 14 March (Tues)

omri374:
Thanks!

miltonsim:
@omri374 Thank you for waiting. I figured that removing the street column might be the best solution. Please help me run the pipeline to check if it passes!

presidio-structured/presidio_structured/analysis_builder.py
```diff
@@ -154,18 +154,26 @@ class TabularAnalysisBuilder(AnalysisBuilder):
 class PandasAnalysisBuilder(TabularAnalysisBuilder):
     """Concrete configuration generator for tabular data."""
 
+    entity_selection_strategies = {"highest_confidence", "mixed", "most_common"}
+
     def generate_analysis(
```
Jakob-98, line 159: Overall, PR looks good to me, nice addition!

One thing which might be nice to address (although you didn't introduce this issue) is that we are redefining the function definition of generate_analysis from the AnalysisBuilder(ABC)

Perhaps we can address it in this PR as well, and redefine the abstract method in the base class to: 

`
    @abstractmethod
    def generate_analysis(
        self,
        data: Union[Dict, DataFrame],
        *args,
        **kwargs
    ) -> StructuredAnalysis:
`

(.. and then rename df to data to be consistent)

So we don't override the function definition in the concrete implementations of the class.

Jakob-98 (approved):

presidio-structured/presidio_structured/analysis_builder.py
omri374: I suggest to handle this in a subsequent PR if that's ok with everyone.

omri374 (approved):

elbud:
Could this update be pushed to PyPi? There is only the original release available at the moment. https://pypi.org/project/presidio-structured/#history

omri374:
Hi @elbud, we plan to have a new release in the next couple of weeks.

PR #1333, opened 2024-03-15
Fix bug where "bank" and "check" wouldn't work for `US_BANK_NUMBER` recognizer

usr-ein:
Those two keywords in US_BANK_NUMBER wouldn't bump the entity's score because of Python multiline concatenation.

## Change Description

- Added a comma to prevent concatenation
- Move the keyword _below_ the comment to avoid such mistakes in the future

Additionally, I checked every other recognizer, but it didn't seem to occur elsewhere.
Here was my `grep` to look for it:

```
rg --multiline-dotall -U 'CONTEXT.*[^"]+"\n' presidio-analyzer/presidio_analyzer/predefined_recognizers/*.py
```

## Issue reference

~~This PR fixes issue #XX~~ No issue made. The fix is so simple I didn't thing this was needed.

## Checklist


omri374 (approved):
Thanks!

PR #1337, opened 2024-03-20
load analyzer configuration from file

RoeyBC:
## load analyzer configuration from file

giving the option to load analyzer from configuration file. It'll enable loading of:
1. NLP Engine
2. Reconizers
3. Supported languages

## Issue reference

This PR fixes issue #1338

## Checklist


presidio-analyzer/presidio_analyzer/analyzer_loader.py
```diff
@@ -0,0 +1,72 @@
...
+    def load(conf_file: str) -> AnalyzerEngine:
+        """
+        loads Presidio Analyzer from yaml configuration file.
+
+        :param conf_file: yaml file containing definitions
+        :return: analyzer engine initialized with yaml configuration
+        """
+
+        if not os.path.exists(conf_file):
+            return AnalyzerEngine()
```
omri374, line 30: Consider either adding a warning or raising an exception.

presidio-analyzer/presidio_analyzer/analyzer_loader.py
```diff
@@ -0,0 +1,72 @@
...
+
+class PresidioAnalyzerLoader:
+    """
+    Utility function for loading Presidio Analyzer.
+
+    Use this class to load presidio analyzer engine from a yaml file
+    """
+
+    @staticmethod
+    def load(conf_file: str) -> AnalyzerEngine:
```
omri374, line 21: Not that there's anything wrong with this API, but we should consider having similar APIs across the package. in `NlpEngineProvider` the API is `NlpEngineProvider(conf_file).create_engine()`. Can we use something similar? I would also consider having the same name (`Provider`).

presidio-analyzer/conf/analyzer.yaml
```diff
@@ -0,0 +1,49 @@
+recognizer_registry:
+  recognizers: 
+    - 
+      name: CreditCardRecognizer
+      supported_language: 
+      - en
+      - es
```
omri374, line 7: A recognizer can only support one language. If we want `CreditCardRecognizer` to support multiple, we need to have multiple instances of it. The reason is that context words differ between languages.

presidio-analyzer/conf/analyzer.yaml
```diff
@@ -0,0 +1,49 @@
+recognizer_registry:
+  recognizers:
```
omri374, line 2: How/can we have the engine load the predefined recognizers (call `RecognizerRegistry.load_predefined_recognizers()`)? How do we allow the user to provide their own recognizers vs. using the predefined ones or maybe a combination of the two?

presidio-analyzer/presidio_analyzer/analyzer_loader.py
```diff
@@ -0,0 +1,72 @@
...
+            nlp_configuration = configuration["nlp_engine_provider"]["nlp_configuration"]
+            provider = NlpEngineProvider(nlp_configuration=nlp_configuration)
+            nlp_engine = provider.create_engine()
+
+        if "supported_languages" in configuration:
+            supported_languages = configuration["supported_languages"]
+
+        if "recognizer_registry" in configuration and "recognizers" in configuration["recognizer_registry"]:
+            recognizers = configuration["recognizer_registry"]["recognizers"]
+            recognizer_instances = [getattr(presidio_analyzer.predefined_recognizers, PresidioAnalyzerLoader._get_name(recognizer), None)(supported_language=language)
```
omri374, line 49: Consider splitting the list comprehension to for-loops (e.g. one list comprehension within a for-loop) to reduce the code complexity.

presidio-analyzer/conf/analyzer.yaml
omri374: If we decide to have an opt-in approach, let's create a file with all existing recognizers (or at least all the global ones)

presidio-analyzer/conf/analyzer.yaml
```diff
@@ -0,0 +1,49 @@
...
+      supported_language: 
+      - en
+      - es
+
+    - ItFiscalCodeRecognizer
+
+supported_languages: 
+  - en
+  - es
+
```
omri374, line 14: Please add a `default_score_threshold` param

presidio-analyzer/presidio_analyzer/analyzer_loader.py
```diff
@@ -0,0 +1,72 @@
+import yaml
+import os
+import logging
+from typing import List
+
+from presidio_analyzer import AnalyzerEngine, RecognizerRegistry
+from presidio_analyzer.nlp_engine import NlpEngineProvider
+import presidio_analyzer.recognizer_registry
```
omri374, line 8: How is this import used? The two objects (AnalyzerEngine, RecognizerRegistry) are already loaded in line 6.

presidio-analyzer/conf/analyzer.yaml
omri374: If the intention was to allow multiple CreditCardRecognizers, I would suggest to split it into two objects and provide different context words to each. WDYT?

omri374 (commented):
Thanks! This is a great addition which would allow easy configuration not just for the Docker based capabilities, but also for a better code-free configuration. Added a few comments, but happy to discuss each of those as most of them are mainly a matter of style rather than correctness.

omri374:
Fixes #1338

presidio-analyzer/conf/analyzer.yaml
RoeyBC: So in the current implementation, if you don't provide recognizer_registry in the file, the predefined recognizers will be loaded

presidio-analyzer/presidio_analyzer/analyzer_loader.py
RoeyBC: Yeah I like this approach, deal

presidio-analyzer/presidio_analyzer/analyzer_loader.py
RoeyBC: I used to need it when dynamically instantiating the recognizers, might not need it now, let me check :D

RoeyBC: Yeah no I still need it

presidio-analyzer/conf/analyzer.yaml
omri374: Makes sense, how would you suggest to document this? Maybe adding a flag is more explicit?

presidio-analyzer/presidio_analyzer/analyzer_loader.py
omri374: Me too, but less when I don't comprehend the list comprehension 😆

presidio-analyzer/conf/analyzer.yaml
RoeyBC: Currently all params are optional which is cool, but I get the point that it's unclear what recognizer is being chosen. 

Suggestion: 
recognizer_registry: predefined
and if nothing is given resolve to the default which is also predefined.

omri374: Will that support the use case of predefined + new recognizers defined in the conf file?

presidio-analyzer/conf/analyzer.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - 
+      name: CreditCardRecognizer
+      supported_language: es
+
+    - ItFiscalCodeRecognizer
+
+supported_languages: 
+  - en
+  - es
```
omri374, line 15: Please break down to multiple like the credit card one

presidio-analyzer/conf/analyzer.yaml
omri374: I have a suggestion: split to a list of predefined and a list of new. Anything on the predefined would look for the recognizer by name and the new would create new recognizers (like the existing example_recognizers file). We can initially put all the global or most common recognizers in the file to make sure everything is loaded. Wdyt?

presidio-analyzer/conf/analyzer.yaml
RoeyBC: hmmmm, are you sure? supported_languages is a list of strings that the engine receives, breaking it down will make it a list of "supported languages".

omri374: You're right, my bad. I got confused

presidio-analyzer/presidio_analyzer/analyzer_engine_provider.py
```diff
@@ -0,0 +1,132 @@
...
+        """
+        predefined =[]
+        custom =[]
+
+        predefined = [recognizer for recognizer in recognizers if isinstance(recognizer, str) or ("type" in recognizer and recognizer["type"] == "predefined")]
+        # all recognizers are custom by default, type: custom can be mentioned as well
+        custom = [recognizer for recognizer in recognizers if not isinstance(recognizer, str) and ("type" not in recognizer or recognizer["type"] == "custom")]
+        return predefined, custom
+
+    def _create_recognizers(self, recognizer: Dict) -> List[EntityRecognizer]:
```
omri374, line 79:
```suggestion
    def _create_custom_recognizers(self, recognizer: Dict) -> List[EntityRecognizer]:
```
nit: differentiate between the creation of predefined and custom

presidio-analyzer/conf/analyzer.yaml
```diff
@@ -0,0 +1,93 @@
+recognizer_registry:
+  recognizers: 
+    - name: CreditCardRecognizer
+      supported_languages: 
+      - en
```
omri374, line 5: This structure is different from the one we use for custom. Here it's `{"supported_languages": ["en", "es"]}` and there it's `{"supported_languages": {"language": "es", "context": [...]}`. Can we / should we have the same structure? I think it's also aligned with what we agreed on the call, as @SharonHart gave this example structure:
```json
[
{
  "name": "CreditCardRecognizer",
  "languages": [
    {"langauge": "es", "context": ["tarjeta","credito"]},
    {"langauge": "en", "context": ["credit"]}
    ],
	"type": "PREDEFINED"
},
  {
  "name": "CreditCardRecognizer",
    "patterns": ["/b/s*[1]/b"],
  "languages": [
    {"language": "es", "context": ["tarjeta","credito"]},
    {"language": "en", "context": ["credit"]}
    ]
    ,"type": "CUSTOM",
    "enabled": false
}
]
```

omri374 (commented):
Great work! Left lots of comments but it's definitely in the right direction. Feel free to reach out if you want to discuss any of those.

presidio-analyzer/conf/analyzer.yaml
```diff
@@ -0,0 +1,93 @@
+recognizer_registry:
+  recognizers: 
+    - name: CreditCardRecognizer
+      supported_languages:
```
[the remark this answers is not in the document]
RoeyBC: Yeah, I like this approach. I kinda ignored the regex flag because I feel it'd be best if we parse it though from a string because int doesn't really express anything. But it's better than nothing.

presidio-analyzer/conf/analyzer.yaml
RoeyBC: Oh yeah that's just by accident, forgot to change the structure of the predefined, thanks!

presidio-analyzer/presidio_analyzer/analyzer_engine_provider.py
```diff
@@ -0,0 +1,132 @@
...
+            supported_languages=supported_languages,
+            default_score_threshold=default_score_threshold
+        )
+
+        return analyzer
+    
+    def _is_enabled(self, recognizer: Dict) -> bool:
+        return not "enabled" in recognizer or recognizer["enabled"]
+
+    def _get_name(self, recognizer) -> str:
```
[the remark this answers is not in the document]
RoeyBC: This is going to be extracted to a recognizer class so might keep the name, let's see how it would look like in the new implementation.

presidio-analyzer/conf/analyzer.yaml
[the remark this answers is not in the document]
omri374: Yes I was wondering too about this one

presidio-analyzer/conf/analyzer.yaml
```diff
@@ -0,0 +1,93 @@
...
+        - name: "zip code (weak)"
+          regex: "(\\b\\d{5}(?:\\-\\d{4})?\\b)"
+          score: 0.01
+      supported_languages:
+        - language: es
+          context: [tarjeta,credito]
+        - language: en
+          context: [credit]
+      supported_entity: "ZIP"
+      deny_list: [Mr., Mrs., Ms., Miss, Dr., Prof.]
```
SharonHart, line 25: deny list fits the TitlesRecognizer and demonstrated there, can be dropped?

presidio-analyzer/conf/analyzer.yaml
```diff
@@ -0,0 +1,93 @@
...
+      patterns:
+        - name: "zip code (weak)"
+          regex: "(\\b\\d{5}(?:\\-\\d{4})?\\b)"
+          score: 0.01
+        - name: "zip code (weak)"
+          regex: "(\\b\\d{5}(?:\\-\\d{4})?\\b)"
+          score: 0.01
+      supported_languages:
+        - language: es
+          context: [tarjeta,credito]
```
SharonHart, line 21: context is for credit card

presidio-analyzer/conf/analyzer.yaml
SharonHart: following up on that, at some point we decided to move all the predefined recognizers here right?

presidio-analyzer/presidio_analyzer/analyzer_engine_provider.py
```diff
@@ -0,0 +1,132 @@
...
+            copied_recognizer = {k: v for k, v in recognizer.items() if k not in ["enabled", "type", "supported_languages"]}
+            copied_recognizer["supported_language"] = recognizer["supported_languages"]
+            if not isinstance(copied_recognizer["supported_language"], str):
+                copied_recognizer["supported_language"] = supported_language["language"]
+                copied_recognizer["context"] = supported_language["context"]
+            recognizers.append(PatternRecognizer.from_dict(copied_recognizer))
+
+        return recognizers
+
+    def _load_recognizer_registry(self) -> RecognizerRegistry:
```
SharonHart, line 96:
```suggestion
    def _load_recognizer_registry(self) -> Optional[RecognizerRegistry]:
```

presidio-analyzer/presidio_analyzer/analyzer_engine_provider.py
```diff
@@ -0,0 +1,132 @@
...
+                if self._is_enabled(recognizer):
+                    recognizer_instances.append(getattr(presidio_analyzer.predefined_recognizers, self._get_name(recognizer), None)(supported_language=language))
+
+        for recognizer in custom:
+            if self._is_enabled(recognizer):
+                recognizer_instances.append(self._create_recognizers(recognizer))
+
+        return RecognizerRegistry(recognizers=recognizer_instances)
+    
+    def _load_nlp_engine(self) -> NlpEngine:
```
SharonHart, line 124:
```suggestion
    def _load_nlp_engine(self) -> Optional[NlpEngine]:
```

presidio-analyzer/app.py
```diff
@@ -37,7 +38,7 @@ def __init__(self):
         self.logger.setLevel(os.environ.get("LOG_LEVEL", self.logger.level))
         self.app = Flask(__name__)
         self.logger.info("Starting analyzer engine")
-        self.engine = AnalyzerEngine()
```
SharonHart, line 40: Do you think using AnalyzerEngineProvider from within AnalyzerEngine instantiation will be less of breaking change for people using the python package? 
@omri374

presidio-analyzer/presidio_analyzer/analyzer_engine_provider.py
SharonHart, line 79: some of the methods may be static

presidio-analyzer/presidio_analyzer/analyzer_engine_provider.py
```diff
@@ -0,0 +1,132 @@
...
+                logger.warning(
+                "recognizer_registry contains unidentified value. Using default configuration for recognizer registry"
+                )
+            return None
+
+        recognizers=recognizer_registry["recognizers"]
+        recognizer_instances = []
+        predefined, custom = self._split_recognizers(recognizers)
+        for recognizer in predefined:
+            for language in self._get_languages(recognizer):
```
SharonHart, line 114: can be transformed into list comprehension with +=

presidio-analyzer/presidio_analyzer/analyzer_engine_provider.py
```diff
@@ -0,0 +1,132 @@
...
+        if isinstance(recognizer, str):
+            return ["en"]
+        return recognizer["supported_languages"]
+
+    def _split_recognizers(self, recognizers) -> Tuple[List[str],List[str]]:
+        """
+        splits the recognizer list to predefined and custom
+        """
+        predefined =[]
+        custom =[]
```
SharonHart, line 72: redundant lines

presidio-analyzer/presidio_analyzer/analyzer_engine_provider.py
```diff
@@ -0,0 +1,132 @@
...
+
+        predefined = [recognizer for recognizer in recognizers if isinstance(recognizer, str) or ("type" in recognizer and recognizer["type"] == "predefined")]
```
SharonHart, line 74: nit: some of the keys strings are repeating and can be constants in the class, helping in case they would change in the future.

presidio-analyzer/presidio_analyzer/analyzer_engine_provider.py
```diff
@@ -0,0 +1,132 @@
+import yaml
```
SharonHart, line 1: Not necessarily in the PR, but I would add two kind of tests - one for the unit suite covering this class, and one in the integration suite, reading different yamls from the filesystem, and seeing how they are processed by the class.

SharonHart (commented):
Overall looks great, left some minor comments

presidio-analyzer/app.py
omri374: Not sure I understand. How will the `AnalyzerEngineProvider` be used from within `AnalyzerEngine`?

omri374:
@GautierT this PR is still missing the changes to the Dockerfile, so the actual conf file would not be passed correctly until this is sorted out.

presidio-analyzer/conf/analyzer.yaml
omri374: @RoeyBC this still has the old structure, doesn't it?

presidio-analyzer/conf/analyzer.yaml
omri374: Agree, we can perhaps have two examples, one for deny-lists (titles) and one for zip code (regex)

presidio-analyzer/presidio_analyzer/recognizer_registry_provider.py
```diff
@@ -0,0 +1,156 @@
...
+import presidio_analyzer.recognizer_registry
+
+logger = logging.getLogger("presidio-analyzer")
+
+
+class RecognizerRegistryProvider:
+    """
+    Utility class for loading Recognizer Registry.
+
+    Use this class to load recognzer registry from a yaml file
```
omri374, line 19: please add the params to the docstring (conf_file, registry_configuration)

presidio-analyzer/presidio_analyzer/recognizer_registry_provider.py
```diff
@@ -0,0 +1,156 @@
...
+    """
+
+    def __init__(self, conf_file: Optional[Union[Path, str]] = None, registry_configuration: Optional[Dict] = None):
+        self.configuration = self._get_configuration(conf_file, registry_configuration)
```
omri374, line 23: nit: better be pass named parameters to avoid issues with renaming or changes to parameters in the future
```suggestion
        self.configuration = self._get_configuration(conf_file=conf_file, registry_configuration=registry_configuration)
```

presidio-analyzer/presidio_analyzer/recognizer_registry_provider.py
```diff
@@ -0,0 +1,156 @@
...
+        return configuration
+
+    def _get_configuration(self, conf_file: Union[Path, str], registry_configuration: Dict) -> Union[Dict[str, Any]]:
+        configuration = {}
+
+        if registry_configuration:
+            configuration = registry_configuration.copy()
+
+        if not conf_file:
+            configuration = self._add_missing_keys(configuration=configuration, conf_file=self._get_full_conf_path())
```
omri374, line 38: Can this raise an `IOError`? If yes, worth wrapping with try/except

presidio-analyzer/presidio_analyzer/recognizer_registry_provider.py
omri374, line 19:
```suggestion
    Use this class to load recognizer registry from a yaml file
```

presidio-analyzer/tests/test_analyzer_engine_provider.py
```diff
@@ -0,0 +1,71 @@
...
+
+from presidio_analyzer.nlp_engine.transformers_nlp_engine import TransformersNlpEngine
+
+def assert_default_configuration(engine: AnalyzerEngine):
+    assert engine.supported_languages == ["en"]
+    assert engine.registry.global_regex_flags == re.DOTALL | re.MULTILINE | re.IGNORECASE
+    assert engine.default_score_threshold == 0
+    assert len(engine.registry.recognizers) == 29
+    names = [recognizer.name for recognizer in engine.registry.recognizers]
+    assert set(names) == set(["UsBankRecognizer", "UsLicenseRecognizer", "UsItinRecognizer",
```
omri374, line 15: Recognizers change all the time, it's better not to have a static list of recognizers here

presidio-analyzer/presidio_analyzer/recognizer_registry_provider.py
```diff
@@ -0,0 +1,156 @@
...
+    Use this class to load recognzer registry from a yaml file
+    """
+
+    def __init__(self, conf_file: Optional[Union[Path, str]] = None, registry_configuration: Optional[Dict] = None):
+        self.configuration = self._get_configuration(conf_file, registry_configuration)
+        return
+    
+    def _add_missing_keys(self, configuration: Dict, conf_file: Union[Path, str]):
+        defaults = yaml.safe_load(open(conf_file))
+        configuration.update({k: v for k, v in defaults.items() if k not in list(configuration.keys())})
```
omri374, line 28: nice!

presidio-analyzer/presidio_analyzer/recognizer_registry_provider.py
```diff
@@ -0,0 +1,156 @@
...
+            "recognizers": None,
+            "global_regex_flags": None
+        }
+
+        for field in fields.keys():
+            if field not in self.configuration:
+                logger.warning(f'{field} not present in configuration, using default value instead: {self.default_values[field]}')
+            fields[field] = self.configuration.get(field, self.default_values[field])
+
+        self.supported_languages = fields["supported_languages"]
```
omri374, line 129: Does this have to be in self? If yes, please add `self.supported_languages=None` to the init

presidio-analyzer/presidio_analyzer/recognizer_registry_provider.py
```diff
@@ -0,0 +1,156 @@
...
+
+        recognizer_instances = []
+        predefined, custom = self._split_recognizers(fields["recognizers"])
+        for recognizer in predefined:
+            for language in self._get_recognizer_languages(recognizer):
+                if self._is_recognizer_enabled(recognizer):
+                    copied_recognizer = {k: v for k, v in self._get_recognizer_items(recognizer) if k not in ["enabled", "type", "supported_languages", "name"]}
+                    recognizer_instances.append(getattr(presidio_analyzer.predefined_recognizers, self._get_recognizer_name(recognizer), None)(**{**copied_recognizer, **language}))
```
omri374, line 137: WDYT about fetching all subclasses (recursively) instead of `presidio_analyzer.predefined_recognizers`? This could potentially allow more flexibility for users having their own predefined recognizers in code, but not part of the package (as they didn't clone the repo, just used the package).

omri374: I would suggest to simplify this line: `(**{**copied_recognizer, **language}))`. Perhaps create the unified dictionary prior to calling the init.

presidio-analyzer/tests/test_analyzer_engine_provider.py
RoeyBC: The thing is we're now introducing default configuration. I thought it would make sense to break a test in case something over there changes.

presidio-analyzer/conf/default_recognizers.yaml
```diff
@@ -0,0 +1,117 @@
+supported_languages:
```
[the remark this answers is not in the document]
RoeyBC: The use in here is by loading according to class names. users can add custom recognizers in this list too - there's a unit test that verifies it :D
Theoretically a user can add their own class to be loaded this way (which I guess is not a bad thing)

regarding the file - I agree, I'll remove the recognizers from there, but I'll have to add documentation about the analyzer file for clarity.

presidio-analyzer/conf/analyzer.yaml
RoeyBC: custom supports the following:
1. old version
2. new version with context:
      supported_languages:
        - language: en
          context: [credit]
        - language: es
          context: [tarjeta,credito]
3. new version without context (in which case it'll be just a string list):
  supported_languages: 
    - en
    - es

presidio-analyzer/app.py
RoeyBC: @SharonHart, you can still instantiate with the class itself, shouldn't be a breaking change

presidio-analyzer/conf/analyzer.yaml
RoeyBC: removed

presidio-analyzer/conf/analyzer.yaml
RoeyBC: removed

presidio-analyzer/presidio_analyzer/recognizer_registry_provider.py
RoeyBC: Doing it underneath. It would happen only if someone would remove the default conf file (we use the same functionality in the nlp engine provider).

presidio-analyzer/app.py
SharonHart: When we will drop the recognizer registry and make the loading from the file by default it will wont it?

SharonHart (dismissed):

presidio-analyzer/presidio_analyzer/recognizer_registry_provider.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+class RecognizerRegistryProvider:
+    """
+    Utility class for loading Recognizer Registry.
+
+    Use this class to load recognizer registry from a yaml file
+
+    :param conf_file: Path to yaml file containing registry configuration
+    :param registry_configuration: Dict containing registry configuraiton
```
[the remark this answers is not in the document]
RoeyBC: 100%

presidio-analyzer/conf/default_recognizers.yaml
```diff
@@ -0,0 +1,118 @@
...
+    supported_languages: 
+    - it
+    type: predefined
+
+  - name: PlPeselRecognizer
+    supported_languages: 
+    - pl
+    type: predefined
+
+  - CreditCardRecognizer
```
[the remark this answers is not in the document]
RoeyBC: Let's add a comment just for the case someone will add up a language and it would be missing.

Issue #1346, opened 2024-03-28
New Recognizer: Finnish Personal Identity Codes (Henkilötunnus).


honderr:
**Is your feature request related to a problem? Please describe.**
In Finland, personal identity codes (Henkilötunnus) are a key piece of personal data used for identification purposes. The personal identity code is granted upon birth or moving to reside in Finland and remains the same for one's whole life. As such, being able to detect and remove such codes from data is key.

**Describe the solution you'd like**
I have implemented a presidio pattern recognizer to identify such codes using the information available on the Finnish Digital and Population Data Services Agency (the authority manganging the issuance of the personal identity codes):
- https://dvv.fi/en/personal-identity-code  

In addition, some changes are being made to the Finnish personal identity codes which will widen the list of valid personal identity codes. These changes are taken into account in my implementation:
- https://dvv.fi/en/reform-of-personal-identity-code

**Describe alternatives you've considered**
I have been using my presidio recognizer for myself, however, I relized that that this would benefit a wider audience as several persons I know have to handle such data.

honderr:
I have now added this PR #1349

PR #1349, opened 2024-03-31
Added Finnish Personal Identity Code Recognizer.

honderr:
## Change Description
- Added Pattern Recognizer for the Finnish Personal Identity Code (Henkilötunnus).
- Added validation logic for the Finnish Personal Identity Code.
- Added tests for different valid and invalid Finnish Personal Identity Codes.

Describe your changes

## Issue reference

This PR fixes issue #1346 

## Checklist


honderr:

presidio-analyzer/tests/test_fi_personal_identity_code_recognizer.py
```diff
@@ -0,0 +1,142 @@
...
+        (
+            "020504B904H",
+            1,
+            ((0, 11),),
+        ),
+        # invalid Personal Identity Codes scores
+        ("111111-111A", 0, ()),
+        ("111111+110G", 0, ()),
+        ("012245A110G", 0, ()),
+        ("010324A110G", 0, ()),
```
omri374, line 132: Please add a sample with the personal identity code within a sentence (e.g "my personal identity code, ABCD, thank you".

honderr: Thanks a lot for your feedback @omri374 

I would kindly ask for one more piece of advice, before making the changes. Following the example of other pattern recognizers, I used word boudaries `\b` in my regular expression.

In real-life cases, we might have phrases like "my personal id is111111-111A" which would not be caught. This is, in my experience, a common issues especially in chat contexts.

Do you recommend:
1. Keeping the solution as is.
2. Removing word boundaries and reducing the score of the recognizer.
3. Adding another (very weak e.g. 0.1 score) Pattern to the list without the word boundaries.

Thanks!

presidio-analyzer/presidio_analyzer/predefined_recognizers/fi_personal_identity_code_recognizer.py
```diff
@@ -0,0 +1,55 @@
...
+
+    :param patterns: List of patterns to be used by this recognizer
+    :param context: List of context words to increase confidence in detection
+    :param supported_language: Language this recognizer supports
+    :param supported_entity: The entity this recognizer can detect
+    """
+
+    PATTERNS = [
+        Pattern(
+            "Finnish Personal Identity Code (Medium)",
```
omri374, line 19: Another point, if ok- please add a reference to the logic (a government website if possible) for future users to see where this logic is coming from.

honderr: @omri374 thanks. I have now made the changes.

presidio-analyzer/presidio_analyzer/predefined_recognizers/fi_personal_identity_code_recognizer.py
```diff
@@ -0,0 +1,63 @@
...
+            supported_language=supported_language,
+        )
+
+    def validate_result(self, pattern_text: str) -> Optional[bool]:
+        """Validate the pattern by using the control character."""
+
+        # More information on the validation logic from:
+        # https://dvv.fi/en/personal-identity-code
+        # Under "How is the control character for a personal identity code calculated?".
+        date_part = pattern_text[0:6]
```
omri374, line 53: Please check for len prior to running this. If in the future the pattern changes and we pass shorter strings, it will raise an exception and stop the whole process.

presidio-analyzer/presidio_analyzer/predefined_recognizers/fi_personal_identity_code_recognizer.py
omri374: Thanks!

presidio-analyzer/presidio_analyzer/predefined_recognizers/fi_personal_identity_code_recognizer.py
honderr: Thanks a lot for your feedback @omri374 . In my understanding, the regular expression is already taking care of that?

My understanding is that, when entering `validate_result`, we have already matched `(\d{6})([+-ABCDEFYXWVU])(\d{3})([0123456789ABCDEFHJKLMNPRSTUVWXY])` and therefore, we can be sure that the string we mainpulate (`pattern_text`) is already following the right format (`pattern_text[0:6]` are all digits). 

In other words, I understand that the regular expression is already checking the length for us? Is my understanding correct or did I misunderstand your previous comment?

omri374: Yes, you are correct. I wasn't clear enough on my comment. Regex patterns sometimes change, and adding another check for len here is useful as we might unexpectedly break something if the pattern changes but the checksum doesn't. An exception here would not only affect the users of this recognizer, but all users of the tool. Hope this makes sense.

honderr: Thanks @omri374. Would something like this suffice?

```
if len(pattern_text) != 11:
    return False
```

Then, the, `datetime.strptime(date_part, "%d%m%y")` will take care of validating the `date_part`.

While I understand that we would like to avoid exceptions in this context, I must admit that I fail to think of scenarios where we would change the pattern without changing the checksum and, the tests would still pass?

honderr: Thanks a lot for your reply.

I have now added the length check.

omri374 (approved):

omri374:
Thanks for the great contribution! Merged.

PR #1367, opened 2024-04-24
Loading analyzer engine & recognizer registry from configuration file

RoeyBC:
## Change Description

This PR adds a file-based configuration for presidio-analyzer. It allows users to:
1. Configure their app (either Python or REST) using a set of three yaml files for each main engine:
    - Analyzer
    - Recognizer Registry
    - NLP Engine
2. Pass those files into the Docker image and read them when the REST version of Presidio (in flask) is loaded
3. Easily define which recognizer to use, which language they should support + specific configuration for each (for example, a list of context words per language)


omri374 (approved):

omri374 (approved):

presidio-analyzer/Dockerfile
```diff
@@ -2,16 +2,32 @@ FROM python:3.9-slim
...
 ARG NAME
 ARG NLP_CONF_FILE=presidio_analyzer/conf/default.yaml
+ARG ANALYZER_CONF_FILE=presidio_analyzer/conf/default_analyzer.yaml
+ARG RECOGNIZER_REGISTRY_CONF_FILE=presidio_analyzer/conf/default_recognizers.yaml
 ENV PIPENV_VENV_IN_PROJECT=1
 ENV PIP_NO_CACHE_DIR=1
+
+ENV ANALYZER_CONF_FILE=${ANALYZER_CONF_FILE}
+ENV RECOGNIZER_REGISTRY_CONF_FILE=${RECOGNIZER_REGISTRY_CONF_FILE}
+ENV NLP_CONF_FILE=${NLP_CONF_FILE}
```
SharonHart, line 12: I think that the ENV lines are redundant

presidio-analyzer/Dockerfile
```diff
@@ -2,16 +2,32 @@ FROM python:3.9-slim
...
+COPY ${ANALYZER_CONF_FILE} /usr/bin/${NAME}/${ANALYZER_CONF_FILE}
+COPY ${RECOGNIZER_REGISTRY_CONF_FILE} /usr/bin/${NAME}/${RECOGNIZER_REGISTRY_CONF_FILE}
+COPY ${NLP_CONF_FILE} /usr/bin/${NAME}/${NLP_CONF_FILE}
+
 WORKDIR /usr/bin/${NAME}
 
 COPY ./Pipfile* /usr/bin/${NAME}/
+
+# Install essential build tools
+RUN apt-get update \
```
SharonHart, line 23: On every conf or pip change, this will rerun, I think that its better to move it up

SharonHart: Why was it needed by the way?

presidio-analyzer/presidio_analyzer/conf/default_recognizers.yaml
```diff
@@ -0,0 +1,140 @@
...
+    type: predefined
+
+  - name: AuMedicareRecognizer
+    supported_languages: 
+    - en
+    type: predefined
+
+  - name: InPanRecognizer
+    supported_languages: 
+    - en
```
SharonHart, line 84: in?

presidio-analyzer/Dockerfile
omri374: Build failed on mac

presidio-analyzer/presidio_analyzer/conf/default_recognizers.yaml
omri374: Yes, but one can change this pretty easily if needed

presidio-analyzer/Dockerfile
omri374: The app.py reads those values from env variables. Is there a better option to have the user introduce those parameters and have them set as env variables? Because these are paths, I wasn't sure if passing env variables to the docker build call would end up having the right path, but I have quite limited experience with Docker.

SharonHart: Got it, makes sense then

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry.py
```diff
@@ -70,18 +71,55 @@ class RecognizerRegistry:
...
         if recognizers:
             self.recognizers = recognizers
         else:
             self.recognizers = []
         self.global_regex_flags = global_regex_flags
+        self.supported_languages = (
+            supported_languages if supported_languages else ["en"]
+        )
+
+    def _create_nlp_recognizer(
```
SharonHart, line 86: nit: move privates below publics, also in analyzer_engine_provider.py

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry_provider.py
```diff
@@ -0,0 +1,341 @@
...
+        ):
+            return [
+                {
+                    "supported_language": language,
+                    "context": self._get_recognizer_context(recognizer_conf),
+                }
+                for language in self.supported_languages
+            ]
+
+        if isinstance(recognizer_conf["supported_languages"][0], str):
```
SharonHart, line 173: won't dropping it give the same outcome as the list below?

SharonHart (approved):
Amazing outcome :)

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry_provider.py
omri374: What do you mean? in the yaml file, a language could either be a string ("en") or a dictionary, with a name of a language and context per language.

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry.py
omri374: @RoeyBC let's handle on the next PR

Issue #1361, opened 2024-04-22
Enhancing the decision process text when working with images


NuiMrme:
**Is your feature request related to a problem? Please describe.**
The decision process output prints out the entity_type, start_position, end_position and the score. When working with longer sequences of texts or with images, printing start = 204 end = 217 doesn't really mean anything and it is hard to see where that is. 

**Describe the solution you'd like**
Add an entity_text where the the text in question is also printed: printing start = 204 end = 217 entity_text = "Saint Antonio"

I solved this on my version by adding 

`entity_text: str,`

in recognizer_result.py __init__ function which then affected also image_analzer_engine.py and image_recognizer_results.py 
but the output is rather more readable

NuiMrme:
While at it, in analyzer_engine.py line:222 I modified the line so that the code prints out every case in a new line , even more readable 
`json.dumps([str(result.to_dict()) for result in results]**, indent=2**),`

omri374:
Does this help? https://github.com/microsoft/presidio/discussions/925#discussioncomment-3781214

NuiMrme:
Sorry that wasn't well explained. I'm not reporting a bug but rather a feature I implemented on my version of Presidio that might help others too. See when you work with images or a lot of text while having your `log_decision_process=True` , the printed text will be for many many instances where it detected something and the log becomes unreadable. Please remember it prints that automatically no explicit print command is used as in your shared comment above.
If I have one line example thats fine I can look quickly see what these position refer to but when you have many of these stacked together because it is coming from an image of a document , you don't know anymore what is what. So I did the above mentioned modifications to change it a bit to make it more readable

Every new case will begin in a new line and observe that there is now a  'entity_text' which will show that text that is detected (I covered it with red for the obvious reasons), now you don't have to guess what line was that in the image what position etc... This is more readable and help the anlaysis of the annomyization results. 


**before** 
![image](https://github.com/microsoft/presidio/assets/23320288/641e3ae2-9cf0-43ce-9278-b5dfc1b1dc23)

**after** 
![image](https://github.com/microsoft/presidio/assets/23320288/a312d247-acc0-43a6-9308-552f1762d461)

NuiMrme:
> One of the reasons we intentionally left out the actual identified text, is because it is essentially PII you might not want to log or return. If you have a suggestion on how to allow this, perhaps not asa default setting, we'd be happy to hear.
> 
> I totally agree that there are cases, especially with the images module, where returning or logging the actual text is needed.

[...]

PR #1369, opened 2024-04-26
Enhancing the log by adding an entity_text and creating a new line for every case

NuiMrme:
## Change Description

The log by default contains information about the entity_type, start, end and the score in addition to the analysis explanation. Added an entity_text too pin point the text in question too for investigation purposes. 
Also changed the json dumbs to begin in a new line for every case to enhance readability.

## Issue reference

This PR fixes issue #1361

## Checklist

No unit tests , I tracked down every file that is affected manually and tested on my own data.

### DIFF
--- a/presidio-analyzer/presidio_analyzer/analyzer_engine.py
+++ b/presidio-analyzer/presidio_analyzer/analyzer_engine.py
@@ -228,7 +228,7 @@ class AnalyzerEngine:
         if self.log_decision_process:
             self.app_tracer.trace(
                 correlation_id,
-                json.dumps([str(result.to_dict()) for result in results]),
+                json.dumps([str(result.to_dict()) for result in results],indent=4),
             )
 
         # Remove duplicates or low score results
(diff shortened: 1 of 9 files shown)


omri374:
Thanks @NuiMrme. As discussed, it is better to add the entity value to the decision log and not to every result. There's a class called `AnalysisExplanation` which could potentially be a good place to store the entity value.

NuiMrme:
> Thanks @NuiMrme. As discussed, it is better to add the entity value to the decision log and not to every result. There's a class called `AnalysisExplanation` which could potentially be a good place to store the entity value.

I think I'm missing something but I need free time to figure it out. I thought once `log_decision_process="false"` nothing is printed out ...

omri374:
`RecognizerResult` is the object returned when calling `analyzer.analyze`. If you add a new field to it, it would mean that this new field would be added to all results. Our preference here is to only add the identified text to the analysis explanation, so that if someone is interested in debugging or tracing within Presidio, they would be able to do so, but if there are those who don't want PII returned, then such change wouldn't affect them.

PR #1386, opened 2024-05-15
From Pipenv to PDM

SharonHart:
## Change Description

* Drop pipenv for PDM
* Implement DRY by instructing PDM to take dependencies from the new pyproject.toml, dropping the Pipfiles. 
* That change required moving server dependencies to the pyproject.toml, as 'server' optional dependencies. Same goes for dev dependencies.

## Checklist


omri374:
Thanks @SharonHart I like the DRY approach, but it looks like we're doing a detour instead of just using Poetry which directly uses the pyproject.toml file. Maybe we can test poetry and see if it works well for us? If it doesn't fit our needs for some reason, I agree that this PR is the right approach.
WDYT?

SharonHart:
> Thanks @SharonHart I like the DRY approach, but it looks like we're doing a detour instead of just using Poetry which directly uses the pyproject.toml file. Maybe we can test poetry and see if it works well for us? If it doesn't fit our needs for some reason, I agree that this PR is the right approach. WDYT?

Yes, I this is the intermediate step, I thought to introduce poetry later on, and in stages since I think it would mean migrating from setuptools as well

omri374:
Ok sounds good. Please update https://github.com/microsoft/presidio/blob/main/docs/development.md too

presidio-analyzer/Dockerfile
```diff
@@ -17,14 +17,14 @@ COPY ${NLP_CONF_FILE} /usr/bin/${NAME}/${NLP_CONF_FILE}
...
+COPY ./pyproject.toml /usr/bin/${NAME}/
 COPY ./Pipfile* /usr/bin/${NAME}/
 
 # Install essential build tools
 RUN apt-get update \
   && apt-get install -y build-essential
 
-RUN pip install pipenv \
-  && pipenv install --deploy
+RUN pip install pipenv && pipenv install pipenv install -e '.[server,stanza,azure-ai-language]'
```
omri374, line 27: Should the other Dockerfiles be updated too?

SharonHart: Yes, I was POCing with analyzer

SharonHart:
> Ok sounds good. Please update https://github.com/microsoft/presidio/blob/main/docs/development.md too

I am also considering pdm

docs/development.md
```diff
@@ -45,7 +45,7 @@ Follow these steps when starting to work on a Presidio service with Pipenv:
     For example, in the `presidio-analyzer` folder, run:
 
     ```
-    pipenv install --dev --skip-lock
+    pipenv install -e '.[all]' --skip-lock
```
omri374, line 48: pdm

SharonHart:
Closed in favor of https://github.com/microsoft/presidio/pull/1391

SharonHart:
Closed in favor of https://github.com/microsoft/presidio/pull/1391

PR #1393, opened 2024-05-30
changing predefined recognizers to use the config file

RoeyBC:
## Change Description

This PR is going to accomplish the following: 
1. Refactor load_predefined_recognizers to load the settings from the default yml file
2. Add warning in case recognizer is listed with a language not supported by the registry
3. raise an exception in case there's discrepancy between the supported language in the registry and the one listed in the analyzer
4. Edit the docs accordingly I guess :D

## Checklist


RoeyBC:
image
when you're refactored, added documentation & testing and you've got less text, you know it's a good change :D

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry_provider.py
```diff
@@ -281,17 +291,26 @@ def create_recognizer_registry(self) -> RecognizerRegistry:
...
             if isinstance(recognizer_conf, PatternRecognizer):
-                recognizer_conf.global_regex_flags = fields["global_regex_flags"]
+                recognizer_conf.global_regex_flags = self.global_regex_flags
 
-        recognizer_instances = [
-            recognizer
-            for recognizer in recognizer_instances
-            if recognizer.supported_language in self.supported_languages
-        ]
+        index = 0
```
SharonHart, line 296: Feels like this can get bad easily in the future, can be simplified with less mutability, maybe?
`filtered_recognizer_instances = [r for r in recognizer_instances if language_supported_globally(r)]`
and print inside language_supported_globally
Or with `filterfalse

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry.py
```diff
@@ -126,67 +93,17 @@ def load_predefined_recognizers(
...
+        from .recognizer_registry_provider import RecognizerRegistryProvider
+        registry_configuration = {"global_regex_flags": self.global_regex_flags}
+        if languages is not None:
+            registry_configuration["supported_languages"] = languages
+
+        recognizers = RecognizerRegistryProvider(
+            registry_configuration=registry_configuration
+        ).create_recognizer_registry().recognizers
+
+        self.recognizers.extend(recognizers)
```
omri374, line 106: Worth checking for duplications. Isn't the nlp provider added twice?

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry_provider.py
```diff
@@ -255,9 +255,19 @@ def create_recognizer_registry(self) -> RecognizerRegistry:
             fields[field] = self.configuration.get(field, self.default_values[field])
 
         self.supported_languages = fields["supported_languages"]
+        self.global_regex_flags = fields["global_regex_flags"]
```
omri374, line 258: Add to init

presidio-analyzer/tests/test_analyzer_engine.py
```diff
@@ -88,6 +91,15 @@ def test_when_analyze_with_predefined_recognizers_then_return_results(
     assert len(results) == 1
     assert_result(results[0], "CREDIT_CARD", 14, 33, max_score)
 
+def test_when_analyze_with_unsupported_language_different_then_fail(
```
omri374, line 94: Consider adding a test with multiple inputs, e.g. mismatch, one of them is empty, null, etc. This would help ensure this logic is true in future revisions.

docs/analyzer/analyzer_engine_provider.md
```diff
@@ -75,6 +75,10 @@ The configuration file contains the following parameters:
   - `nlp_configuration`: Configuration given to the NLP engine which will detect the PIIs and extract features for the downstream logic.
   - `recognizer_registry`: All the recognizers that will be used by the analyzer. 
 
+!!! note "Note"
+
+    supported_languages must be identical to the same field in recognizer_registry
```
omri374, line 80: Nit
```suggestion
    `supported_languages` must be identical to the same field in recognizer_registry
```

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry_provider.py
RoeyBC: Love it, deal

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry_provider.py
RoeyBC: That's what I did initially, just take note that you still have to do this assignment because the functions in there will still `self.supported_lauguages` etc

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry.py
```diff
@@ -126,67 +93,17 @@ def load_predefined_recognizers(
...
-                for rc in recognizers_map.get("ALL", [])
-            ]
-            self.recognizers.extend(all_recognizers)
+
+        from .recognizer_registry_provider import RecognizerRegistryProvider
+        registry_configuration = {"global_regex_flags": self.global_regex_flags}
+        if languages is not None:
+            registry_configuration["supported_languages"] = languages
+
+        recognizers = RecognizerRegistryProvider(
```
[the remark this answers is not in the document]
RoeyBC: There are two issues with a new module:
1. `recognizer_registry_provider` was made to load recognizers from a config file, so I'm not sure we should extract it any further, it's pretty thin in functionality. 
2. We can't create this logic in the registry because we have no nlp engine at this point (which could be scpaCy or transformer or whatever. 

The only entity that is aware of both the registry and the nlp engine is the `analyzer_engine`. How about we create this logic here? and just make sure to keep on using the same logic?

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry.py
RoeyBC: `recognizer_registry` doesn't add the nlp recognizer, so there's no duplication. In fact when the analyzer get initialized without a registry we do the following: 
```
        if not registry:
            logger.info("registry not provided, creating default.")
            provider = RecognizerRegistryProvider(
                registry_configuration={"supported_languages": self.supported_languages}
            )
            registry = provider.create_recognizer_registry()
            registry.add_nlp_recognizer(nlp_engine=self.nlp_engine)
```

Which is basically this function now :D

SharonHart (dismissed):

SharonHart (approved):

viclang:
When will this feature be released?

viclang:
Currently I can't import `from presidio_analyzer import RecognizerRegistryProvider` is this problem solved by this release?

omri374 (approved):

PR #1439, opened 2024-08-18
Update installation.md

MWest2020:
up*

## Change Description

Simple improvement on docker command

No issue, or tests needed. JUst the current command doesn't work

This PR fixes issue #XX

## Checklist

- [V ] I have reviewed the [contribution guidelines](https://github.com/microsoft/presidio/blob/main/CONTRIBUTING.md)
- [V ] I have signed the CLA (if required)
- [V ] My code includes unit tests
- [ V] All unit tests and lint checks pass locally
- [V ] My PR contains documentation updates / additions if required

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -143,7 +143,7 @@ Then, build the containers locally.
 From the root folder of the repo:
 
 ```sh
-docker-compose --build
+docker-compose up --build
 ```
 
 To run all Presidio services:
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -143,7 +143,7 @@ Then, build the containers locally.
 From the root folder of the repo:
 
 ```sh
-docker-compose --build
+docker-compose up --build
```
omri374, line 146: The command after (in line 152) does the `up` part, but you're right there's a problem here. 
I would suggest to join the two and just call `docker-compose up --build`

omri374 (approved):

PR #1446, opened 2024-09-01
Add UK National Insurance Number Recognizer

hhobson:
## Change Description

- Add recognizer to detect UK National Insurance Number + tests
- Update docs on creating pre-defined recognizers to use config file

## Issue reference

NA

## Checklist


presidio-analyzer/presidio_analyzer/predefined_recognizers/uk_nino_recognizer.py
```diff
@@ -0,0 +1,40 @@
...
+
+    :param patterns: List of patterns to be used by this recognizer
+    :param context: List of context words to increase confidence in detection
+    :param supported_language: Language this recognizer supports
+    :param supported_entity: The entity this recognizer can detect
+    """
+
+    PATTERNS = [
+        Pattern(
+            "NINO (medium)",
```
omri374, line 18: Couldc you please provide the source of this regex? How did you test it aside from the added unit tests?

presidio-analyzer/tests/test_uk_nino_recognizer.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+@pytest.fixture(scope="module")
+def entities():
+    return ["UK_NINO"]
+
+
+@pytest.mark.parametrize(
+    "text, expected_len, expected_positions, expected_score_ranges",
+    [
```
SharonHart, line 19: I would also modify some test cases to use context words

SharonHart (dismissed):

presidio-analyzer/tests/test_uk_nino_recognizer.py
hhobson: Had `national insurance` in there and have added `nino`.

As the recognizer is being run outside of Analyzer Engine (which looks like the standard approach for all recognizer tests), the context works don't affect the score. So I've left the expected scores at 0.5.

presidio-analyzer/presidio_analyzer/predefined_recognizers/uk_nino_recognizer.py
hhobson: The UK Government defines the National Insurance Number format [Here](https://www.gov.uk/hmrc-internal-manuals/national-insurance-manual/nim39110). I wrote this taking some inspiration from t[his Stackoverflow discussion](https://stackoverflow.com/questions/10204378/regular-expression-to-validate-uk-national-insurance-number/17779536).

As the National Insurance Number is case insensitive and Presido allows users to set regex flags at global/registry levels, I included both upper and lower case characters in the regex.

presidio-analyzer/tests/test_uk_nino_recognizer.py
SharonHart: Sorry, for some reason I remembered context was possible in how we define the tests, approved

SharonHart (approved):

Issue #1452, opened 2024-09-15
Context with hash marks do not work as expected 


claesmk:
**Describe the bug**
In the predefined US_SSN recognizer [context](https://github.com/microsoft/presidio/blob/4aeb56bf5e2e1187dd535785cc41590ca3a4d5eb/presidio-analyzer/presidio_analyzer/predefined_recognizers/us_ssn_recognizer.py#L24) the ssn# and ss# context words do not work as expected because spacy tokenizes the hash mark: 
```
[decision_process][INFO][None][nlp artifacts:{"entities": ["My ssn#"], "tokens": ["My", "ssn", "#", "is", "123", "-", "45", "-", "1234"], "lemmas": ["my", "ssn", "#", "be", "123", "-", "45", "-", "1234"], "tokens_indices": [0, 3, 6, 8, 11, 14, 15, 17, 18], "keywords": ["ssn", "123", "45", "1234"], "scores": [0.85]}]
```
**ssn#** works OK because this is tokenized as ["ssn" and "#"] and ssn is also in the context list, however, **ss#** does not work 

```
[decision_process][INFO][None][nlp artifacts:{"entities": [], "tokens": ["My", "ss", "#", "is", "123", "-", "45", "-", "1234"], "lemmas": ["my", "ss", "#", "be", "123", "-", "45", "-", "1234"], "tokens_indices": [0, 3, 5, 7, 10, 13, 14, 16, 17], "keywords": ["ss", "123", "45", "1234"], "scores": []}]
```

**To Reproduce**
1. Use US_SSN entity and a string with ss# 
2. See that it is not detected as context (0.5 default score is not improved)
```
{'recognizer': 'UsSsnRecognizer', 'pattern_name': 'SSN5 (medium)', 'pattern': '\\b([0-9]{3})[- .]([0-9]{2})[- .]([0-9]{4})\\b'...edium)`', 'score_context_improvement': 0, 'supportive_context_word': '', 'validation_result': None, 'regex_flags': regex.I|M|S}
````

**Expected behavior**
All words in the predefined [context](https://github.com/microsoft/presidio/blob/4aeb56bf5e2e1187dd535785cc41590ca3a4d5eb/presidio-analyzer/presidio_analyzer/predefined_recognizers/us_ssn_recognizer.py#L24) should be detected correctly.  If values with a hash mark cannot be detected correctly they should at least be removed from the list.

**Screenshots**
See logging and debugger output above

**Additional context**
n/a

[...]

PR #1455, opened 2024-09-28
Update US_SSN CONTEXT and unit test

claesmk:
## Change Description

As discussed in #1452, two of the CONTEXT words ("ssn#" and "ss#")  in UsSsnRecognizer() did not work correctly since # is tokenized.

The US_SSN tests in context_sentences_tests.txt were also updated.  078051120 is [invalidated](https://github.com/microsoft/presidio/blob/c54ce2b9cf8795a168711a9fc6954d47d3269683/presidio-analyzer/presidio_analyzer/predefined_recognizers/us_ssn_recognizer.py#L76) so most of these tests were not doing anything.  The correct place to check valid/invalid numbers is in [test_us_ssn_recognizer](https://github.com/microsoft/presidio/blob/c54ce2b9cf8795a168711a9fc6954d47d3269683/presidio-analyzer/tests/test_us_ssn_recognizer.py#L17)

#### Backwards Compatibility Note

Prior to this change if the text was "My ssn# is xxx-xx-xxxx" then "ssn" instead of "ssn#" would have been used for a context match.  Since ssn is already in the CONTEXT list, this would have matched, and will continue to match.

If the text was "My ss# is xxx-xx-xxxx" then "ss" would have been used and since that was not present in the CONTEXT list the match would have failed. Therefore, removing ss# should have no functional effect.

## Issue reference

This PR fixes issue #1452 

## Checklist

- [ x] I have reviewed the [contribution guidelines](https://github.com/microsoft/presidio/blob/main/CONTRIBUTING.md)
- [x ] My code includes unit tests
- [x ] All unit tests and lint checks pass locally

### DIFF
--- a/presidio-analyzer/presidio_analyzer/predefined_recognizers/us_ssn_recognizer.py
+++ b/presidio-analyzer/presidio_analyzer/predefined_recognizers/us_ssn_recognizer.py
@@ -27,8 +27,8 @@ class UsSsnRecognizer(PatternRecognizer):
         # "sec", # Task #603: Support keyphrases ("social sec")
         "ssn",
         "ssns",
-        "ssn#",
-        "ss#",
+        # "ssn#",  # iss:1452 - a # does not work with LemmaContextAwareEnhancer
+        # "ss#",  # iss:1452 - a # does not work with LemmaContextAwareEnhancer
         "ssid",
     ]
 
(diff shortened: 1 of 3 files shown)


claesmk:

omri374 (approved):
LGTM. Thanks!

Issue #1463, opened 2024-10-07
Transformer model may be ignoring general entity types


michhar:
Amazing project, thank you so much for `presidio` and all of the work here.

I am noticing with a few different Hugging Face transformer models, that some of the listed entities associated with the engine, are not being picked up even with mapping.

I am using the following on macOS:

```
presidio_analyzer==2.2.355
presidio_anonymizer==2.2.355
```

For example, here is my code to set up the engine:

```python
# Transformer model config
tf_model_config = [
    {"lang_code": "en",
     "model_name": {
         "spacy": "en_core_web_sm", 
         "transformers": "StanfordAIMI/stanford-deidentifier-base"
    }
}]

# Entity mappings
mapping = dict(
    PER="PERSON",
    LOC="LOCATION",
    ORG="ORGANIZATION",
    AGE="AGE",
    ID="ID",
    EMAIL="EMAIL",
    DATE="DATE_TIME",
    PHONE="PHONE_NUMBER",
    PERSON="PERSON",
    LOCATION="LOCATION",
    GPE="LOCATION",
    ORGANIZATION="ORGANIZATION",
    NORP="NRP",
    PATIENT="PERSON",
    STAFF="PERSON",
    HOSP="ORGANIZATION",
    PATORG="ORGANIZATION",
    TIME="DATE_TIME",
    HCW="PERSON",
    HOSPITAL="ORGANIZATION",
    FACILITY="LOCATION"
)

tf_model_configuration = NerModelConfiguration(
    model_to_presidio_entity_mapping=mapping,
    alignment_mode="expand", # "strict", "contract", "expand"
    aggregation_strategy="max") # "simple", "first", "average", "max"

tf_engine = TransformersNlpEngine(
    models=tf_model_config,
    ner_model_configuration=tf_model_configuration)

# Transformer-based analyzer
analyzer_tf = AnalyzerEngine(
    nlp_engine=tf_engine, 
    supported_languages=["en"]
)

# Default anonymizer
anonymizer_default = AnonymizerEngine()
```

When I list the entities for the transformer engine with this code:
```python
print(f'Entities for transformer engine:  {analyzer_tf.nlp_engine.get_supported_entities()}')
```
I get `Entities for transformer engine:  ['DATE_TIME', 'NRP', 'EMAIL', 'PHONE_NUMBER', 'ORGANIZATION', 'AGE', 'LOCATION', 'PERSON', 'ID']
Selection deleted
`

[...]

PR #1467, opened 2024-10-10
Updates to the transformers conf docs and yaml file

omri374:
updated the docs for a transformers based pipeline, to make it clearer, and added an example for configuration through code. Based on the discussion in #1463 
Also changed the default config in the docs to match the one in the code.

### DIFF
--- a/docs/analyzer/nlp_engines/transformers.md
+++ b/docs/analyzer/nlp_engines/transformers.md
@@ -55,7 +55,75 @@ Then, also download a spaCy pipeline/model:
 python -m spacy download en_core_web_sm
 ```
 
-#### Creating a configuration file
+
+### Configuring the NER pipeline
+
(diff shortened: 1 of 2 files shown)


SharonHart (approved):

michhar:
@omri374 Thank you for making these changes.  One question in my mind still is how to find the Hugging Face NER model labels...might be out of scope for the docs here, but perhaps a link or pointer to this could help since it's needed for the configuration file.  Again, thanks for making things clearer.

omri374:
Thanks @michhar. Will directing the user to look into the model's config (e.g. https://huggingface.co/dslim/bert-base-NER-uncased/blob/main/config.json) be the right approach? Not all models explicitly state the classes on their huggingface hub website.

omri374:
Please see #1472

Issue #1496, opened 2024-12-12
Update Presidio Containers to use Gunicorn


RKapadia01:
### Summary:
Replace Flask's built-in server with Gunicorn in Presidio's Docker containers for better performance, scalability, and production readiness.

### Benefits:

- Operate more efficiently with improved resource utilization and performance, particularly under higher traffic loads
- Added ability to scale workers and handle concurrent requests
- Aligns with best practices for deploying Python web applications into production scenarios

RKapadia01:
Have raised https://github.com/microsoft/presidio/pull/1497 to address this

PR #1497, opened 2024-12-12
Update presidio containers to use gunicorn

RKapadia01:
## Change Description

- Modified the main application entry points in `presidio-analyzer/app.py`, `presidio-anonymizer/app.py`, and `presidio-image-redactor/app.py` to use the `app` instance directly instead of creating a new `Server` instance when running the application. 
-  Added `PORT` and `WORKERS` environment variables to the Dockerfiles for all 3 services, to allow configuration of the server port and number of worker processes. 
- Updated the command from Flaks to Gunicorn in the Dockerfiles

## Issue reference

This PR fixes issue #1496 

## Checklist


SharonHart (approved):
Looks great, thanks!

presidio-analyzer/pyproject.toml
```diff
@@ -36,9 +36,10 @@ azure-ai-textanalytics = { version = "*", optional = true }
 azure-core = { version = "*", optional = true }
 transformers = { version = "*", optional = true }
 huggingface_hub = { version = "*", optional = true }
+gunicorn = {version = "*", optional = true}
 
 [tool.poetry.extras]
-server = ["flask"]
+server = ["flask", "gunicorn"]
```
omri374, line 42: Please add the `gunicorn` license into the NOTICE file.

omri374 (commented):
Looks great! Thanks :) I've added two minor points

presidio-analyzer/Dockerfile
```diff
@@ -10,6 +10,9 @@ ENV ANALYZER_CONF_FILE=${ANALYZER_CONF_FILE}
 ENV RECOGNIZER_REGISTRY_CONF_FILE=${RECOGNIZER_REGISTRY_CONF_FILE}
 ENV NLP_CONF_FILE=${NLP_CONF_FILE}
 
+ENV PORT=3000
```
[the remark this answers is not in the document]
RKapadia01: No, same as the default port. I added this to the Dockerfile, so it can be configured if the user desires to change the port.

presidio-analyzer/pyproject.toml
RKapadia01: Updated the NOTICE file, thanks

presidio-analyzer/app.py
```diff
@@ -135,8 +135,9 @@ def supported_entities() -> Tuple[str, int]:
         def http_exception(e):
             return jsonify(error=e.description), e.code
 
+server = Server()
```
omri374, line 138: Might be nitty picking, but I'm curious on what is the difference between instantiating Server here compared to within the `if name == main`. For instance, if I import app.py, the Server instance would be instantiated. I'm not sure if importing app.py is a use case, but it would be a side effect of this change in case somebody does this anywhere. WDYT?

RKapadia01: Thats a great pickup. When running with Gunicorn, it doesn't execute the code in the`__main__` block. It just imports the module and looks for the app object to serve the application, which is why I moved the server code out of the `__main__` block.

However, I didn't remove the `__main__` block entirely, so if anyone wants to run just the flask development server they still have to ability to do so.

To avoid your concern of instantiating the server on import of app.py, we can avoid it by doing something like this:
```python
def create_app():
    server = Server()
    return server.app

if __name__ == "__main__":
    app = create_app()
    port = int(os.environ.get("PORT", DEFAULT_PORT))
    app.run(host="0.0.0.0", port=port)
```

We can then run Gunicorn in the Dockerfile like this:
```
CMD poetry run gunicorn -w $WORKERS -b 0.0.0.0:$PORT 'app:create_app()'
```

This should ensure that if anyone is just importing app.py it won't create the server instance. If you're happy with this, I'll commit the changes

omri374: Thanks! This is a better solution IMHO so let's go with it.

omri374:
Thanks! Looks great

omri374 (approved):

omri374:
Added `noqa` to ignore linting in this case

omri374 (approved):

PR #1504, opened 2025-01-02
example GLiNER integration

omri374:
## Change Description

Integrating GLiNER as an `EntityRecognizer`

## Issue reference

This PR fixes issue #XX

## Checklist

### DIFF
--- a/presidio-analyzer/presidio_analyzer/predefined_recognizers/gliner_recognizer.py
+++ b/presidio-analyzer/presidio_analyzer/predefined_recognizers/gliner_recognizer.py
@@ -102,7 +102,8 @@ class GLiNERRecognizer(LocalRecognizer):
 
     def load(self) -> None:
         """Load the GLiNER model."""
-
+        if not GLiNER:
+            raise ImportError("GLiNER is not installed. Please install it.")
         self.gliner = GLiNER.from_pretrained(self.model_name)
 
     def analyze(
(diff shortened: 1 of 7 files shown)


presidio-analyzer/presidio_analyzer/predefined_recognizers/gliner_recognizer.py
```diff
@@ -0,0 +1,167 @@
...
+                raise ValueError(
+                    "entity_mapping and supported_entities cannot be used together"
+                )
+
+            self.model_to_presidio_entity_mapping = entity_mapping
+        else:
+            if not supported_entities:
+                logger.info(
+                    "No supported entities provided, "
+                    "using default entities from NerModelConfiguration"
```
SharonHart, line 72: I see quite a lot more entotoes than what we saw so far from NER modles, do you want to extend the list of keys following GliNER at some point?

SharonHart (approved):

presidio-analyzer/presidio_analyzer/predefined_recognizers/gliner_recognizer.py
omri374: Hey, do you mean that we should update the list of entities we have as default? The way I see it, GLiNER's added value is when you provide a custom set of entities, as it can work in a zero-shot way.
With regards to mapping, there's a mapping dict in the class, with the ambition to standardize the entities with Presidio's (where possible)

navalev (approved):

presidio-analyzer/presidio_analyzer/predefined_recognizers/gliner_recognizer.py
omri374: @SharonHart let's take it offline

PR #1513, opened 2025-01-21
Update presidio_streamlit.py

rz1027:
In the demo the analyzer is being instantiated but not used.
This might work in Streamlit (because of st.cache), but in other scenarios the model will get reloaded on each run of the function.


omri374:
Thanks. What other scenarios are you looking at? The code here is used primarily to experiment with different setups.

Issue #1526, opened 2025-02-14
IN_PAN indian pan number predefined recognizer overwrites every 10 character alphanumeric string- fitting to regex - identifier.


sasever:
**Describe the bug**
IN_PAN Indian pan number added with "PR: New Predefined Recognizer: IN_PAN #1100" by @devopam ,  does not have an implemented checksum although mentioned multiple times  in the PR and only looks at the regex. therefore overwrites every 10 digit alphanumeric identifier that fits the regex. being it predefined  also creates an issue in disabling, even after disabling with supported_entities. remove  keeps marking unrelated words with IN_PAN

**To Reproduce**
providing 3 faker generated sample text that it should not mark as IN_PAN. words that are falsely marked as IN_PAN are highlighted with bold:

1. "Near way game. Rather full alone some meeting medical. His yes whether. The user, **EMP34050**, reported an issue with their phone. The problem is described as Unable to receive messages, and the issue occurs at 771 Victor Land
East Evan, IL 70246. Further details include that the phone is often showing no signal."
2. "Actually break point world trade skin federal. Others head building message yeah. Team bad skill fly couple environment. The user, **188.22.80.89**, reported an issue with their phone. The problem is described as Error **connecting** to network, and the issue occurs at Unit 1179 Box 9200
DPO AP 77290. Further details include that the phone is often showing no signal."
3. "Bar senior star effect **commercial** room able. Beat blood beyond whatever. Day special thing able.
Weight bad that. Measure poor push role certain. The user, LIC8208114, reported an issue with their phone. The problem is described as Call drops intermittently, and the issue occurs at 99796 Kennedy Courts
Zacharytown, VT 28689. Further details include that the phone is often showing no signal."

running with no predefined as removed and adding custom recognizers:

```
emp_patterns = [Pattern(name="emp_id", regex=r'EMP\d{5}', score=0.85)]
emp_recognizer = PatternRecognizer(supported_entity="EMP_ID", patterns=emp_patterns)
```

and anonymize function:
```
def anonymize_text(text: str) -> str:
    if text:
        results = analyzer.analyze(text=text, language='en')
        anonymized_results = anonymizer.anonymize(
            text=text,
            analyzer_results=results,
            operators={
                "EMAIL_ADDRESS": OperatorConfig("replace", {"new_value": "<ANONYMIZED_EMAIL>"}),
[...]

PR #1586, opened 2025-05-11
Country specific default recognizers to enabled=false

omri374:
## Change Description

Presidio's default recognizers list is constantly growing, and each additional recognizer can bring to false positive results. 
This PR suggests to disable country specific recognizers for English (aside from US and a few UK with strong regex patterns) to reduce false positives. If users are interested in country specific recognizers, they can always re-enable them through the configuration or through code.

To re-enable them, either change the [default YAML](https://github.com/microsoft/presidio/blob/main/presidio-analyzer/presidio_analyzer/conf/default_recognizers.yaml) to have them as `enabled: true`, or via code, add them to the recognizer registry manually. 
  - Yaml based: see more here: [YAML based configuration](https://microsoft.github.io/presidio/analyzer/analyzer_engine_provider/).
  - Code based:
    ```py
    from presidio_analyzer.predefined_recognizers import AuAbnRecognizer

    # Initialize an analyzer engine with the recognizer registry
    analyzer = AnalyzerEngine()

    # Create an instance of the AuAbnRecognizer
    au_abn_recognizer = AuAbnRecognizer()

    # Add the recognizer to the registry
    analyzer.registry.add_recognizer(au_abn_recognizer)
    ```

## Issue reference

This PR fixes issue #1526

### DIFF
--- a/presidio-analyzer/tests/test_recognizer_registry.py
+++ b/presidio-analyzer/tests/test_recognizer_registry.py
@@ -52,8 +52,7 @@ def test_when_get_recognizers_then_all_recognizers_returned(mock_recognizer_regi
     registry = mock_recognizer_registry
     registry.load_predefined_recognizers()
     recognizers = registry.get_recognizers(language="en", all_fields=True)
-    for rec in recognizers:
-        print(rec.name)
+    
     # 1 custom recognizer in english + 27 predefined - 11 disabled
     assert len(recognizers) == 1 + 27 - 11
 
--- a/presidio-analyzer/tests/test_recognizer_registry.py
+++ b/presidio-analyzer/tests/test_recognizer_registry.py
@@ -52,8 +52,10 @@ def test_when_get_recognizers_then_all_recognizers_returned(mock_recognizer_regi
     registry = mock_recognizer_registry
     registry.load_predefined_recognizers()
     recognizers = registry.get_recognizers(language="en", all_fields=True)
-    # 1 custom recognizer in english + 27 predefined
-    assert len(recognizers) == 1 + 27
+    for rec in recognizers:
+        print(rec.name)
+    # 1 custom recognizer in english + 27 predefined - 11 disabled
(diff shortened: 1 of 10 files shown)


SharonHart (dismissed):

presidio-analyzer/tests/test_recognizer_registry.py
```diff
@@ -52,8 +52,10 @@ def test_when_get_recognizers_then_all_recognizers_returned(mock_recognizer_regi
     registry = mock_recognizer_registry
     registry.load_predefined_recognizers()
     recognizers = registry.get_recognizers(language="en", all_fields=True)
-    # 1 custom recognizer in english + 27 predefined
-    assert len(recognizers) == 1 + 27
+    for rec in recognizers:
+        print(rec.name)
```
ShakutaiGit, line 56: this print is needed ? 
if we want to validate the name maybe we should do it this way  : 

assert all(getattr(rec, "name", None) for rec in recognizers), "Some recognizers are missing a .name"

ShakutaiGit (approved):
LGTM 
Left a minor comments

presidio-analyzer/tests/test_stanza_nlp_engine.py
```diff
@@ -15,6 +15,7 @@ def tags_equal(act, exp):
 
 @pytest.fixture(scope="module")
 def stanza_pipeline():
+    pytest.importorskip("stanza")
     import stanza
```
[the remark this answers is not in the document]
omri374: Some capabilities assume an extra package is required. We want tests for these capabilities to only run when the extra package is available. In our CI, all extras are installed so all tests run, but for local installations, users might not need it

presidio-analyzer/tests/test_recognizer_registry.py
omri374: nope :) good catch

ShakutaiGit (approved):

Issue #1573, opened 2025-04-28
Anonymizer does not merge entities of same type separated by whitespace if the analyzer result is not sorted by start position


mkh1991:
**Describe the bug**
If for some reason the Analyzer result `List[RecognizerResult]` is not sorted by `start` (which probably is the actual bug and has happened to me for a few cases), even without any conflicts, the Anonymizer does not merge entities correctly. 

**To Reproduce**
```python3
from presidio_analyzer import RecognizerResult
from presidio_anonymizer import AnonymizerEngine

text = "John Doe attended the meeting"

analyzer_results = [
    RecognizerResult(start=5, end=8, entity_type="PERSON", score=0.9),
    RecognizerResult(start=0, end=4, entity_type="PERSON", score=0.9),
]
anonymized_result = anonymizer.anonymize(text=text, analyzer_results=analyzer_results)
print(anonymized_result)
```

**Output**
```
text: <PERSON> <PERSON> attended the meeting
items:
[
    {'start': 9, 'end': 17, 'entity_type': 'PERSON', 'text': '<PERSON>', 'operator': 'replace'},
    {'start': 0, 'end': 8, 'entity_type': 'PERSON', 'text': '<PERSON>', 'operator': 'replace'}
]
```

**Expected behavior**
```
text: <PERSON> attended the meeting
items:
[
    {'start': 0, 'end': 8, 'entity_type': 'PERSON', 'text': '<PERSON>', 'operator': 'replace'}
]
```

If instead I reverse the analyzer results (thus sorted in this example):
```python3
anonymized_result = anonymizer.anonymize(text=text, analyzer_results=analyzer_results[::-1])
print(anonymized_result)
```

I get the expected result

**Installed version**
 I have `2.2.358` installed for both the analyzer and anonymizer

**If this is occasionally expected behaviour (that the analyzer results are not guaranteed to be sorted by start), then it should be clearer in the docs** -- I couldn't find a mention of it except for the HuggingFace Demo sorting the results before anonymizing it at https://huggingface.co/spaces/presidio/presidio_demo/blob/main/presidio_helpers.py

mkh1991:
As a temporary workaround for now, I've added

```python3
analyzer_results = sorted(analyzer_results, key = lambda x: (x.start, x.end))
```

before sending it to the Anonymizer. Interestingly the Anonymizer modifies the original Analyzer result object in-place, maybe it would be better to work with a copy here?

omri374:
[...]

PR #1588, opened 2025-05-11
Fix: Anonymizer should sort analyzer results input by start and end for whitespace merging to work

mkh1991:
## Change Description

- Sort the input `List[RecognizerResult]` object by `RecognizerResult.start` and `RecognizerResult.end`
- Without this, for unsorted input, the expected merging of two entities of the same type separated by whitespace will not occur
- Ideally I think the analyzer should also have this check, and always return results sorted by start and end

## Issue reference

- This PR fixes the main issue described in #1573

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
   
 ### Anonymizer
 - Fix mutation of input analyzer results object in Anonymizer by creating a copy first
+- Fix whitespace merging not occurring for unsorted (by start, end) analyzer results input by sorting it first
   
 ### Image Redactor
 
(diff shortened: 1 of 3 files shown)


omri374 (approved):
Great stuff!!

omri374:
can we merge this one with the other? there seems to be a duplication of some of the logic, right?

mkh1991:
@omri374 there isn't duplication exactlyy, but I branched this PR off of #1587 (and is current with main) just to keep any conflict resolution up to date, because the in-place sort by start, end depends on the creation of the copy first. I think once the previous one is merged, this diff will become clear. Alternatively, you can approve/merge only this one that has all the commits. Let me know if you prefer doing it a different way!

omri374:
Merged! thanks. On the other one, unless it has something additional about it, let's close it.

PR #1667, opened 2025-07-22
feat(presidio-analyzer): add de, es, fr langauges

gerardduch-contentsquare:


gerardduch-contentsquare:
Sorry I opened it again

PR #1722, opened 2025-09-10
Optimize Docker Images

tamirkamara:
## Change Description

A few changes to optimize docker images:
1. Use the right repo for build cache
2. Add dockerignore to remove things we don't need in the images, mainly the tests folder
3. Save layers by running apt install once
4. Use updated syntax for CMD command.
5. Remove NAME build arg

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -176,7 +176,7 @@ jobs:
             --push \
             --tag ${{ env.REGISTRY_NAME }}/${{ matrix.image }}:${{ env.TAG }}-${PLATFORM_TAG} \
             --build-arg NAME=${{ matrix.image }} \
-            --cache-from type=registry,ref=${{ env.REGISTRY_NAME }}/${{ matrix.image }}:latest \
+            --cache-from type=registry,ref=${{ env.REGISTRY_NAME }}/public/${{ matrix.image }}:latest \
             --cache-to type=inline \
             ./${{ matrix.image }}
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -337,17 +337,17 @@ jobs:
           # Build all images locally without pushing
           docker compose -f docker-compose.yml build
         env:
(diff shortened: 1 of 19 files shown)


.github/workflows/ci.yml
```diff
@@ -175,8 +175,7 @@ jobs:
             --platform ${{ matrix.platform }} \
             --push \
             --tag ${{ env.REGISTRY_NAME }}/${{ matrix.image }}:${{ env.TAG }}-${PLATFORM_TAG} \
-            --build-arg NAME=${{ matrix.image }} \
-            --cache-from type=registry,ref=${{ env.REGISTRY_NAME }}/${{ matrix.image }}:latest \
+            --cache-from type=registry,ref=${{ env.REGISTRY_NAME }}/public/${{ matrix.image }}:latest \
```
SharonHart, line 178: what will happen in cases when MCR is not accessible, like if the build is on an airgapped machine?

presidio-image-redactor/Dockerfile
```diff
@@ -13,29 +12,20 @@ ENV NLP_CONF_FILE=${NLP_CONF_FILE}
...
-COPY ./pyproject.toml /usr/bin/${NAME}/
-COPY ./README.md /usr/bin/${NAME}/
+COPY ./pyproject.toml /app/
 RUN pip install poetry && poetry install --no-root --only=main -E server
 
-COPY . /usr/bin/${NAME}/
+COPY . /app/
 EXPOSE ${PORT}
-CMD poetry run gunicorn -w $WORKERS -b 0.0.0.0:$PORT 'app:create_app()'
+CMD ["./entrypoint.sh"]
```
SharonHart, line 31: why the extraction?

.github/workflows/ci.yml
tamirkamara: Cache errors don't fail the build

SharonHart (approved):

PR #1675, opened 2025-07-27
Added Korean Resident Registration Number (RRN) recognizer (KrRrnRecognizer)

siwoo-jung:
## Change Description

- A medium regex pattern with the score of 0.5 was added.
- RRNs issued prior to October 2020 follow certain rules with checksum, so `validate_result` additionally checks those rules.

## Issue reference

This PR fixes issue #XX

## Checklist

- [X] I have reviewed the [contribution guidelines](https://github.com/microsoft/presidio/blob/main/CONTRIBUTING.md)
- [X] I have signed the CLA (if required)
- [X] My code includes unit tests
- [X] All unit tests and lint checks pass locally
- [X] My PR contains documentation updates / additions if required

PR #1742, opened 2025-10-05
fix(analyzer): update Korean language code from 'kr' to 'ko'

heesng-jung:
## Change Description

Correction of incorrect language code in the register added by https://github.com/microsoft/presidio/pull/1675 based on [ISO 639](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes). (`kr` -> `ko`)

This is a minor bug, but if you add a Korean nlp model (e.g., spacy) in the future, you may have a bug in the load.

## Checklist

### DIFF
--- a/presidio-analyzer/presidio_analyzer/conf/default_recognizers.yaml
+++ b/presidio-analyzer/presidio_analyzer/conf/default_recognizers.yaml
@@ -156,6 +156,7 @@ recognizers:
   - name: KrRrnRecognizer
     supported_languages: 
     - ko
+    - kr
     type: predefined
     enabled: false
 


heesng-jung:

presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/korea/kr_rrn_recognizer.py
```diff
@@ -53,7 +53,7 @@ def __init__(
         self,
         patterns: Optional[List[Pattern]] = None,
         context: Optional[List[str]] = None,
-        supported_language: str = "kr",
+        supported_language: str = "ko",
```
omri374, line 56: How about keeping both? This would mean it does break things for those already using kr instead of ko.

heesng-jung: I didn't think of compatibility with the existing code, but thank you for pointing it out.
First of all, I reflected it in https://github.com/microsoft/presidio/pull/1742/commits/70bef519a07211cf8a93fddf1404f53e9cb14862 you reviewed.
One thing I can only point out is that `kr` is [the official language code of the Kanuri language](https://www.loc.gov/standards/iso639-2/php/langcodes_name.php?iso_639_1=kr), although it is a less used language.
Is it okay to make both `ko` and `kr` compatible like the modified config?

omri374: Thanks for pointing this out. I would suggest to keep both and deal with the issue with Kanuri when we get to it. It is more likely that Presidio users define `kr` as the language.

omri374 (approved):

PR #1747, opened 2025-10-19
Add dependency lock files

tamirkamara:
## Change Description

Dependency lock files are used to secure supply chains, and presidio is able to benefit from this. I added a few and tested under all supported python versions.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -374,7 +374,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
-          python-version: '3.x'
+          python-version: '3.10'
 
       - name: Cache E2E dependencies
         uses: actions/cache@v4


omri374:
@tamirkamara thanks. How are you thinking about the locking process? to be done locally for every PR? we had issues with this in the past. It could be that we would need locking to happen as part of a CI step and pushed to the branch

tamirkamara:
> How are you thinking about the locking process? to be done locally for every PR? we had issues with this in the past. It could be that we would need locking to happen as part of a CI step and pushed to the branch

@omri374 I believe this should be the responsibility of the person who added / updated the dependencies, so yes - locally. I can include a modified script I used to do this across the different sub projects.

tamirkamara:
@SharonHart @omri374
Thinking about this again - the PR essentially pins the entire dependency tree to specific versions. While I think this creates a time-stable and reproducible package (beyond the security benefits), it will make dependabot far more active than it is today.
IMO, it's worth the added complexity and time.

PR #1758, opened 2025-10-21
fix release-docs workflow credentials

tamirkamara:
## Change Description

mkdocs needs credentials to push to the pages branch. One method is to persist the GitHub token from the checkout step (which we remove everywhere else).

## Checklist

### DIFF
--- a/.github/workflows/release-docs.yml
+++ b/.github/workflows/release-docs.yml
@@ -19,15 +19,17 @@ jobs:
         uses: actions/checkout@v5
         with:
           fetch-depth: 0  # Fetch all history for proper gh-pages deployment
-          persist-credentials: false
+          persist-credentials: true  # So that the token is available for pushing
       
       - name: Set up Python
         uses: actions/setup-python@v6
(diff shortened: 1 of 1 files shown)


.github/workflows/release-docs.yml
```diff
@@ -19,15 +19,17 @@ jobs:
...
         with:
           fetch-depth: 0  # Fetch all history for proper gh-pages deployment
-          persist-credentials: false
+          persist-credentials: true  # So that the token is available for pushing
       
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
           python-version: '3.x'
+
```
SharonHart, line 28: How are the dependencies installed then? Is the mkdocs command looking for requirements file in docs/ ?

SharonHart (approved):

.github/workflows/release-docs.yml
tamirkamara: It's blow, just split to another step.

PR #1763, opened 2025-10-23
AHDS authentication security with ChainedTokenCredential and add CodeQL suppression

dorlugasigal:
## Change Description
our previous env check change was lacking a suppression for the codeql warning,
added that inline

## Issue reference

Fixes #XX

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -198,6 +198,8 @@ jobs:
             --tag ${{ env.REGISTRY_NAME }}/${{ matrix.image }}:${{ env.TAG }}-${PLATFORM_TAG} \
             --cache-from type=registry,ref=${{ env.REGISTRY_NAME }}/public/${{ matrix.image }}:latest \
             --cache-to type=inline \
+            --attest type=sbom \
+            --attest type=provenance,mode=max \
             ./${{ matrix.image }}
         env:
           REGISTRY_NAME: ${{ env.REGISTRY_NAME }}
--- a/.github/workflows/defender-for-devops.yml
+++ b/.github/workflows/defender-for-devops.yml
@@ -48,7 +48,7 @@ jobs:
       uses: microsoft/security-devops-action@latest
       id: msdo
       env:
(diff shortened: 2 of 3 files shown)


presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/ahds_recognizer.py
```diff
@@ -68,7 +68,7 @@ def __init__(
             if os.getenv('ENV') == 'production':
```
tamirkamara: I don't remember this version... 
Shouldn't the more secure option be the default, i.e. prod is the default and if you want to use dev you need to add an env var?

dorlugasigal: Right, having prod as the default would be more least-privileged approach, but DefaultAzureCredential keeps the dev experience better, so devs wont need to specify an env var in their work env.

Wdyt?

omri374: Maybe the docs should say something about this.

tamirkamara: Yes, it should.
I think the default should be secured, meaning handle the production case without any settings.

SharonHart: yet this is a breaking change @tamirkamara

tamirkamara: Breaking change for a developer which might be fine if we want to prioritize security in production.

dorlugasigal: I modified to use `ChainedTokenCredential` by default, and if `env` set to `dev` then we use `defaultAzureCredentials`
and also added relevant details in the docs

PR #1775, opened 2025-11-02
Language models integration (LangExtract)

RonShakutai:
## Change Description

Added language model-based PII/PHI detection using LangExtract library with Ollama backend. This feature introduces a three-tier recognizer hierarchy that enables flexible, prompt-driven entity recognition with configurable few-shot examples and entity mappings. The implementation supports local LLM deployment via Ollama for privacy-sensitive environments.

## Architecture:

- __`LMRecognizer`__: Abstract base class for all language model-based recognizers
- __`LangExtractRecognizer`__: Model-agnostic LangExtract integration layer
- __`OllamaLangExtractRecognizer`__: Concrete Ollama implementation with server validation

## Key Additions:

- __LM Recognizer hierarchy__: `LMRecognizer` → `LangExtractRecognizer` → `OllamaLangExtractRecognizer`
- __Configuration system__: YAML configs for prompts, few-shot examples, and entity mappings
- __Docker integration__: Ollama service in docker-compose with health checks
- __Testing__: Unit tests with mocking + E2E integration tests
- __Documentation__: Usage guide with setup instructions and troubleshooting

## Implementation Details:

- Supports local LLM deployment via Ollama (default: gemma2:2b)
- Configurable entity detection (13+ entity types)
- Disabled by default in `default_recognizers.yaml` (opt-in feature)
- Extensible architecture for future LM provider integrations

## Issue reference

Fixes #XX

## Checklist

- [v] I have reviewed the [contribution guidelines](https://github.com/microsoft/presidio/blob/main/CONTRIBUTING.md)
- [v] I have signed the CLA (if required)
- [v] My code includes unit tests
- [v] All unit tests and lint checks pass locally
- [v] My PR contains documentation updates / additions if required

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,7 +163,7 @@ jobs:
           MINIMUM_GREEN: 85
           MINIMUM_ORANGE: 70
           COVERAGE_DATA_BRANCH: coverage-data
-          SUBPROJECT_ID: ${{ matrix.component.name }}
+          SUBPROJECT_ID: ${{ matrix.component.path }}
           COVERAGE_PATH: ${{ matrix.component.path }}
 
       - name: Build wheel package


presidio-analyzer/presidio_analyzer/conf/langextract_prompts/default_pii_examples.yaml
```diff
@@ -0,0 +1,96 @@
+# LangExtract PII Extraction Examples for Presidio
+# These examples guide the model on how to extract PII entities
+
+examples:
+  # Example 1: Basic contact information
+  - text: "John Doe lives at 123 Main Street, New York, NY and his phone is 555-123-4567"
+    extractions:
+      - extraction_class: "person"
```
omri374: Would it work if those are all caps, like the other Presidio entities? It could make the mapping later easier.

presidio-analyzer/presidio_analyzer/conf/langextract_prompts/default_pii_prompt.txt
```diff
@@ -0,0 +1,29 @@
...
+
+Focus on extracting:
+- Person names (full names, first names, last names, titles)
+- Phone numbers (mobile, landline, international formats)
+- Email addresses
+- Physical addresses (street addresses, cities, states, countries)
+- Social security numbers
+- Credit card numbers
+- Dates of birth and other temporal information
+- Medical record numbers
```
omri374: Should we extend this to PHI? Personal medical information?

presidio-analyzer/presidio_analyzer/conf/langextract_config_e2e.yaml
```diff
@@ -0,0 +1,44 @@
+# LangExtract LLM-based PII Recognition - E2E Test Configuration
```
omri374: Unrelated to e2e tests

presidio-analyzer/presidio_analyzer/conf/langextract_config_e2e.yaml
```diff
@@ -0,0 +1,44 @@
...
+    organization: ORGANIZATION
+    phone: PHONE_NUMBER
+    email: EMAIL_ADDRESS
+    date: DATE_TIME
+    ssn: US_SSN
+    credit_card: CREDIT_CARD
+    medical_record: MEDICAL_LICENSE
+    ip_address: IP_ADDRESS
+    url: URL
+    iban: IBAN_CODE
```
omri374: how is it treating entities that are not mapped here? Should we let it detect entities not in the initial list?

omri374: Consider having the list of entities in the prompt being taken from here. This would not require two sets of entity lists.

docs/analyzer/adding_recognizers.md
```diff
@@ -165,6 +165,12 @@ On how to integrate Presidio with AHDS De-Identification Protected Health Inform
 and a sample for a ADHS Remote Recognizer, refer to the
 [AHDS de-Identification Integration document](../samples/python/ahds/index.md).
 
+### LangExtract LLM-based recognizer
```
omri374: Generally speaking, the big change here is the integration of LLMs. LangExtract is an implementation detail. Consider positioning this as an LLM-based recognizer, and mention that it's leveraging LangExtract under the hood. Most people don't know LangExtract and it might confuse them.

docs/samples/python/langextract/index.md
```diff
@@ -0,0 +1,72 @@
+# LangExtract LLM-based PII Detection Integration
```
omri374: Same here, it could be "LLM-based PII Detection using LangExtract"

e2e-tests/tests/test_analyzer.py
```diff
@@ -572,3 +572,19 @@ def test_given_regex_flags_and_normal_entities_are_returned():
     assert equal_json_strings(
         expected_response, response_content
     )
+
+
+@pytest.mark.api
+def test_given_langextract_enabled_then_detects_pii_with_llm():
```
omri374: This is a very generic test which can pass even if LangExtract failed or didn't return anything. Is it needed? If yes, maybe we can use the analysis_explanation to extract the recognizer which provided the result?

presidio-analyzer/presidio_analyzer/conf/e2e_recognizers.yaml
```diff
@@ -0,0 +1,202 @@
+supported_languages:
```
omri374, line 1: should this be in conf or test/conf?

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/langextract_recognizer.py
```diff
@@ -0,0 +1,364 @@
...
+from presidio_analyzer import AnalysisExplanation, RecognizerResult, RemoteRecognizer
+from presidio_analyzer.nlp_engine import NlpArtifacts
+
+logger = logging.getLogger("presidio-analyzer")
+
+LANGEXTRACT_DOCS_URL = "https://github.com/google/langextract"
+OLLAMA_INSTALL_DOCS = "https://github.com/google/langextract#using-local-llms-with-ollama"
+
+
+class LangExtractRecognizer(RemoteRecognizer):
```
omri374: Will the other langextract recognizers inherit from this one? Consider having a base LangExtract recognizer and have ollama and openai/custom inherit from it. Otherwise we might need to change the name which is not backward compatible.

RonShakutai: exactly what i thought

e2e-tests/tests/test_analyzer.py
RonShakutai: Will do.

presidio-analyzer/presidio_analyzer/conf/langextract_prompts/default_pii_examples.yaml
RonShakutai: the only potential issue is that  if LangExtract has its own entity taxonomy, 
Currently ive added another mapping to alignt with presidio mapping.
[langextract mapping](https://github.com/microsoft/presidio/pull/1775/files#diff-c7da646f13d1a6fa356316fb830ce18b15ebdea1b5227d2a06ae37400ff2bca7R17)

presidio-analyzer/presidio_analyzer/conf/langextract_config_e2e.yaml
RonShakutai: > how is it treating entities that are not mapped here? Should we let it detect entities not in the initial list?

LangExtract defines its own entity names — right now, entities not explicitly mapped here are dropped.
This is a delicate point, since variations like email, email_address, or email-address might appear under different labels.
We may need to expand our mappings or handle alias matching to ensure these cases aren’t missed.

RonShakutai:
Hey @omri374 and @SharonHart ,
It looks like the unit-test matrix is making the runners run out of disk space. 
This will probably keep happening as we add more tests.
I believe this isn’t an issue with the PR tests itself but to the number of tests added.

I think we have a few options:
- Increase the runner’s disk size
- Split the matrix into separate jobs so each worker uses less space
- Add cleanup steps to reduce disk usage
- Make the matrix run in seq order.

Let’s sync on this on Sunday.

SharonHart:
> Hey @omri374 and @SharonHart , It looks like the unit-test matrix is making the runners run out of disk space. This will probably keep happening as we add more tests. I believe this isn’t an issue with the PR tests itself but to the number of tests added.
> 
> I think we have a few options:
> 
> * Increase the runner’s disk size
> * Split the matrix into separate jobs so each worker uses less space
> * Add cleanup steps to reduce disk usage
> * Make the matrix run in seq order.
> 
> Let’s sync on this on Sunday.

FYI @tamirkamara 

I don't think that this has to do with the matrix strategy, since every record within it has its own seperate job, therefore seperate runner machines. 
As to the mitigations, you can't increase the non-self-hosted runners. 
Each job starts from scratch so I do not know if cleanups would help

Overall, would be good to investigate what's changed in this PR, is only one specific matrix record fails or it is soporadic? where is the space running out in?

RonShakutai:
> > Hey @omri374 and @SharonHart , It looks like the unit-test matrix is making the runners run out of disk space. This will probably keep happening as we add more tests. I believe this isn’t an issue with the PR tests itself but to the number of tests added.
> > I think we have a few options:
> > 
> > * Increase the runner’s disk size
> > * Split the matrix into separate jobs so each worker uses less space
> > * Add cleanup steps to reduce disk usage
> > * Make the matrix run in seq order.
> > 
> > Let’s sync on this on Sunday.
> 
> FYI @tamirkamara
> 
> I don't think that this has to do with the matrix strategy, since every record within it has its own seperate job, therefore seperate runner machines. As to the mitigations, you can't increase the non-self-hosted runners. Each job starts from scratch so I do not know if cleanups would help
> 
> Overall, would be good to investigate what's changed in this PR, is only one specific matrix record fails or it is soporadic? where is the space running out in?

solved in this PR https://github.com/microsoft/presidio/pull/1784

presidio-analyzer/presidio_analyzer/conf/ollama_config.yaml
```diff
@@ -0,0 +1,52 @@
...
+    medical_record: MEDICAL_LICENSE
+    health_insurance: MEDICAL_LICENSE
+    medical_license: MEDICAL_LICENSE
+    ip_address: IP_ADDRESS
+    url: URL
+    iban: IBAN_CODE
+    blood_type: NRP
+    condition: NRP
+    medication: NRP
+    procedure: NRP
```
SharonHart: I would drop those from the default options, in some cases they are also not PIIs
Also, its not really NRP

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/langextract_recognizer.py
```diff
@@ -0,0 +1,411 @@
...
+    LANGEXTRACT_AVAILABLE = False
+    lx = None
+
+from presidio_analyzer import AnalysisExplanation, RecognizerResult
+from presidio_analyzer.lm_recognizer import LMRecognizer
+from presidio_analyzer.nlp_engine import NlpArtifacts
+
+logger = logging.getLogger("presidio-analyzer")
+
+LANGEXTRACT_DOCS_URL = "https://github.com/google/langextract"
```
SharonHart: drop

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/langextract_recognizer.py
```diff
@@ -0,0 +1,411 @@
...
+
+        if not langextract_config:
+            raise ValueError(
+                "Configuration file must contain 'langextract' section "
+                "with shared config"
+            )
+
+        return langextract_config
+
+    def _load_config(
```
SharonHart: why another load method?

e2e-tests/tests/test_package_e2e_integration_flows.py
```diff
@@ -0,0 +1,244 @@
...
+    analyzer = AnalyzerEngine(
+        nlp_engine=nlp_engine,
+        supported_languages=["en"]
+    )
+    analyzer.registry.add_recognizer(ollama_recognizer)
+
+    # Analyze text - just verify we get results (from any recognizer)
+    results = analyzer.analyze(text_to_test, language="en")
+
+    # Verify at least some entities were detected (by Ollama or other recognizers)
```
SharonHart: How is it different than other tests? What does it test besides instantiation of ollama recognizer? 
We need to test e2e integration with langextract specifically, I would check the response for specific result from langextract (or subclasses), or instnatiate a analyzer engine with just langextract

e2e-tests/resources/ollama_test_config.yaml
```diff
@@ -0,0 +1,46 @@
...
+    email: EMAIL_ADDRESS
+    date: DATE_TIME
+    ssn: US_SSN
+    identification_number: US_SSN
+    credit_card: CREDIT_CARD
+    medical_record: MEDICAL_LICENSE
+    ip_address: IP_ADDRESS
+    url: URL
+    iban: IBAN_CODE
+    payment_status: IGNORE
```
omri374: Why do we need to explicitly ignore entities? Do we get limited generalizability from LangExtract?

e2e-tests/tests/test_package_e2e_integration_flows.py
RonShakutai: Validation for the specific recognizer added.
Thanks.

e2e-tests/tests/conftest.py
```diff
@@ -0,0 +1,37 @@
...
+        help="Run Ollama integration tests (requires Ollama server running)"
+    )
+
+
+def pytest_configure(config):
+    """Configure pytest with custom markers and settings."""
+    config.addinivalue_line(
+        "markers",
+        "ollama: mark test as requiring Ollama server"
+    )
```
SharonHart: 1. Why do we need too add the marker and cli option? Can't it be done with annotating with @pytest.mark + running with `pytest -m` 
2. Why is it optional, if e2e should start the whole cluster anyway? Either way, locally if you do not want it to run it can fail anyway

e2e-tests/tests/test_package_e2e_integration_flows.py
```diff
@@ -0,0 +1,123 @@
...
+            
+            langextract_detected_at_least_one |= (
+                recognizer_name == "Ollama LangExtract PII"
+            )
+    
+    # Verify that Ollama LangExtract recognizer participated in detection
+    assert langextract_detected_at_least_one, \
+        f"Expected 'Ollama LangExtract PII' recognizer to detect at least one entity. Recognizers used: {recognizers_used}"
+
+    # Anonymize the detected entities
```
SharonHart: if it tests just the integration with llm, you can drop the anonymization check

e2e-tests/tests/conftest.py
RonShakutai: fixed.

e2e-tests/tests/test_package_e2e_integration_flows.py
RonShakutai: fixed.

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/ollama_langextract_recognizer.py
```diff
@@ -0,0 +1,91 @@
...
+            }
+
+            if self.temperature is not None:
+                extract_params["temperature"] = self.temperature
+
+            extract_params.update(kwargs)
+
+            return lx.extract(**extract_params)
+        except Exception:
+            logger.exception(
```
odedfis, line 69: simplify exception

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/ollama_langextract_recognizer.py
```diff
@@ -0,0 +1,91 @@
...
+        try:
+            extract_params = {
+                "text_or_documents": text,
+                "prompt_description": prompt,
+                "examples": examples,
+                "model_id": self.model_id,
+                "model_url": self.model_url,
+            }
+
+            if self.temperature is not None:
```
odedfis: should be in upper class

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/langextract_recognizer.py
```diff
@@ -0,0 +1,189 @@
...
+    def _call_llm(self, text: str, entities: List[str], **kwargs):
+        """Call LangExtract LLM and convert to RecognizerResult."""
+        langextract_result = self._call_langextract(
+            text=text,
+            prompt=self.prompt_description,
+            examples=self.examples
+        )
+
+        results = []
+        if not langextract_result or not langextract_result.extractions:
```
odedfis: export conversion from langextarct output to presidio output to langextract helper

presidio-analyzer/presidio_analyzer/llm_utils/langextract_helper.py
```diff
@@ -0,0 +1,84 @@
...
+# Default alignment score mappings for LangExtract extractions
+DEFAULT_ALIGNMENT_SCORES = {
+    "MATCH_EXACT": 0.95,
+    "MATCH_FUZZY": 0.80,
+    "MATCH_LESSER": 0.70,
+    "NOT_ALIGNED": 0.60,
+}
+
+
+def get_langextract_module():
```
odedfis: for which case is this necessary?

RonShakutai: since we are in third party recognizers the dependencies will not automatically installed, 
We need to ensure exception will not be thrown.

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/ollama_langextract_recognizer.py
RonShakutai: fixed.

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/ollama_langextract_recognizer.py
RonShakutai: fixed.

docs/samples/python/langextract/index.md
```diff
@@ -0,0 +1,116 @@
...
+## Prerequisites
+
+### Setting up Ollama
+
+You have two options to set up Ollama:
+
+**Option 1: Docker Compose** (recommended)
+```bash
+# Start Ollama service
+docker compose up -d ollama
```
omri374: Please be very explicit about this phase:
- Where should it be run from (the root presidio path?)
- Install Docker if I don't have it
- Differences between Windows/Mac/Linux if there are any

omri374: Also consider specifying that this would install the gemma2:2b model and that it is ~3Gb

docs/samples/python/langextract/index.md
```diff
@@ -0,0 +1,116 @@
...
+1. Install with langextract support and set up Ollama (see Prerequisites above):
+   ```sh
+   pip install presidio-analyzer[langextract]
+   ```
+
+2. Enable the recognizer in `default_recognizers.yaml`:
+   ```yaml
+   - name: OllamaLangExtractRecognizer
+     enabled: true  # Change from false to true
+   ```
```
omri374: Consider being more explicit here - what should the user do after changing the yaml file? If the user pip installs presidio, they don't even know that there is this yaml file. If they clone, they still might be using the installed package and changing it would have no effect. My suggestion:
1. Add a link to default_recognizers.yaml on the repo (like you show with the custom yaml below)
2. Add a short python snippet that shows how to load this config with ollama.

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/langextract_recognizer.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+class LangExtractRecognizer(LMRecognizer, ABC):
+    """
+    Base class for LangExtract-based PII recognizers.
+
+    Subclasses implement _call_langextract() for specific LLM providers.
+    """
+
+    def __init__(self, config_path: str, name: str = "LangExtract LLM PII"):
```
omri374: Recognizers should accept the full list of inputs for `EntityRecognizer.__init__`, otherwise the yaml loading fails.

```py
    def __init__(
        self,
        supported_entities: List[str],
        name: str = None,
        supported_language: str = "en",
        version: str = "0.0.1",
```
For this input
```yaml
...recognizers:
  - name: OllamaLangExtractRecognizer
    supported_languages:
      - en
    type: predefined
    enabled: true
    config_path: presidio_analyzer/conf/langextract_config_ollama.yaml
...
```
I got this:
`TypeError: OllamaLangExtractRecognizer.__init__() got an unexpected keyword argument 'supported_language'`

omri374: also consider adding `**kwargs` like you did in LMRecognizer to allow users to pass more parameters to control the instance.

omri374: Having said that, supported_entities should not be used as it's read from the config. In this case I would suggest to raise an error if someone passes supported_entities (but not remove it as it's part of the interface for recognizers)

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/ollama_langextract_recognizer.py
```diff
@@ -0,0 +1,67 @@
...
+        during initialization. Any connectivity or model issues will be reported
+        when analyze() is first called.
+        """
+        # Determine actual config path
+        actual_config_path = (
+            config_path if config_path else str(self.DEFAULT_CONFIG_PATH)
+        )
+
+        try:
+            super().__init__(
```
omri374: It's better not to put super in try/except. Consider raising the error within the base class init.

presidio-analyzer/presidio_analyzer/conf/default_recognizers.yaml
```diff
@@ -199,3 +199,10 @@ recognizers:
     - en
     type: predefined
     enabled: false
+
+  - name: OllamaLangExtractRecognizer
+    supported_languages:
+      - en
+    type: predefined
+    enabled: false
+    config_path: presidio_analyzer/conf/langextract_config_ollama.yaml
```
omri374: This path didn't work for me. I think it should be appended to the root of the package to form the absolute path.

presidio-analyzer/presidio_analyzer/lm_recognizer.py
```diff
@@ -0,0 +1,161 @@
...
+    ):
+        """Initialize LM recognizer.
+
+        :param supported_entities: Entity types to detect.
+        :param labels_to_ignore: Entity labels to skip.
+        :param enable_generic_consolidation: Consolidate unknown
+            entities to GENERIC_PII_ENTITY.
+        """
+        super().__init__(
+            supported_entities=supported_entities or [],
```
omri374: I would suggest to raise an expception if the user tries to pass a list of entities here, as they are overridden below anyway.

presidio-analyzer/presidio_analyzer/llm_utils/langextract_helper.py
```diff
@@ -0,0 +1,161 @@
...
+
+def convert_langextract_to_presidio_results(
+    langextract_result,
+    entity_mappings: Dict,
+    supported_entities: List[str],
+    enable_generic_consolidation: bool,
+    recognizer_name: str,
+    alignment_scores: Optional[Dict[str, float]] = None
+) -> List[RecognizerResult]:
+    """Convert LangExtract results to Presidio format."""
```
omri374: Add full docstring or change to private/protected

presidio-analyzer/presidio_analyzer/llm_utils/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""Utilities for LLM-based recognizers."""
```
omri374, line 1: Instead of a list of utility functions, can these functions be functions within `LangExtractRecognizer`?

presidio-analyzer/presidio_analyzer/llm_utils/__init__.py
```diff
@@ -0,0 +1,7 @@
...
+
+from .config_loader import *  # noqa: F403
```
omri374: There's a reason why F403 is enforced :)

presidio-analyzer/presidio_analyzer/lm_recognizer.py
```diff
@@ -0,0 +1,161 @@
...
+            if filtered_results:
+                logger.info(
+                    "LLM recognizer found %d PII entities",
+                    len(filtered_results),
+                )
+
+            return filtered_results
+
+        except Exception as e:
+            logger.exception("LLM analysis failed: %s", str(e))
```
omri374: Continuing after an exception is tricky here, as there might be a problem the user wants to know about (and might not review the logs). Can we create a few types of exceptions here maybe? If there's a problem with the LLM service (e.g. ollama doesn't have the requested model), then it should raise an exception. If it's just a parsing issue because of some weird input, it should log an exception and continue. WDYT?
Here's the list of exceptions that LangExtract exposes: https://github.com/google/langextract/blob/0c1af879cc9bc18817faad8a541247d76c14e931/langextract/exceptions.py#L15

docs/samples/python/langextract/index.md
```diff
@@ -0,0 +1,116 @@
...
+
+For the default entity mappings and examples, see the [default configuration](https://github.com/microsoft/presidio/blob/main/presidio-analyzer/presidio_analyzer/conf/langextract_config_ollama.yaml).
+
+## Prerequisites
+
+### Setting up Ollama
+
+You have two options to set up Ollama:
+
+**Option 1: Docker Compose** (recommended)
```
omri374: For Mac, the native option is recommended over Docker as it leverages the MPS (Mac GPU). I got timeouts when trying to run the docker version on Mac M2

docs/samples/python/langextract/index.md
dorlugasigal: @ShakutaiGit for changes in the readme please take the relevant readme section from my PR and extend it so we wont have large conflicts as before

docker-compose.yml
```diff
@@ -29,3 +49,6 @@ services:
       - PORT=5001
     ports:
       - "5003:5001"
+
+volumes:
```
SharonHart, line 53: I think that this one is not needed as we do not save any context between iterations. Trye to remove and see if it affects the tests / local development experience

SharonHart (dismissed):

docker-compose.yml
RonShakutai: You're correct but i think we better keep it wdyt ? 
https://geshan.com.np/blog/2025/02/ollama-docker-compose/
image

presidio-analyzer/presidio_analyzer/llm_utils/__init__.py
RonShakutai: I prefer keeping this logic external to the class for better readability and reuse.

presidio-analyzer/presidio_analyzer/llm_utils/__init__.py
RonShakutai: Fixed

presidio-analyzer/presidio_analyzer/llm_utils/langextract_helper.py
RonShakutai: fixed

presidio-analyzer/presidio_analyzer/lm_recognizer.py
RonShakutai: fixed.

presidio-analyzer/presidio_analyzer/lm_recognizer.py
RonShakutai: I think we should raise an exception as well, and keep the logger exception so the error message from LangExtract is still recorded. Since we’re not actually recovering from LangExtract exceptions only logging them I would prefer not to catch them here.

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/ollama_langextract_recognizer.py
RonShakutai: fixed.

docs/samples/python/langextract/index.md
RonShakutai: Fixed

docs/samples/python/langextract/index.md
RonShakutai: fixed.

presidio-analyzer/presidio_analyzer/conf/default_recognizers.yaml
RonShakutai: fixed

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/langextract_recognizer.py
RonShakutai: fixed.

presidio-analyzer/presidio_analyzer/lm_recognizer.py
omri374: Not sure I understand. So remove the try/except?

RonShakutai: Already removed.
Since we don’t recover from LangExtract errors here, catching them doesn’t help — it only hides the real issue.
So we log the error and let the exception propagate naturally.

SharonHart (approved):

presidio-analyzer/presidio_analyzer/llm_utils/entity_mapper.py
```diff
@@ -0,0 +1,212 @@
...
+
+    :param results: List of recognizer results to validate.
+    :return: Filtered list with only valid position ranges.
+    """
+    filtered = []
+
+    for result in results:
+        if result.start is None or result.end is None:
+            logger.warning(
+                "LLM returned result without start/end positions, skipping"
```
omri374: Add the result to the warning (or just the type)

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/ollama_langextract_recognizer.py
```diff
@@ -0,0 +1,75 @@
...
+    """LangExtract recognizer using Ollama backend."""
+
+    DEFAULT_CONFIG_PATH = (
+        Path(__file__).parent.parent.parent / "conf" / "langextract_config_ollama.yaml"
+    )
+
+    def __init__(
+        self,
+        config_path: Optional[str] = None,
+        supported_language: str = "en",
```
omri374, line 21: As discussed, make sure the recognizer holds self.supported_entities and self.supported_language (or implement get_supported_language and get_supported_entities)

docs/samples/python/langextract/index.md
```diff
@@ -0,0 +1,176 @@
+# Language Model-based PII/PHI Detection
```
omri374: As discussed, add a comment that this is still experimental

RonShakutai: done

presidio-analyzer/presidio_analyzer/llm_utils/entity_mapper.py
RonShakutai: done

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/ollama_langextract_recognizer.py
RonShakutai: done.

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/ollama_langextract_recognizer.py
```diff
@@ -0,0 +1,76 @@
...
+        :param config_path: Path to configuration file (optional).
+        :param supported_language: Language this recognizer supports
+            (optional, default: "en").
+        :param context: List of context words
+            (optional, currently not used by LLM recognizers).
+        :param kwargs: Additional keyword arguments passed from YAML configuration
+            (e.g., name, version, etc.).
+        :raises ImportError: If langextract is not installed.
+        """
+        check_langextract_available()
```
omri374: nit: this is happening twice

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/langextract_recognizer.py
```diff
@@ -0,0 +1,138 @@
...
+        )
+        self.model_id = model_config["model_id"]
+        self.temperature = model_config.get("temperature")
+
+        super().__init__(
+            supported_entities=supported_entities,
+            supported_language=supported_language,
+            name=name,
+            version="1.0.0",
+            model_id=self.model_id,
```
omri374: nit: passing self.model_id and self.temperature to the parent before initializing them makes no sense. Best case it would pass a None, worst case it would change the default if someone changed the default in the parent's code.

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/ollama_langextract_recognizer.py
```diff
@@ -0,0 +1,76 @@
...
+    DEFAULT_CONFIG_PATH = (
+        Path(__file__).parent.parent.parent / "conf" / "langextract_config_ollama.yaml"
+    )
+
+    def __init__(
+        self,
+        config_path: Optional[str] = None,
+        supported_language: str = "en",
+        context: Optional[list] = None,
+        **kwargs,
```
omri374, line 23: nit: kwargs are currently not used. Does it make sense to pass kwargs to the langextract caller? If a user wants to configure langextract differently?

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/langextract_recognizer.py
```diff
@@ -0,0 +1,138 @@
...
+        self.prompt_description = render_jinja_template(
+            prompt_template,
+            supported_entities=self.supported_entities,
+            enable_generic_consolidation=self.enable_generic_consolidation,
+            labels_to_ignore=self.labels_to_ignore,
+        )
+
+        self.entity_mappings = langextract_config["entity_mappings"]
+
+        logger.info("Loaded recognizer: %s", self.name)
```
omri374: This already happens on the `EntityRecognizer` level

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/langextract_recognizer.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def _call_llm(self, text: str, entities: List[str], **kwargs):
+        """Call LangExtract LLM."""
+        # Build extract params
+        extract_params = {
+            "text": text,
```
omri374, line 110: consider adding debug=False (or a way to specify whether debug is needed, e.g. through kwargs)

omri374 (dismissed):
Great work! Left a few comments, all are small and for you to decide if to handle or not.

SharonHart (dismissed):

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/ollama_langextract_recognizer.py
RonShakutai: Thanks for the catch.

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/langextract_recognizer.py
RonShakutai: Fixed.

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/ollama_langextract_recognizer.py
RonShakutai: This is a bug in all third-party recognizers i will open separate PR

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/langextract_recognizer.py
RonShakutai: removed.

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/langextract_recognizer.py
RonShakutai: done

omri374 (approved):

PR #1785, opened 2025-11-16
Remove poetry cache from docker images

tamirkamara:
## Change Description

Poetry keeps a cache directory with things not needed for runtime so removing the cache reduces the docker images.
Analyzer was 2.15GB and is now 1.99GB
Anonymizer was 388MB and is now 375MB

## Checklist

### DIFF
--- a/presidio-analyzer/Dockerfile
+++ b/presidio-analyzer/Dockerfile
@@ -26,7 +26,10 @@ RUN apt-get update \
 
 COPY ./pyproject.toml /app/
 
-RUN pip install poetry && poetry install --no-root --only=main -E server
+RUN pip install poetry \
+    && poetry install --no-root --only=main -E server \
+    && rm -rf $(poetry config cache-dir)
+    
 # install nlp models specified in NLP_CONF_FILE
 COPY ./install_nlp_models.py /app/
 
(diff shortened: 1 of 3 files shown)


ShakutaiGit (approved):

presidio-analyzer/Dockerfile
```diff
@@ -26,7 +26,10 @@ RUN apt-get update \
 
 COPY ./pyproject.toml /app/
 
-RUN pip install poetry && poetry install --no-root --only=main -E server
+RUN pip install poetry \
+    && poetry install --no-root --only=main -E server \
+    && rm -rf $(poetry config cache-dir)
+    
 # install nlp models specified in NLP_CONF_FILE
 COPY ./install_nlp_models.py /app/
```
omri374, line 34: @tamirkamara the install_nlp_models also does a bunch of pip installs. Maybe a better solution would be to clear cache after this step. Have you considered it?

tamirkamara: @omri374 we disable pip cache within the docker files. For example: https://github.com/microsoft/presidio/blob/main/presidio-analyzer/Dockerfile#L6

PR #1794, opened 2025-11-24
CI coverage test 

RonShakutai:
## Change Description

This PR introduces two improvements to our CI coverage workflow:

- Per-component coverage comments on every PR
Each pipeline run posts a comment that shows the coverage of the specific component being tested.
The action updates the existing comment instead of creating new ones on every run.

Example:
image

- Strict PR diff-coverage enforcement using diff-cover
We now verify coverage only for the lines changed in the PR.
If the PR reduces diff-coverage below 80%, the CI step fails.
Example of failure: 
image

Additionally, a COVERAGE_THRESHOLD env variable was added so the required percentage can be easily raised in the future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -156,7 +156,7 @@ jobs:
           fi
 
       - name: Comment PR with Coverage
-        if: github.event_name == 'pull_request' && matrix.python-version == env.PRIMARY_PYTHON
+        if: matrix.python-version == env.PRIMARY_PYTHON
         uses: py-cov-action/python-coverage-comment-action@v3
         with:
           GITHUB_TOKEN: ${{ github.token }}


.github/workflows/ci.yml
```diff
@@ -135,10 +140,30 @@ jobs:
...
+          if [ "${{ github.event_name }}" = "pull_request" ]; then
+            echo "Checking PR diff coverage (>= ${{ env.COVERAGE_THRESHOLD }}%) with diff-cover..."
+            poetry run diff-cover coverage.xml \
+              --compare-branch=origin/${{ github.base_ref }} \
+              --fail-under=${{ env.COVERAGE_THRESHOLD }} \
+              --show-uncovered
+          fi
+
+      - name: Comment PR with Coverage
+        if: github.event_name == 'pull_request' && matrix.python-version == '3.11'
```
SharonHart: take the last metrix value instead of fixed version

RonShakutai: fixed.

PR #1805, opened 2025-12-02
Fix gliner truncates text

### LINKED ISSUE #1569: GLiNER Recognizer Truncates Long Text, Leading to Poor Redaction Results
**Describe the bug**
When using **Presidio Image Redactor** with the **GLiNER Recognizer**, the model issues the following warning:
```
"UserWarning: Sentence of length 20415 has been truncated to 384
warnings.warn(f"Sentence of length {len(tokens)} has been truncated to {max_len}")
```
Due to this truncation, the redaction performance is significantly degraded, as some PII remains undetected in the untruncated portions of the text.

**To Reproduce**
1. Use a single-page image-based PDF and convert it to a PIL Image (e.g., using `pymupdf` or `pdf2image`)
2. Instantiate an **`ImageRedactorEngine`**
3. Pass the PIL Image from step 1 to the [`ImageRedactorEngine.redact`](https://github.com/microsoft/presidio/blob/main/presidio-image-redactor/presidio_image_redactor/image_redactor_engine.py#L27-L34) method
4. Since the **GLiNERRecognizer** is used, its [`analyze`](https://github.com/microsoft/presidio/blob/main/presidio-analyzer/presidio_analyzer/predefined_recognizers/gliner_recognizer.py#L125-L131) method processes the full OCR-generated text, causing the GLiNER model to truncate the input

**Expected behavior**
The [`analyze`](https://github.com/microsoft/presidio/blob/main/presidio-analyzer/presidio_analyzer/predefined_recognizers/gliner_recognizer.py#L125-L131) method should implement **text chunking** before passing the input to the GLiNER model. Possible solutions include using text splitters from LangChain, such as:
- [spaCy Text Splitter](https://python.langchain.com/docs/how_to/split_by_token/#spacy)
- [Sentence Transformers Splitter](https://python.langchain.com/docs/how_to/split_by_token/#sentencetransformers)

**Additional context**
The [`gliner-spacy`](https://github.com/theirstory/gliner-spacy) wrapper (which integrates GLiNER into spaCy pipelines) already implements a **chunking strategy** before feeding text to the model ([see implementation](https://github.com/theirstory/gliner-spacy/blob/main/gliner_spacy/pipeline.py#L60-L96)). This is considered a best practice when working with transformer-based models and should be adopted in Presidio's GLiNERRecognizer.

jedheaj314:
## Change Description

Change to be able to to call recogniser with chunker and call text chunking (character based) when calling analyse in gliner recogniser.

- Updated gliner recognizer to call predict function after chunking long text 
- New base and character based chunkers
- New chunking util for deduplication and offset calculations
- Unit tests

## Issue reference

Fixes #1569 

## Checklist

Architectural/Technical Decisions:

## 🎯 Problem Statement

**Issue #1569:** GLiNER checkpoint has a maximum sequence length of 384 tokens. Documents exceeding this limit are silently truncated with a warning "UserWarning: Sentence of length 20415 has been truncated to 384".

**Impact:**
- **Security:** PII entities beyond 384 tokens are not detected
- **Compliance:** Incomplete data protection scanning (GDPR, HIPAA violations)
- **Reliability:** Silent failure mode with no user notification

**Root Cause:** GLiNER truncating text to fixed context window of 384 tokens to detect PIIs.

---

### Option 1: Increase Model Context Window
**Description:** Retrain or use larger GLiNER model with higher token limit
**Decision:** ❌ **Rejected** - Not feasible, GLiNER architecture is fixed

---

### Option 2: Token-Based Chunking
**Description:** Use GLiNER's tokenizer to chunk at exact token boundaries

| Pros | Cons |
|------|------|
| Precise token count control | Requires tokenizer loading |
| Guarantees < 384 tokens | Model-specific implementation |
| Optimal model utilization | Complex offset calculation |
| | Tokenizer adds latency |
| | Ties implementation to model architecture |

**Decision:** ❌ **Rejected** - Complexity outweighs benefits

---

### Option 3: Character-Based Chunking ✅
**Description:** Split text by character count with configurable overlap

| Pros | Cons |
|------|------|
| Simple implementation | Approximate token count |
| Model-agnostic | Needs safety margin |
| Fast (no tokenization) | May underutilize context window |
| Easy to configure | Requires deduplication |
| Works with any NER model | |

**Decision:** ✅ **SELECTED** - Best balance of simplicity and effectiveness

**Rationale:**
- GLiNER averages ~1.5 chars/token
- 250 chars = ~166 tokens (safe margin under 384)
- Simplicity enables maintainability
- Works with future models without changes

---

### Option 4: Sentence-Based Chunking
**Description:** Split on sentence boundaries using NLP

| Pros | Cons |
|------|------|
| Natural language boundaries | Requires sentence tokenizer |
| Preserves context | Variable chunk sizes |
| Better semantic coherence | Long sentences problematic |
| | Adds dependency |

**Decision:** ❌ **Rejected** - Unnecessary complexity for current needs

**Future Consideration:** Could implement as alternative strategy later

---

### Decision 1: Chunk Size = 250 characters

**Options Considered:**
- **100 chars:** Too conservative, many chunks, slow
- **250 chars:** ✅ Selected - ~166 tokens, safe margin
- **500 chars:** Too risky, may exceed 384 token limit

**Analysis:**
```
Token estimation: chars / 1.5 = tokens
250 / 1.5 = ~166 tokens (< 384 ✓)
Safety margin: 218 tokens = 57% buffer
```

**Why 250?**
- Proven ratio from gliner-spacy reference implementation
- Sufficient context for entity recognition
- Safe margin for token variation (special chars, unicode)

---

### Decision 2: Overlap = 50 characters (20%)

**Problem:** Entities at chunk boundaries might be split

**Options Considered:**
- **0% overlap:** Fast but misses boundary entities ❌
- **10% (25 chars):** ❌ Too small, may still split entities
- **20% (50 chars):** ✅ Selected - catches boundary entities
- **50% (125 chars):** ❌ Excessive duplication, slow

**Why 20%?**
- Average entity length: 10-30 characters
- 50 char overlap covers typical entity spans
- Balances coverage vs redundant processing
- Standard practice in NLP chunking

---

### Decision 3: Word Boundary Preservation

**Problem:** Mid-word breaks confuse NER models

```python
# Extends chunk to nearest word boundary
while end < len(text) and text[end] not in [" ", "\n"]:
    end += 1
```

**Options Considered:**
| Approach | Pros | Cons | Decision |
|----------|------|------|----------|
| Hard cutoff | Simple | Breaks words | ❌ |
| Word boundary | Preserves context | Variable chunk size | ✅ |
| Sentence boundary | Best context | Too complex | ❌ |

**Trade-off Accepted:**
- Chunks may exceed 250 chars slightly
- Still well under token limit (tested: max ~280 chars)
- Better entity detection outweighs strict size limit

---

### Decision 4: Deduplication Strategy

**Problem:** Overlapping chunks produce duplicate entities

**Approach:** Score-based deduplication with overlap threshold

```python
overlap_ratio = overlap_length / min(entity1_length, entity2_length)
if overlap_ratio > 0.5:  # 50% threshold
    keep_highest_score_entity()
```

**Why 50% threshold?**
- Catches true duplicates: [10:20] and [10:20] → 100% overlap
- Allows partial overlaps: [10:20] and [15:25] → 50% overlap
- Avoids false positives: [10:20] and [30:40] → 0% overlap

**Alternative Considered:**
- Exact match only (start==start, end==end)
  - ❌ Misses duplicates with slight offset differences
  - Models may return [10:20] in one chunk, [11:20] in another

---

### Decision 5: Architecture Pattern = Strategy Pattern

**Why Strategy Pattern?**

```python
BaseTextChunker (interface)
    ↓
LocalTextChunker (current implementation)
    ↓
Future: RemoteChunker, LangChainChunker, SemanticChunker
```

**Benefits:**
- **Open/Closed Principle:** New strategies without modifying existing code
- **Testability:** Easy to mock chunker in tests
- **Flexibility:** Users can plug in custom chunkers
- **Future-proof:** Supports evolution of chunking approaches

**Trade-off:**
- More code upfront (abstract base class)
- ✅ Worth it: Enables extensibility without breaking changes

---

### Decision 6: Utility Functions - Simple Over Generic

**Approach:** Utilities designed for character-based chunking with clear assumptions

```python
predict_with_chunking(
    text=text,
    predict_func=any_prediction_function,  # Model-agnostic
    chunker=any_chunker_implementation      # Character-based expected
)
```

**Key Assumptions:**
- Chunker has `chunk_size` (characters)
- Chunker has `chunk_overlap` (characters)
- Short-circuit check: `len(text) <= chunker.chunk_size`

**Why Simple Design?**
- **YAGNI Principle:** Token-based chunking was rejected, no need to optimize for it
- **Clarity:** Code is immediately understandable
- **Maintainability:** Fewer abstractions = easier debugging
- **Strategy Pattern Already Provides Extensibility:** New chunker types can be added

**Alternative Considered: More Generic Interface**
```python
# Rejected approach
chunker.should_chunk(text)  # Instead of len() check
chunker.get_overlap_size()  # Dynamic overlap
```

**Why Rejected:**
- ❌ Premature abstraction for unused feature
- ❌ More complex interface
- ❌ Harder to understand
- ✅ Can refactor later if token-based is actually needed (unlikely)

**Trade-off Accepted:**
- Future chunker types must work with character-based assumptions OR
- Refactor utils if genuinely needed (straightforward change)

---

### Offset Calculation

**Critical Decision:** How to map chunk entity positions to original text?

**Approach Selected:**
```python
offset += len(chunk) - chunk_overlap
```

**Why `len(chunk)` not `chunk_size`?**
- Word boundary extension creates variable-length chunks
- Using actual length ensures accurate position mapping
- Example: chunk_size=250, actual=273 (word boundary)

**Validation:**
- Tested with word boundaries: ✅ Positions correct
- Tested with CJK text (no spaces): ✅ Works
- Tested with special characters: ✅ Accurate

---

### Error Handling Philosophy

**Decision:** Trust upstream components (GLiNER), minimal defensive coding

**Rationale from Code Review:**
- 23 potential issues identified
- 22 were false alarms (defensive programming paranoia)
- 1 real bug (parameter redundancy - fixed)

**Approach:**
- ✅ Trust GLiNER to return valid predictions
- ✅ Trust Python to handle edge cases (empty strings, etc.)
- ❌ Avoid unnecessary validation code
- ❌ Don't add error handling "just in case"

**Example - Rejected Defensive Code:**
```python
# Considered but rejected
if pred["end"] > len(text):
    logger.warning("Entity beyond text")
    continue
```

**Why rejected?** GLiNER never returns invalid positions. Adding checks adds complexity for zero benefit.

---

### Performance vs Accuracy
**Decision:** Prioritize accuracy over raw speed

| Aspect | Choice | Rationale |
|--------|--------|-----------|
| Overlap | 20% duplication | Catches boundary entities |
| Word boundaries | Variable chunk size | Better entity detection |
| Deduplication | O(n²) algorithm | Simple and correct |

**Performance Result:**
- 1,000 entities: 12,824/sec (fast enough)
- Typical doc: 10-200 entities, < 0.1s overhead
- ✅ No optimization needed yet

---

### Simplicity vs Flexibility
**Decision:** Strategy pattern for future extensibility

**Trade-off:**
- More code upfront (base class + concrete)
- **Benefit:** Easy to add new chunking strategies
- **Verdict:** ✅ Worth it - prevents future breaking changes

---

### Character-based vs Token-based
**Decision:** Character-based for simplicity

**Trade-off:**
- Less precise token control
- **Benefit:** Model-agnostic, no tokenizer overhead
- **Mitigation:** Large safety margin (57% buffer)
- **Verdict:** ✅ Simplicity wins

---

### Testing Strategy

**Test Coverage:**
- 27 unit tests
- Edge cases: CJK text, newlines, empty strings, zero-length entities
- Real-world scenarios: Long documents, overlapping entities

**Key Test Insights:**
1. Word boundaries don't cause infinite loops (verified with CJK)
2. Offset calculation handles variable chunks (verified with entity positions)
3. Deduplication handles zero-length entities (edge case discovered)

## Chunking Approaches Comparison

| Aspect | Current (Overlap) | LangChain Pattern | LangChain Dependency |
|--------|-------------------|-------------------|----------------------|
| Dependencies | None | None | `langchain-text-splitters` (~2MB+) |
| Overlap | ✅ Yes | ❌ No | ✅ Yes |
| Deduplication | ✅ Required | ❌ Not needed | If overlap enabled |
| Separator Logic | Word boundary | `\n\n` → `\n` → ` ` → `` | `\n\n` → `\n` → ` ` → `` |
| Boundary Accuracy | Higher | Lower | Configurable |
| Maintenance Risk | Low | Low | Medium (external) |

## Why Overlap + Deduplication?

**Without overlap:** Entity "Dr. John Smith" split across chunks → missed

**With overlap:** Same entity detected in multiple chunks → needs deduplication

### Immediate Monitoring Needs
1. **Chunk count distribution:** How many chunks per document?
2. **Deduplication rate:** How many duplicates removed?
3. **Latency impact:** Overhead from chunking?

#### Enhancement 1: Parallel Chunk Processing
**When:** If latency becomes issue (>1s per document)
**Approach:** Process chunks concurrently
**Expected gain:** 2-3x speedup for large documents

#### Enhancement 2: Adaptive Chunk Size
**When:** If we see frequent boundary misses
**Approach:** Adjust chunk size based on entity density
**Trade-off:** Added complexity vs marginal gain

#### Enhancement 3: Alternative Chunking Strategies
**When:** User needs semantic chunking
**Approach:** Implement via Strategy pattern
```python
SemanticChunker()  # Preserves paragraphs
SentenceChunker()  # Natural sentence boundaries  
LangChainChunker() # Integration with LangChain
```

#### Enhancement 4: Deduplication Optimization
**When:** Typical document has >1000 entities
**Approach:** Spatial indexing (O(n log n))
**Current:** Not needed - O(n²) is fast enough

---

### DIFF
--- a/presidio-analyzer/pyproject.toml
+++ b/presidio-analyzer/pyproject.toml
@@ -58,7 +58,7 @@ gliner = [
     "transformers",
     "huggingface_hub",
     "gliner (>=0.2.13,<1.0.0) ; python_version >= '3.10'",
-    "onnxruntime (>=1.19) ; python_version >= '3.10'",
+    "onnxruntime (>=1.19) ; python_version >= '3.10'"
 ]
 langextract = [
     "langextract (>=1.0.0)",


presidio-analyzer/presidio_analyzer/chunkers/character_based_text_chunker.py
```diff
@@ -0,0 +1,70 @@
...
+        """Initialize the character-based text chunker.
+
+        Note: Chunks may slightly exceed chunk_size to preserve complete words.
+        When this occurs, the actual overlap may vary from the specified value.
+
+        :param chunk_size: Target maximum characters per chunk (must be > 0)
+        :param chunk_overlap: Target characters to overlap between chunks
+            (must be >= 0 and < chunk_size)
+        """
+        if chunk_size <= 0:
```
RonShakutai, line 37: Should we export the validation logic into validation function ?

presidio-analyzer/presidio_analyzer/chunkers/__init__.py
```diff
@@ -0,0 +1,19 @@
+"""Text chunking strategies for handling long texts."""
```
RonShakutai, line 1: Looking at these classes, I'm thinking about applying the Factory pattern here.

The idea is that users would define the chunker type via a string in their YAML config (e.g., chunker_type: "character"), and the factory would instantiate the appropriate chunker for them.

This would align with how Presidio handles other configs

WDYT?

presidio-analyzer/presidio_analyzer/chunkers/character_based_text_chunker.py
```diff
@@ -0,0 +1,70 @@
...
+        while start < len(text):
+            # Calculate end position
+            end = (
+                start + self.chunk_size
+                if start + self.chunk_size < len(text)
+                else len(text)
+            )
+
+            # Extend to complete word boundary (space or newline)
+            while end < len(text) and text[end] not in [" ", "\n"]:
```
RonShakutai: Should we export it to constant level parameter and think about more word boundary ? 
Should we let the user option to enhance this list via config or something else ?

presidio-analyzer/presidio_analyzer/chunkers/character_based_text_chunker.py
```diff
@@ -0,0 +1,70 @@
+"""Character-based text chunker with word boundary preservation.
```
RonShakutai, line 1: Could you add a bit logs in debug mode ?

presidio-analyzer/presidio_analyzer/chunkers/character_based_text_chunker.py
```diff
@@ -0,0 +1,70 @@
...
+        chunk_size. For texts without spaces (e.g., CJK languages), chunks
+        may extend to end of text.
+
+        :param text: The input text to chunk
+        :return: List of text chunks with overlap
+        """
+        if not text:
+            return []
+
+        chunks = []
```
RonShakutai, line 100: what would happen when we using languages with no spaces , /n ?  should we log a warning ?

presidio-analyzer/presidio_analyzer/chunkers/chunking_utils.py
```diff
@@ -0,0 +1,105 @@
+"""Utility functions for processing text with chunking strategies."""
```
RonShakutai: Did we consider either (1) reusing an existing splitter (LangChain / NLTK / spaCy / HF tokenizers) or (2) at least aligning our implementation with their separator hierarchy pattern (paragraph → line → word → char)?

presidio-analyzer/presidio_analyzer/chunkers/chunking_utils.py
```diff
@@ -0,0 +1,105 @@
...
+    for chunk in chunks:
+        chunk_predictions = process_func(chunk)
+
+        # Adjust offsets to match original text position
+        for pred in chunk_predictions:
+            pred["start"] += offset
+            pred["end"] += offset
+
+        all_predictions.extend(chunk_predictions)
+        offset += len(chunk) - chunk_overlap
```
RonShakutai: The offset calculation assumes fixed overlap, but CharacterBasedTextChunker extends to word boundaries. Could this cause entity position errors?

presidio-analyzer/presidio_analyzer/chunkers/chunking_utils.py
```diff
@@ -0,0 +1,105 @@
...
+    chunks = chunker.chunk(text)
+    all_predictions = []
+    offset = 0
+
+    for chunk in chunks:
+        chunk_predictions = process_func(chunk)
+
+        # Adjust offsets to match original text position
+        for pred in chunk_predictions:
+            pred["start"] += offset
```
RonShakutai: Should we validate that predictions have required keys? or catch exception if one chunk fail?  and log warning ?

presidio-analyzer/presidio_analyzer/chunkers/chunking_utils.py
```diff
@@ -0,0 +1,105 @@
...
+    :return: Deduplicated list of predictions sorted by position
+    """
+    if not predictions:
+        return predictions
+
+    # Sort by score descending to keep highest scoring entities
+    sorted_preds = sorted(predictions, key=lambda p: p["score"], reverse=True)
+    unique = []
+
+    for pred in sorted_preds:
```
RonShakutai: For n predictions, this is O(n²), could we optimize it using any library or more sophisticated approach ?

presidio-analyzer/presidio_analyzer/chunkers/chunking_utils.py
```diff
@@ -0,0 +1,105 @@
...
+    chunker: BaseTextChunker,
+    process_func: Callable[[str], List[Dict[str, Any]]],
+    chunk_overlap: int,
+) -> List[Dict[str, Any]]:
+    """Process text in chunks and adjust entity offsets.
+
+    :param text: Input text to process
+    :param chunker: Text chunking strategy
+    :param process_func: Function that takes chunk text and returns predictions
+    :param chunk_overlap: Number of characters overlapping between chunks
```
RonShakutai: Why pass chunk_overlapas a separate parameter when it's already available on the chunker? Could this lead to inconsistencies?

ShakutaiGit (commented):
This PR would be a Great addition to presidio capabilities !! and probably used in other use cases.
Left a few comments

presidio-analyzer/presidio_analyzer/chunkers/character_based_text_chunker.py
jedheaj314: This is a good point, but following YAGNI now I think No, I don't think we need to. The validation is only 2 lines and very straightforward. Extracting it would add complexity without real benefit since there's only one chunker type currently and no other code needs this validation. We can refactor later if we add more chunker implementations that share validation logic.

presidio-analyzer/presidio_analyzer/chunkers/character_based_text_chunker.py
jedheaj314: Good catch, but No warning needed. The docstring already documents this: "For texts without spaces (e.g., CJK languages), chunks may extend to end of text."  and warning will just add unnecessary noise.
and we also have a unit test suggesting the behaviour to devs using this.

Most real-world CJK text has punctuation or newlines for boundaries. For pure spaceless text, not splitting mid-character is the right choice to avoid corrupting Unicode. If CJK truncation becomes a real issue, we can add character-based fallback chunking or other ways of chunking approach as a future enhancements.
wdyt?

presidio-analyzer/presidio_analyzer/chunkers/chunking_utils.py
jedheaj314: Good point! We considered this but chose a simple custom implementation for several reasons
Please check the PR description for the justification/info around approaches considered

presidio-analyzer/presidio_analyzer/chunkers/chunking_utils.py
jedheaj314: Verified this works correctly. The offset tracks where the next chunk starts, not where it ends. When chunks extend to word boundaries, the chunker's [start = end - chunk_overlap]
This is verified by test_long_text_with_offset_adjustment in test_chunking_utils.py which passes.

presidio-analyzer/presidio_analyzer/chunkers/chunking_utils.py
jedheaj314: I'd say not adding validation or error handling here would be the right way

Type hints define the contract - callers must provide the correct format
Fail fast is better - if predictions are malformed, a KeyError immediately shows where the bug is
Consistent with existing code
Performance - this runs for every prediction, so validation adds unnecessary overhead
If a chunk fails, it's a recognizer bug that needs fixing, not something to silently skip.

presidio-analyzer/presidio_analyzer/chunkers/chunking_utils.py
jedheaj314: This is a very good observation! have discussed this with Sharon H
I have mentioned this and justification in the commit message as well

TLDR is
I'd suggest keeping the current simple implementation for now since:
It's readable and maintainable
Performance is acceptable for typical entity counts
Adding a dependency just for this would increase complexity
wdyt?

presidio-analyzer/presidio_analyzer/chunkers/__init__.py
jedheaj314: Good idea for future extensibility! However, I'd suggest keeping the current approach for now:

Only one chunker exists so a factory would be premature abstraction
Current API is flexible - Users can already pass any chunker via the parameter in GLiNERRecognizer
YAGNI principle - We can add the factory pattern when we actually have multiple chunker types that need runtime selection 
I have ensured that the current design allows for furture implementation of new chunkers (e.g., sentence-based, semantic-based), we can introduce a factory then. The current design doesn't prevent that future addition.
Does that make sense?

presidio-analyzer/presidio_analyzer/chunkers/character_based_text_chunker.py
jedheaj314: This is good suggestion, I can certainly consider doing this in another PR, would that be okay?
As the current simple implementation solves the immediate problem (GLiNER truncation). We can enhance boundary detection as a separate feature if there's actual demand.

presidio-analyzer/presidio_analyzer/chunkers/chunking_utils.py
jedheaj314: You are right! Removed.

presidio-analyzer/presidio_analyzer/chunkers/character_based_text_chunker.py
jedheaj314: Done!

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/gliner_recognizer.py
```diff
@@ -35,6 +40,9 @@ def __init__(
         multi_label: bool = False,
         threshold: float = 0.30,
         map_location: str = "cpu",
+        chunk_size: int = 250,
+        chunk_overlap: int = 50,
+        text_chunker: Optional[BaseTextChunker] = None,
```
omri374, line 42: Should the chunk_size and chunk_overlap be parameters for the BaseTextChunker? Otherwise the user might get confused - they would pass a text_chunker and a chunk_size, but the chunk_size would not be used (as the text_chunker would have a different chunk_size). To me it feels that passing a text chunker is sufficient, and if someone wants to change chunking parameters, they can configure their text chunker differently. WDYT?

presidio-analyzer/presidio_analyzer/chunkers/chunking_utils.py
```diff
@@ -0,0 +1,102 @@
...
+        return predict_func(text)
+
+    predictions = process_text_in_chunks(
+        text=text,
+        chunker=chunker,
+        process_func=predict_func,
+    )
+    return deduplicate_overlapping_entities(predictions)
+
+def process_text_in_chunks(
```
omri374: I know we are only looking into GliNER now, but how do you see this generalizing to arbitrary functions in Presidio? say a recognizer or a transformer model? How do you see recognizer results (`List[RecognizerResult]`) merged together to align with the original text?

presidio-analyzer/presidio_analyzer/chunkers/chunking_utils.py
omri374: If one wants to use this for other types of predictions, knowing the data class's schema is mandatory. Right now it assumes that the output has "start" and "end" values, but the contract is implicit. Consider defining it in a type or reusing a type from Presidio (like `RecognizerResult`). This would also be useful when we generalize this capability to other parts.

presidio-analyzer/presidio_analyzer/chunkers/chunking_utils.py
omri374: How does this handle cases where we might have broken the entity in the middle? "My name is Alex and I live in New", "York, United States". Is this considered part of the deduplication or a different component?

presidio-analyzer/presidio_analyzer/chunkers/chunking_utils.py
```diff
@@ -0,0 +1,102 @@
+"""Utility functions for processing text with chunking strategies."""
+from typing import Any, Callable, Dict, List
+
+from presidio_analyzer.chunkers.base_chunker import BaseTextChunker
+
+
+def predict_with_chunking(
```
omri374: Could these methods be part of the base chunker? What are the advantages of having it as a utility? Are there cases where one would use those outside the scope of chunking?

RonShakutai:
Hi :) @jedheaj314 
Are you still planning to complete this PR?

jedheaj314:
> Hi :) @jedheaj314 Are you still planning to complete this PR?

Hi, yes, I have worked on review comments will push today or tomorrow

jedheaj314:
> Hi :) @jedheaj314 Are you still planning to complete this PR?

Yes, working through the comments now

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/gliner_recognizer.py
jedheaj314: Agreed! Implemented

presidio-analyzer/presidio_analyzer/chunkers/chunking_utils.py
jedheaj314: I've updated the utilities to use RecognizerResult

presidio-analyzer/presidio_analyzer/chunkers/chunking_utils.py
jedheaj314: This is handled by prevention, not deduplication. CharacterBasedTextChunker extends chunks to word boundaries, so "New York" stays together in both overlapping chunks rather than being split into "New" | "York". Deduplication just removes the duplicate from the overlap region.

Example: With 50-char overlap, "New York" appears complete in both chunks → deduplication keeps the higher-scoring one.

presidio-analyzer/presidio_analyzer/chunkers/chunking_utils.py
jedheaj314: Yes, it generalizes to any recognizer.

presidio-analyzer/presidio_analyzer/chunkers/chunking_utils.py
jedheaj314: Done!

presidio-analyzer/presidio_analyzer/chunkers/__init__.py
jedheaj314: Implemented

presidio-analyzer/pyproject.toml
```diff
@@ -29,6 +29,7 @@ dependencies = [
     "pyyaml",
     "phonenumbers (>=8.12,<10.0.0)",
     "pydantic (>=2.0.0,<3.0.0)",
+    "langchain-text-splitters (>=0.2.0)",
```
RonShakutai: I'm not sure if we want to make this mandatory. What do you think about adding it only to the gliner section?

Also, should we update the requirements for the E2E tests to include its installation?
https://github.com/microsoft/presidio/blob/main/e2e-tests/requirements.txt#L3
```python
-e ../presidio-analyzer[langextract,gliner]
```

presidio-analyzer/presidio_analyzer/chunkers/base_chunker.py
```diff
@@ -0,0 +1,140 @@
...
+            is_duplicate = False
+            for kept in unique:
+                # Check if same entity type and overlapping positions
+                if pred.entity_type == kept.entity_type:
+                    overlap_start = max(pred.start, kept.start)
+                    overlap_end = min(pred.end, kept.end)
+
+                    if overlap_start < overlap_end:
+                        # Calculate overlap ratio
+                        overlap_len = overlap_end - overlap_start
```
RonShakutai, line 130: Is it safe to change the returned results in-place, 
or should we avoid side effects and return new adjusted results?

there is no issue in this PR, but maybe when other will use this chuncker ?

presidio-analyzer/presidio_analyzer/chunkers/base_chunker.py
```diff
@@ -0,0 +1,140 @@
...
+        :return: Deduplicated list of RecognizerResult sorted by position
+        """
+        if not predictions:
+            return predictions
+
+        # Sort by score descending to keep highest scoring entities
+        sorted_preds = sorted(predictions, key=lambda p: p.score, reverse=True)
+        unique = []
+
+        for pred in sorted_preds:
```
RonShakutai, line 120: Note: This ror could lead to O(n^2) complexity, but we can address it later if it becomes a real issue.

presidio-analyzer/presidio_analyzer/chunkers/text_chunker_provider.py
```diff
@@ -0,0 +1,54 @@
...
+        chunker_type = config.pop("chunker_type", "langchain")
+
+        if chunker_type not in _CHUNKER_REGISTRY:
+            raise ValueError(
+                f"Unknown chunker_type '{chunker_type}'. "
+                f"Available: {list(_CHUNKER_REGISTRY.keys())}"
+            )
+
+        chunker_class = _CHUNKER_REGISTRY[chunker_type]
+        return chunker_class(**config)
```
RonShakutai: Do we want a clearer error when config has a typo that correlate to the chosen chuncker, instead of a raw TypeError?
```python
chunker_class = _CHUNKER_REGISTRY[chunker_type]
try:
    return chunker_class(**config)
except TypeError as e:
    raise ValueError(
        f"Invalid configuration for chunker_type '{chunker_type}': {config}"
    ) from e
    ```

Comment (jedheaj314) [reply presidio-analyzer/presidio_analyzer/chunkers/text_chunker_provider.py]: Done

File: presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/gliner_recognizer.py [hunk 18]
```diff
@@ -7,6 +7,7 @@
     LocalRecognizer,
     RecognizerResult,
 )
+from presidio_analyzer.chunkers import BaseTextChunker, LangChainTextChunker
```
RonShakutai: This import should be inside the constractor this way the improt will be conditional

e2e-tests/requirements.txt
```diff
@@ -1,4 +1,4 @@
 requests>=2.32.4
 pytest
--e ../presidio-analyzer[langextract]
+-e ../presidio-analyzer[langextract,gliner]
```
omri374: While having an e2e test for gliner is a great idea, this specific change reduces our ability to test the vanilla version of Presidio. For example, if one of the changes we made works with torch but doesn't work without it, we would not be able to catch it. My suggestion is to either perform the integration tests within the presidio-analyzer module, or perform two e2e tests - one without extra dependencies and one with (langextract, gliner, transformers)

@SharonHart @RonShakutai

omri374: BTW, this is also true for langextract, but there the list of dependencies is simpler than gliner's

presidio-analyzer/pyproject.toml
omri374: @RonShakutai see my other comment on this. 
On `langchain-text-splitters`, I would strongly suggest to find an alternative which doesn't require this entire stack.

omri374: Possible available alternative is spaCy's sentence splitter: https://spacy.io/usage/linguistic-features#sbd

omri374:
Thanks @jedheaj314! This is shaping up quite nicely. I added a few comments on potential side effects of this on Presidio in general to think about, but overall the logic and architecture are solid.

presidio-analyzer/pyproject.toml
RonShakutai: Hi @omri374 Thanks, valid concern.

We’re using langchain-text-splitters only for RecursiveCharacterTextSplitter; it pulls in langchain-core, but none of the broader LangChain/LLM stack is used.

spaCy’s sentence splitter is semantically “smarter”, and the rule-based variant is close, but neither provides built-in size-based chunking with character overlap. Matching the current behavior would still require custom packing/overlap logic.

At the moment, the splitter is only used by GLiNER, which is why the dependency is scoped under the GLiNER extras in the .toml. Longer term, the intent is for Presidio to support multiple chunker implementations and allow the community to contribute additional chunkers.

At that point, it may make sense to either introduce a dedicated “chunkers” dependency group in the toml or promote commonly used chunkers to the vanilla dependencies.

I’m still weighing the trade-off between re-implementing a minimal recursive splitter internally to avoid external dependencies versus relying on a small, well-scoped external library.

on my opinion we should avoid implementing a full chuncker by our own... 
WDYT ? @omri374  @SharonHart

e2e-tests/requirements.txt
RonShakutai: @omri374  I think this is a fair point, and to be honest it’s the first time I have to face such considerations.

On the one hand, we want to be able to test that the vanilla version works. On the other hand, Presidio is innovating and transforming using lang extract, GLiNER, transformers, and more… and we want to have tests as a guard for these critical innovative features.

So if you ask me, the best solution is to have tests for the vanilla version and for the new features.
Maybe to create e2e tests for the extras.

Also, we should consider how many people truly rely on the vanilla version ? The vanilla is also very important as the first experience in Presidio?

omri374: The base package is the bread and butter. It's the first thing that new users would try. We had one case a few years ago where there was a bug which prevented the package from installing and the current state of e2e is the result of this issue. I agree with you that having two e2e scenarios is the right approach.

presidio-analyzer/pyproject.toml
omri374: We should weigh the pros of using this specific chunker with the cons of additional deps, supply chain attack risks, security overhead, package size, oss licenses etc. Do we really need a sophisticated chunker here given that we don't use it for embedding text and capturing context the way langchain uses it?

SharonHart: If we already have a package that has a good enough splitter (a few additional lines of codes in presidio is acceptable) than I vote to not introduce a new dependency. 
We installed langextract in the e2e since we used the framework is an easy way to start a local slm with docker, but definetly if it can be installed in code or depdendency, like with gliner, it should be tested in a lower layer, like spacy or stanza inside the package tests

SharonHart: If we already have a package that has a good enough splitter (a few additional lines of codes in presidio is acceptable) than I vote to not introduce a new dependency. 
We installed langextract in the e2e since we used the framework is an easy way to start a local slm with docker, but definetly if it can be installed in code or depdendency, like with gliner, it should be tested in a lower layer, like spacy or stanza inside the package tests

RonShakutai: I agree with Omri that LangChain may be an over-dependency here. 
Based on broader ecosystem feedback, it’s often considered heavy for simple chunking and increases security and supply-chain risk for relatively small value.

I’m also not in favor of implementing our own chunker for this, it’s not core functionality, 
and we should prefer existing OSS where possible.

One possible alternative is semchunk. 
which supports fixed-length chunking with overlap and returns offsets out of the box:
https://github.com/isaacus-dev/semchunk?tab=readme-ov-file#chunk

We should spike more in my opinon ..

RonShakutai: If both of you vote for our own implementation, we will proceed with this approach. Please have your final vote.
@omri374  @SharonHart

jedheaj314: Thanks @omri374 @SharonHart @RonShakutai for the input!

Good news - we actually implemented an in-house `CharacterBasedTextChunker` earlier in this PR before switching to LangChain.

Here's the comparison from our initial analysis (already had this in PR description):

| Feature | In-House | LangChain |
|---------|----------|-----------|
| Dependencies | **None** | langchain-text-splitters (~2MB+) |
| Overlap | ✅ Yes | ✅ Yes |
| Separator Logic | Word boundary | \n\n → \n → space → char |
| Maintenance Risk | **Low** | Medium (external) |

Given the consensus on avoiding external dependencies and supply chain risks, I'll revert to the in-house approach. It's ~50-80 lines and covers exactly what GLiNER needs.

Will proceed with the update unless there are any concerns.

e2e-tests/requirements.txt
jedheaj314: I've removed langchain-text-splitters and switched to the in-house chunker, so this PR no longer adds dependencies to the vanilla Presidio package.

Happy to create a separate PR for the e2e test restructuring (vanilla vs extras).

presidio-analyzer/presidio_analyzer/chunkers/base_chunker.py
jedheaj314: Done

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/gliner_recognizer.py
jedheaj314: Done

e2e-tests/requirements.txt
omri374: Thanks. Not for this PR, but IMHO we should implement another e2e-testing pipeline for the vanilla version. @RonShakutai @SharonHart

omri374: Or, for now, remove the gliner addition to the e2e tests.

omri374: I see that the e2e flow is failing due to the gliner installation. Let's remove it for now and think about adding more e2e tests for extras in a future PR.

omri374 (approved):

PR #1813, opened 2025-12-12
fix(analyzer): Enable GPU inference for GLiNERRecognizer

eveningcafe:
## What does this PR do?

Enables GPU acceleration for GLiNER-based PII detection by fixing the `map_location` parameter handling.

## Why is this needed?

Users who want to run GLiNER inference on GPU for faster PII detection currently cannot do so. Even when configuring `map_location: "cuda"` in the recognizer settings, the model still loads on CPU because the parameter is never passed to the underlying GLiNER library.

**Before this fix:**
```yaml
# recognizers.yaml - This setting was ignored!
- name: GLiNERRecognizer
  map_location: "cuda"  # ❌ Model still loads on CPU
```

**After this fix:**
```yaml
# recognizers.yaml - Now works as expected
- name: GLiNERRecognizer
  map_location: "cuda"  # ✅ Model loads on GPU
```

## Changes

- Pass `map_location=self.map_location` to `GLiNER.from_pretrained()` in the `load()` method

## Testing

- Verified GPU inference works with `map_location="cuda"` on NVIDIA RTX 2080 Ti
- Backward compatible: default value remains `"cpu"`


omri374:
Thanks! Good catch

omri374:
cc @RonShakutai

RonShakutai:
> cc @RonShakutai

Good catch, this is already addressed in the GPU optimizations PR.
We can merge this for now and it will be overwritten there afterward.

RonShakutai (approved):

PR #1827, opened 2026-01-04
fix: Fixed context enhancement substring matching bug 

### LINKED ISSUE #1061: Context enhancement increasing context even if no context word is present in configured vicinity
**Describe the bug**
After detecting PII from text, while searching for context, context list is being used such that it also looks for
substrings for matching, which causes context enhancement even if there is no precise context available. 
Ex. For Driver license, 'lic' is one of the word in context list. Program is supposed to consider context if 'lic' appear anywhere around the identified token. But program also consider matches for substring. which causes false context enhancement. If there is a word 'duplicate' in context distance considered, it's being counted for enhancing score as 'lic' is present in 'dupLICate' as a substring. 

**To Reproduce**
1. Try detecting entity US_DRIVER_LICENSE for text -> "Add 15 µL Taqman reaction mixture to a 7900HT PCR plate (384-well, duplicate) (A1-4 – H1-4)"
A1 and H1 are getting detected as license numbers, with score 0.65 [0.3 pattern match and 0.35 context enhancement]
**Expected behavior**
Context enhancement score should not be added as 

**Screenshots**
this is happening because of the code block below
image

**Additional context**
Similar false positive occurs in case of passports frequently because 'us' is one of the word in context list and most likely 'us' can be substrings in multiple words. Please enlighten on if this is intentional and have specific scenarios.

ravi-jindal:
## Change Description
Fixed context enhancement substring matching bug where context words were incorrectly matched as substrings (e.g., 'lic' matching 'duplicate'). Now uses whole-word matching for accurate context detection 
Describe your changes

## Issue reference

Fixes #1061 

## Checklist

### DIFF
--- a/docs/samples/python/streamlit/requirements.txt
+++ b/docs/samples/python/streamlit/requirements.txt
@@ -6,7 +6,7 @@ streamlit-tags
 pandas>=1.0.4
 python-dotenv
 st-annotated-text
-torch>=2.7.0
+torch>=2.8.0
 flair>=0.15.0
 openai
 azure-ai-textanalytics


ravi-jindal:
> @microsoft-github-policy-service agree

omri374:
Thanks! There was a reason behind this implementation. Context words might not always be separated by spaces. For example, card and credit_card. We can decide that context words must be exhaustive, but I wonder if you have suggestions on having both options available (either defined by config or working in parallel)

ravi-jindal:
> Thanks! There was a reason behind this implementation. Context words might not always be separated by spaces. For example, card and credit_card. We can decide that context words must be exhaustive, but I wonder if you have suggestions on having both options available (either defined by config or working in parallel)

Please check now.
Added context_matching_mode parameter to LemmaContextAwareEnhancer:
"whole_word" (default): exact whole-word matching (fixes the bug)
"substring": substring matching (for compound words like "creditcard")
"hybrid": tries whole-word first, then substring if no match

presidio-analyzer/presidio_analyzer/context_aware_enhancers/lemma_context_aware_enhancer.py
```diff
@@ -23,6 +23,10 @@ class LemmaContextAwareEnhancer(ContextAwareEnhancer):
     :param min_score_with_context_similarity: Minimum confidence score
     :param context_prefix_count: how many words before the entity to match context
     :param context_suffix_count: how many words after the entity to match context
+    :param context_matching_mode: Matching mode for context words. Options:
+        - "whole_word" (default): Match context words only as whole words (e.g., 'lic' matches 'lic' but not 'duplicate')
+        - "substring": Match context words as substrings (e.g., 'card' matches 'creditcard', 'lic' matches 'duplicate')
+        - "hybrid": Try whole-word matching first, then substring matching if no match found
```
omri374: How is hybrid different from substring? The "whole_word" mode is contained within "substring" so isn't "hybrid" going to give the exact same results as "substring"?

presidio-analyzer/presidio_analyzer/context_aware_enhancers/lemma_context_aware_enhancer.py
```diff
@@ -31,13 +35,20 @@ def __init__(
         min_score_with_context_similarity: float = 0.4,
         context_prefix_count: int = 5,
         context_suffix_count: int = 0,
+        context_matching_mode: str = "whole_word",
```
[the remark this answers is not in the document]
omri374: I would suggest to change the default to "substring" to match the existing behavior. @SharonHart would you agree?

presidio-analyzer/presidio_analyzer/context_aware_enhancers/lemma_context_aware_enhancer.py
omri374: One thing we can consider here is boosting the score higher if it's a whole word match vs. substring

omri374:
hi @ravi-jindal, this is a great improvement that we'd be happy to have merged. Are you interested in finalizing this PR?

presidio-analyzer/presidio_analyzer/context_aware_enhancers/lemma_context_aware_enhancer.py
ravi-jindal: Removed the "hybrid" mode.

omri374:
Hi @ravi-jindal, could you please fix the linting issues?

ravi-jindal:
> Hi @ravi-jindal, could you please fix the linting issues?

done

omri374 (approved):

PR #1834, opened 2026-01-17
feat: Add HuggingFaceNerRecognizer for direct NER model inference

### LINKED ISSUE #1833: Add HuggingFaceNerRecognizer for direct NER model inference
## Summary
I'd like to propose adding a `HuggingFaceNerRecognizer` that uses HuggingFace Transformers pipeline directly for NER, bypassing spaCy tokenizer alignment issues.

## Problem
The current approach using spaCy tokenizer with HuggingFace NER models has alignment issues for agglutinative languages (Korean, Japanese, Turkish, etc.):

- Particles/postpositions attach to nouns: "김태웅**이고**" (Kim Taewoong + particle)
- spaCy tokenizer: `"김태웅이고"` (includes particle)
- NER model result: `"김태웅"` (name only)
- `char_span()` alignment fails with `alignment_mode="strict"`
- `alignment_mode="expand"` includes particles, causing downstream issues

### Example (Korean)

| Step | Result |
|------|--------|
| Input Text | `"내 이름은 김태웅이고 전화번호는 010-1234-5678이야"` |
| spaCy Token | `"김태웅이고"` (name + particle) |
| NER Result | `"김태웅"` (name only, start=6, end=9) |
| char_span(strict) | **SKIP** (boundary mismatch) |
| char_span(expand) | `"김태웅이고"` (includes particle → wrong!) |

## Proposed Solution
Create `HuggingFaceNerRecognizer` that:
1. Uses HuggingFace Transformers pipeline directly (bypasses spaCy tokenizer)
2. Returns NER results without char_span alignment
3. Supports any language with a HuggingFace NER model

### Key Features
- **Language-agnostic**: Works with any HuggingFace NER model (English, Korean, Japanese, etc.)
- **Direct inference**: No spaCy tokenizer dependency for entity boundaries

ultramancode:
Fixes #1833

## Description
This PR adds `HuggingFaceNerRecognizer`, a recognizer that uses HuggingFace Transformers pipeline directly for NER, bypassing spaCy tokenizer alignment issues.

### Why is this needed?
The standard approach using spaCy tokenizer with TransformersNlpEngine has alignment issues for agglutinative languages (Korean, Japanese, Turkish, etc.):
- Particles/postpositions attach to nouns
- spaCy tokenizer includes particles: `"김태웅이고"` (name + particle)
- NER model returns only the entity: `"김태웅"` (name only)
- `char_span()` alignment fails, causing entities to be skipped or incorrectly bounded

### Solution
`HuggingFaceNerRecognizer` bypasses spaCy alignment by using HuggingFace pipeline directly.

### Key Features
- **Language-agnostic**: Works with any HuggingFace NER model
- **Direct inference**: No spaCy tokenizer dependency for entity boundaries
- **Extensible YAML Configuration**: Implemented a dedicated Pydantic model `HuggingFaceRecognizerConfig`
 to strictly validate recognizer-specific parameters (e.g., model_name, device, label_mapping) directly from YAML, ensuring type safety and clear error messages.

## Changes
- **NEW**: `presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py`
- **NEW**: `tests/test_huggingface_ner_recognizer.py`
- **MODIFY**: `presidio_analyzer/input_validation/yaml_recognizer_models.py` (Allow ML fields in config)
- **MODIFY**: `presidio_analyzer/recognizer_registry/recognizers_loader_utils.py` (Pass custom args to recognizers)
  - Updated argument preparation to correctly pass list-based arguments like supported_entities to the new recognizer.
- **MODIFY**: `presidio_analyzer/predefined_recognizers/ner/__init__.py` (export)
- **MODIFY**: `presidio_analyzer/predefined_recognizers/__init__.py` (export)
- **MODIFY**: `presidio_analyzer/conf/default_recognizers.yaml` (config, disabled by default)
- **MODIFY**: `docs/analyzer/recognizer_registry_provider.md` (Added configuration example)
- **MODIFY**: `CHANGELOG.md`

### Verification Example: Side-by-Side Comparison

To prove the necessity of this feature, I configured a test environment where **both** the new `HuggingFaceNerRecognizer` and the default `SpacyRecognizer` are enabled for Korean. This allows for comparing their performance directly on the same text.

**Test Configuration:**
```yaml
recognizers:
  - name: "HuggingFace NER KR"
    class_name: "HuggingFaceNerRecognizer"
    model_name: "Leo97/KoELECTRA-small-v3-modu-ner"
    supported_languages: ["ko"]
  
  - name: "SpacyRecognizer" # Intentionally enabled for comparison
    class_name: "SpacyRecognizer"
    supported_languages: ["ko"]
    enabled: true
```

postman_2

**Request:**
```bash
curl --location 'http://localhost:3000/analyze' \
--header 'Content-Type: application/json' \
--data '{
    "text": "제 이름은 김태웅이고 서울에 살고 있습니다.",
    "language": "ko",
    "return_decision_process": true
}'
```

**Result Analysis:**
The response contains three entities. The first two are correct detections enabled by this PR, while the third is the incorrect "noise" produced by the default system.

*   **Result 1 & 2 (Success)**: The `HuggingFaceNerRecognizer` correctly splits and identifies "김태웅" (Kim Taewoong) as PERSON and "서울" (Seoul) as LOCATION.
*   **Result 3 (Failure)**: The `SpacyRecognizer` fails to handle Korean agglutination, capturing the phrase "이름은 김태웅이고..." (My name is Kim Taewoong and...) as a single PERSON entity.

```json
[
    {
        "analysis_explanation": {
            "original_score": 0.9791115522384644,
            "pattern": null,
            "pattern_name": null,
            "recognizer": "HuggingFace NER KR",
            "regex_flags": null,
            "score": 0.9791115522384644,
            "score_context_improvement": 0,
            "supportive_context_word": "",
            "textual_explanation": "Identified as PERSON by Leo97/KoELECTRA-small-v3-modu-ner (original label: PS)",
            "validation_result": null
        },
        "end": 9,
        "entity_type": "PERSON",
        "score": 0.9791115522384644,
        "start": 6
    },
    {
        "analysis_explanation": {
            "original_score": 0.9564878344535828,
            "pattern": null,
            "pattern_name": null,
            "recognizer": "HuggingFace NER KR",
            "regex_flags": null,
            "score": 0.9564878344535828,
            "score_context_improvement": 0,
            "supportive_context_word": "",
            "textual_explanation": "Identified as LOCATION by Leo97/KoELECTRA-small-v3-modu-ner (original label: LC)",
            "validation_result": null
        },
        "end": 14,
        "entity_type": "LOCATION",
        "score": 0.9564878344535828,
        "start": 12
    },
    {
        "analysis_explanation": {
            "original_score": 0.85,
            "pattern": null,
            "pattern_name": null,
            "recognizer": "SpacyRecognizer",
            "regex_flags": null,
            "score": 0.85,
            "score_context_improvement": 0,
            "supportive_context_word": "",
            "textual_explanation": "Identified as PERSON by Spacy's Named Entity Recognition",
            "validation_result": null
        },
        "end": 23,
        "entity_type": "PERSON",
        "score": 0.85,
        "start": 2
    }
]
```

###  Production Configuration Tip

Although the HuggingFace recognizer functions independently, the Presidio Analyzer Platform requires a default NLP engine declaration for startup. For production environments where Spacy NER is not needed, I recommend explicitly disabling the default SpacyRecognizer to avoid duplication and performance overhead.

```yaml
- name: "SpacyRecognizer"
  type: "predefined"
  supported_languages: ["ko"]
  enabled: false
```

## Testing
Added unit tests covering:
- English person/location/organization detection
- Korean text with particles (agglutinative language demo)
- Multiple entity types
- Low confidence filtering
- Custom label mapping
- Empty text handling
- Model name validation
- Long text chunking & truncation safety

### DIFF
--- a/docs/analyzer/nlp_engines/gpu_usage.md
+++ b/docs/analyzer/nlp_engines/gpu_usage.md
@@ -2,6 +2,28 @@
 
 Presidio supports GPU acceleration for NLP models, which can significantly improve performance when processing large volumes of text. GPU is automatically detected and used when available—no code changes required.
 
+## Environment Variable Override
+
+You can explicitly control device selection using the `PRESIDIO_DEVICE` environment variable. Any valid PyTorch device string is accepted:
+
+```bash
+# Force CPU usage
+export PRESIDIO_DEVICE=cpu
+
+# Force GPU usage
+export PRESIDIO_DEVICE=cuda
+
+# Use a specific GPU
+export PRESIDIO_DEVICE=cuda:0
+export PRESIDIO_DEVICE=cuda:1
+
+# Or when running Docker containers
+docker run -e PRESIDIO_DEVICE=cuda mcr.microsoft.com/presidio-analyzer:latest
+```
+
+!!! note "Priority"
+    When `PRESIDIO_DEVICE` is set, it takes priority over automatic detection. If not set (or empty), Presidio will auto-detect available hardware.
+
 ## Prerequisites
 
 ### Hardware Requirements
--- a/presidio-analyzer/presidio_analyzer/nlp_engine/device_detector.py
+++ b/presidio-analyzer/presidio_analyzer/nlp_engine/device_detector.py
@@ -3,15 +3,23 @@
 This module provides lazy device detection using module-level __getattr__.
 Access via `device_detector` attribute which is lazily initialized on first use.
 
+The device can be explicitly set via the PRESIDIO_DEVICE environment variable.
+Any valid PyTorch device string is accepted (e.g., 'cpu', 'cuda', 'cuda:0', 'cuda:1').
+
+If PRESIDIO_DEVICE is not set, automatic detection is performed.
+
 Usage:
     from presidio_analyzer.nlp_engine.device_detector import device_detector
     device = device_detector.get_device()
 """
 
 import logging
+import os
 import threading
 
 logger = logging.getLogger("presidio-analyzer")
+
+PRESIDIO_DEVICE_ENV_VAR = "PRESIDIO_DEVICE"
 
 _detector_instance = None
 _lock = threading.Lock()
@@ -20,24 +28,41 @@ _lock = threading.Lock()
 class DeviceDetector:
     """Detect and expose PyTorch device availability.
 
-    Prefer CUDA when available (Linux/Windows/NVIDIA),
-    otherwise fall back to CPU.
+    Device selection priority:
+    1. PRESIDIO_DEVICE environment variable (any valid device string)
+    2. Automatic detection: Prefer CUDA when available, otherwise CPU
 
     Note: MPS (Apple Silicon/Metal) is currently not supported.
     """
 
     def __init__(self) -> None:
-        self._device = "cpu"
-        self._detect()
+        self._device = self._get_device_from_env() or self._detect()
         logger.info(f"Using device of type: {self._device}")
 
-    def _detect(self) -> None:
-        """Detect PyTorch CUDA support once."""
+    def _get_device_from_env(self) -> str | None:
+        """Get device from PRESIDIO_DEVICE environment variable.
+
+        Returns
+            Device string if env var is set, None otherwise.
+        """
+        env_device = os.environ.get(PRESIDIO_DEVICE_ENV_VAR, "").strip()
+        if not env_device:
+            return None
+
+        logger.info(f"Device set to '{env_device}' via {PRESIDIO_DEVICE_ENV_VAR}")
+        return env_device
+
+    def _detect(self) -> str:
+        """Auto-detect PyTorch CUDA support.
+
+        Returns
+            'cuda' if available, 'cpu' otherwise.
+        """
         try:
             import torch
         except ImportError:
             # torch not installed - this is expected, silently fall back to CPU
-            return
+            return "cpu"
 
         try:
             if torch.cuda.is_available():
@@ -45,9 +70,11 @@ class DeviceDetector:
                 _ = torch.cuda.get_device_name(0)
                 torch.cuda.get_device_capability(0)
                 torch.cuda.empty_cache()
-                self._device = "cuda"
+                return "cuda"
         except Exception as e:
             logger.warning(f"CUDA device detection failed, falling back to CPU: {e}")
+
+        return "cpu"
 
     def get_device(self) -> str:
         """Return device string ('cuda' or 'cpu')."""
--- a/presidio-analyzer/tests/test_device_detector.py
+++ b/presidio-analyzer/tests/test_device_detector.py
@@ -1,10 +1,74 @@
 """Unit tests for DeviceDetector."""
 
+import os
 from unittest.mock import MagicMock, patch
 
 import pytest
 
-from presidio_analyzer.nlp_engine.device_detector import DeviceDetector, device_detector
+from presidio_analyzer.nlp_engine.device_detector import (
+    DeviceDetector,
+    PRESIDIO_DEVICE_ENV_VAR,
+    device_detector,
+)
+
+
+class TestDeviceDetectorEnvironmentVariable:
+    """Test suite for PRESIDIO_DEVICE environment variable handling."""
+
+    def test_when_env_var_set_to_cpu_then_uses_cpu(self):
+        """Test that CPU is used when PRESIDIO_DEVICE=cpu."""
+        with patch.dict(os.environ, {PRESIDIO_DEVICE_ENV_VAR: "cpu"}):
+            detector = DeviceDetector()
+            assert detector.get_device() == "cpu"
+
+    def test_when_env_var_set_to_cuda_then_uses_cuda(self):
+        """Test that CUDA is used when PRESIDIO_DEVICE=cuda."""
+        with patch.dict(os.environ, {PRESIDIO_DEVICE_ENV_VAR: "cuda"}):
+            detector = DeviceDetector()
+            assert detector.get_device() == "cuda"
+
+    def test_when_env_var_set_to_specific_gpu_then_uses_it(self):
+        """Test that specific GPU device strings are passed through."""
+        with patch.dict(os.environ, {PRESIDIO_DEVICE_ENV_VAR: "cuda:0"}):
+            detector = DeviceDetector()
+            assert detector.get_device() == "cuda:0"
+
+        with patch.dict(os.environ, {PRESIDIO_DEVICE_ENV_VAR: "cuda:1"}):
+            detector = DeviceDetector()
+            assert detector.get_device() == "cuda:1"
+
+    def test_when_env_var_has_whitespace_then_trimmed(self):
+        """Test that whitespace in env var value is trimmed."""
+        with patch.dict(os.environ, {PRESIDIO_DEVICE_ENV_VAR: "  cpu  "}):
+            detector = DeviceDetector()
+            assert detector.get_device() == "cpu"
+
+        with patch.dict(os.environ, {PRESIDIO_DEVICE_ENV_VAR: "  cuda:0  "}):
+            detector = DeviceDetector()
+            assert detector.get_device() == "cuda:0"
+
+    def test_when_env_var_empty_then_auto_detects(self):
+        """Test that empty env var triggers auto-detection."""
+        with patch.dict(os.environ, {PRESIDIO_DEVICE_ENV_VAR: ""}):
+            detector = DeviceDetector()
+            assert detector.get_device() in ["cpu", "cuda"]
+
+    def test_when_env_var_not_set_then_auto_detects(self):
+        """Test that missing env var triggers auto-detection."""
+        env_without_presidio_device = {
+            k: v for k, v in os.environ.items() if k != PRESIDIO_DEVICE_ENV_VAR
+        }
+        with patch.dict(os.environ, env_without_presidio_device, clear=True):
+            detector = DeviceDetector()
+            assert detector.get_device() in ["cpu", "cuda"]
+
+    def test_when_env_var_set_skips_auto_detection(self):
+        """Test that setting env var skips auto-detection entirely."""
+        with patch.dict(os.environ, {PRESIDIO_DEVICE_ENV_VAR: "cpu"}):
+            with patch.object(DeviceDetector, "_detect") as mock_detect:
+                detector = DeviceDetector()
+                mock_detect.assert_not_called()
+                assert detector.get_device() == "cpu"
 
 
 class TestDeviceDetectorErrorPaths:


ultramancode:

docs/analyzer/recognizer_registry_provider.md
```diff
@@ -100,3 +110,16 @@ The recognizer list comprises of both the predefined and custom recognizers, for
...
+
+    When using `HuggingFaceNerRecognizer` for languages where the default spaCy tokenizer performs poorly, keep the default NLP engine enabled (for startup) but disable its output by setting `supported_entities` to an empty list.
+
+    ```yaml
+    - name: "SpacyRecognizer"
+      type: "predefined"
+      class_name: "SpacyRecognizer"
+      supported_languages: ["ko"]
+      enabled: true # Ensure at least one NLP engine is active for this language
+      supported_entities: []  # Disable output while keeping the engine active
```
omri374: Why not disable it instead?

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
```diff
@@ -0,0 +1,430 @@
...
+                "Install with: pip install transformers torch"
+            )
+
+        if not self.model_name:
+            raise ValueError(
+                "model_name must be set before calling load(). "
+                "Pass it to __init__() or set it directly."
+            )
+
+        device = -1
```
omri374: Please leverage Presidio's DeviceDetector

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
```diff
@@ -0,0 +1,430 @@
...
+        """Normalize label by removing B-/I- prefix.
+
+        Some models return BIO-format labels (e.g., "B-PER", "I-PER") even with
+        aggregation_strategy="simple". This method strips the prefix so that
+        label_mapping can correctly map to Presidio entities.
+
+        Example: "B-PER" -> "PER" -> mapped to "PERSON"
+        """
+        if isinstance(label, str):
+            if label.startswith("B-") or label.startswith("I-"):
```
omri374: in other schemas such as BILOU the label could also start with `U` or `L`

ultramancode:
I’ll reuse PR #1805’s chunking logic once it’s merged to keep things consistent. Any other feedback is welcome.

ultramancode:
Hi @omri374 I’ve updated the PR to adopt the BaseTextChunker chunking logic from #1805. When you have time, could you please review it?

omri374:
Thanks! Sure, will do.

ultramancode:
@omri374 
I added a few more tests so the CI checks pass. Thanks!

presidio-analyzer/presidio_analyzer/input_validation/yaml_recognizer_models.py
```diff
@@ -141,6 +141,21 @@ class PredefinedRecognizerConfig(BaseRecognizerConfig):
...
     type: str = Field(default="predefined", description="Type of recognizer")
 
+    # HuggingFaceNerRecognizer and other ML model configuration
+    model_name: Optional[str] = Field(None, description="HuggingFace model name")
+    tokenizer_name: Optional[str] = Field(
+        None, description="HuggingFace tokenizer name"
+    )
+    label_mapping: Optional[Dict[str, str]] = Field(None, description="Label mapping")
+    threshold: Optional[float] = Field(None, description="Confidence threshold")
+    aggregation_strategy: Optional[str] = Field(
```
omri374, line 168: These are specific to the huggingface recognizer, but this change adds them to the generic validation for all recognizers. I would suggest to either:
1. allow users to define the recognizer with kwargs (i.e. they would add parameters to the yaml and those would be propagated into the huggingface recognizer
2. Create a new pydantic class specifically for this recognizer (similar to / inheriting from  PredefinedRecognizerConfig) which would define those new parameters.

Also, we have the `NerModelConfiguration` class which already holds all these parameters. Maybe we can use it as an input? just a thought, no need to implement if it makes this too complex.

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
```diff
@@ -0,0 +1,354 @@
...
+            # Run inference on the chunk
+            preds = self.ner_pipeline(chunk_text)
+
+            # Helper to process a single prediction dictionary
+            def process_pred(pred):
+                raw_label = pred.get("entity_group", pred.get("entity"))
+                model_label = HuggingFaceNerRecognizer._normalize_label(raw_label)
+
+                presidio_entity = self.label_mapping.get(model_label)
+                if not presidio_entity:
```
omri374: I would suggest to return it anyway. It's better to catch it and have the user improve mapping than to ignore it even though it was detected by the model.

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
```diff
@@ -0,0 +1,354 @@
...
+        model_name: Optional[str] = None,
+        label_mapping: Optional[Dict[str, str]] = None,
+        threshold: float = 0.3,
+        aggregation_strategy: str = "simple",
+        chunk_overlap: int = 40,
+        chunk_size: int = 400,
+        max_text_length: Optional[int] = None,
+        device: Optional[Union[str, int]] = None,
+        tokenizer_name: Optional[str] = None,
+        text_chunker: Optional[CharacterBasedTextChunker] = None,
```
omri374: Should be `Optional[BaseTextChunker]`

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
```diff
@@ -0,0 +1,354 @@
...
+        # Use the standard predict_with_chunking method from BaseTextChunker
+        # This handles chunking, processing, and duplicating/merging results
+        results = self.text_chunker.predict_with_chunking(
+            text=text,
+            predict_func=self._predict_chunk,
+        )
+
+        # Filter by requested entities
+        if entities:
+            results = [r for r in results if r.entity_type in entities]
```
omri374: Here too- we should only filter out entities that the model supports but the user is uninterested in. Presidio Analyzer would pass all the entities the model supports into this method, resulting in missed entities if they were not mapped properly. This way we let the user understand that there are unmapped entities, instead of silently removing them which could result in missed PII.

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
```diff
@@ -264,34 +268,48 @@ def _prepare_recognizer_kwargs(
...
-            kwargs.pop("supported_entity", None)
-
-        return kwargs
+        # Merge order: recognizer_conf (base) -> language_conf (overrides)
+        # language_conf contains the specific 'supported_language' and 'context'
+        # for this instance, so it should act as an override for the generic config.
+        raw_kwargs = {**recognizer_conf, **language_conf}
+
+        try:
+            sig = inspect.signature(recognizer_cls)
```
omri374: Could you explain why inspect is needed here? While this is a good workaround for not defining the parameters explicitly, it would fail where a recognizer accepts **kwargs as input and in some other cases where there is some adaptation of parameters to be done prior to initializing the recognizer. See my previous suggestions on how to tackle the recognizer yaml loading.

omri374: If we decide to keep this, please add unit tests.

docs/analyzer/recognizer_registry_provider.md
omri374: If you enable the nlp engine but disable the SpacyRecognizer, it would not output any spacy based entities.

ultramancode: Agreed. I've updated the documentation to use `enabled: false` as suggested.

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
ultramancode: Thanks! I've updated the code to leverage `DeviceDetector` for automatic detection when no specific device is provided.

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
ultramancode: I've added support for BILOU formats (including U- and L-) and also made the prefixes configurable for better flexibility.

presidio-analyzer/presidio_analyzer/input_validation/yaml_recognizer_models.py
ultramancode: I followed suggestion #2 and created a dedicated `HuggingFaceRecognizerConfig` class. I also added a mapping mechanism so this specific config is used only for `HuggingFaceNerRecognizer`, keeping the base configuration clean

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
ultramancode: Agreed. I've updated the logic to return unmapped entities using their original model label instead of ignoring them. This should help with discovery and debugging

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
ultramancode: Fixed. Updated the type hint to `Optional[BaseTextChunker]`

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
ultramancode: Agreed. I've updated the logic to keep unmapped entities for better visibility, instead of silently filtering them out.

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
ultramancode: I removed the inspect logic entirely since I implemented the dedicated `HuggingFaceRecognizerConfig`.

ultramancode:
@omri374 
Thank you for the detailed feedback — it was really helpful!
I’ve addressed all the comments and pushed the updates. Could you please take another look when you have a moment?

presidio-analyzer/presidio_analyzer/input_validation/yaml_recognizer_models.py
```diff
@@ -420,3 +459,10 @@ def validate_language_presence(self):
                     "or specify languages for each custom recognizer."
                 )
         return self
+
+
+# Map specific recognizer classes to their dedicated config models
+# This allows for modular expansion without polluting the base config
+CONFIG_MODEL_MAP: Dict[str, Type[BaseModel]] = {
```
omri374, line 464: Please move to the top of the file

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
```diff
@@ -0,0 +1,446 @@
...
+
+    def _parse_device(self, device: Optional[Union[str, int]]) -> int:
+        """Parse device string or int into a transformer-compatible int.
+
+        Normalize diverse device inputs (None, "cpu", "cuda", "cuda:1", 0)
+        to a standard integer format for the Transformers pipeline.
+        If None, it uses Presidio's DeviceDetector for auto-detection.
+        """
+        if device is None:
+            from presidio_analyzer.nlp_engine import device_detector
```
omri374: Please move imports to the top (if possible)

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
```diff
@@ -0,0 +1,446 @@
...
+            raise ValueError(
+                "model_name must be set before calling load(). "
+                "Pass it to __init__() or set it directly."
+            )
+
+        # Device validation and fallback
+        device = self.device
+        if device >= 0:
+            try:
+                import torch
```
omri374: We can simplify this by putting on top like the transformers import

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
```diff
@@ -0,0 +1,446 @@
...
+            presidio_entity = self.label_mapping.get(model_label)
+            if not presidio_entity:
+                # If label is not mapped, use the model's label as is. This allows
+                # discovering entities not explicitly defined in the mapping.
+                presidio_entity = model_label
+
+            try:
+                score = float(pred.get("score") or 0.0)
+            except (TypeError, ValueError):
+                return
```
omri374, line 344: What kinds of errors are we expecting here? Should we raise/warn if they occur?

presidio-analyzer/presidio_analyzer/input_validation/yaml_recognizer_models.py
omri374: Great!

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
```diff
@@ -0,0 +1,446 @@
...
+        2. Hardware acceleration setup (CUDA validation and fallback)
+        3. Lazy-loading of the heavyweight ML pipeline.
+
+        :raises ImportError: If transformers library is not installed
+        :raises ValueError: If model_name is not set
+        """
+        if self.ner_pipeline is not None:
+            return
+
+        if not hf_pipeline:
```
omri374: Consider moving this assertion to the init. This could let the user know about the missing package earlier than when running it for the first time.

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
```diff
@@ -0,0 +1,446 @@
...
+        chunk_results = []
+        # Run inference on the chunk
+        try:
+            preds = self.ner_pipeline(chunk_text)
+        except Exception as e:
+            logger.warning(f"NER prediction failed for chunk: {e}", exc_info=True)
+            return []
+
+        # Helper to process a single prediction dictionary
+        def process_pred(pred):
```
omri374: please add type hints and a one-line docstring

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
```diff
@@ -0,0 +1,446 @@
...
+                RecognizerResult(
+                    entity_type=presidio_entity,
+                    start=start,
+                    end=end,
+                    score=score,
+                    analysis_explanation=explanation,
+                )
+            )
+
+        if isinstance(preds, list):
```
omri374: is it expected for a huggingface model to return a list of lists or a dict?

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
```diff
@@ -0,0 +1,446 @@
...
+        :param nlp_artifacts: Ignored (spaCy artifacts not used)
+        :return: List of RecognizerResult with detected entities
+        """
+        if not text or not text.strip():
+            return []
+
+        # Defensive guard for entities input
+        entities = entities or []
+
+        if self.max_text_length and len(text) > self.max_text_length:
```
omri374: Is there need to truncate text if we use the chunker right after?

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
```diff
@@ -0,0 +1,446 @@
...
+        # 1. If an entity is requested, it is always kept.
+        # 2. If it's a known 'supported' entity but NOT requested, it is filtered out.
+        # 3. If it's an unmapped/unexpected entity, it is kept for Discovery.
+        if entities:
+            requested = set(entities)
+            supported = set(self.supported_entities)
+            results = [
+                r
+                for r in results
+                if (r.entity_type in requested) or (r.entity_type not in supported)
```
omri374, line 436: nice!

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
```diff
@@ -275,8 +278,13 @@ def _prepare_recognizer_kwargs(
         """
         kwargs = {**recognizer_conf, **language_conf}
 
+        # HuggingFaceNerRecognizer expects `supported_entities` (plural) as-is.
+        # Do not normalize/remove it via PatternRecognizer/non-PatternRecognizer rules.
+        if recognizer_cls.__name__ == "HuggingFaceNerRecognizer":
```
omri374: is this needed? all predefined recognizers expect supported_entities (plural) as-is.

presidio-analyzer/presidio_analyzer/input_validation/yaml_recognizer_models.py
```diff
@@ -363,7 +390,19 @@ def parse_recognizers(
...
+                    # Determine the config model based on the recognizer class_name or name.
+                    recognizer_class_name = recognizer.get("class_name")
+                    recognizer_name = recognizer.get("name")
+                    
+                    # Prioritize class_name for lookup
+                    # (e.g., custom instance of HuggingFaceNerRecognizer)
+                    config_model_key = recognizer_class_name or recognizer_name
+
+                    config_model = CONFIG_MODEL_MAP.get(
+                        config_model_key, PredefinedRecognizerConfig
```
omri374, line 400: please add a unit test for this

presidio-analyzer/presidio_analyzer/input_validation/yaml_recognizer_models.py
ultramancode: Should I move this map? Since `CONFIG_MODEL_MAP` references `HuggingFaceRecognizerConfig`, it needs to be defined after that class to avoid a `NameError` during import.

If you’d prefer to keep it near the top of the file, I can switch to a small helper for lazy initialization instead.

```python
def get_config_model_map() -> Dict[str, Type[BaseModel]]:
    return {"HuggingFaceNerRecognizer": HuggingFaceRecognizerConfig}
```

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
ultramancode: Agreed. I’ve moved all optional imports to the top level.

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
ultramancode: Agreed. I’ve moved all optional imports to the top level.

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
ultramancode: Added logger.warning to handle malformed scores gracefully without failing the entire request.

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
ultramancode: Done. Updated the logic to only expect list[dict] and included a defensive check.

presidio-analyzer/presidio_analyzer/input_validation/yaml_recognizer_models.py
ultramancode: Thanks! Done. Added unit tests

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
ultramancode: I kept this as a defensive guard against extremely large inputs before chunking, to avoid excessive memory or runtime usage. Would you prefer to remove it?

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
ultramancode: Fixed. I initially added this check to avoid touching existing code. I've now refactored the logic to only target `PatternRecognizer` for parameter conversion, which removes the need for this check.

ultramancode:
@omri374 
Thanks again for the thorough review! it really helped clarify my understanding of Presidio’s internals.
I’ve addressed the latest round of comments and pushed the updates. Could you please take another look when you have a moment?

presidio-analyzer/presidio_analyzer/input_validation/yaml_recognizer_models.py
omri374: Can it be placed after the imports?

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
omri374: Yes, the chunker should be able to handle large texts. Otherwise we're risking not identifying PII on long texts.

presidio-analyzer/presidio_analyzer/input_validation/yaml_recognizer_models.py
omri374: Oh I see. In this case we can use `from __future__ import annotations`

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
```diff
@@ -0,0 +1,454 @@
...
+        # BERT-base-NER style labels (with B-/I- prefix stripped)
+        "PERSON": "PERSON",
+        "LOCATION": "LOCATION",
+        "ORGANIZATION": "ORGANIZATION",
+        "DATE_TIME": "DATE_TIME",
+    }
+    DEFAULT_HF_TASK = "token-classification"
+    DEFAULT_CHUNK_SIZE = 400
+    DEFAULT_CHUNK_OVERLAP = 40
+    DEFAULT_THRESHOLD = 0.3
```
omri374: Nit: immutable objects don't require default fields, so int/float values could be placed directly in the init. This makes the docstring (and generated API schema) more readable.

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
ultramancode: Done. I've moved the dependency checks to init for faster failure

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
ultramancode: Done. Added type hints and docstring.

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
ultramancode: Thanks! I have removed the max_text_length parameter and the truncation logic.

presidio-analyzer/presidio_analyzer/input_validation/yaml_recognizer_models.py
ultramancode: I tried moving `CONFIG_MODEL_MAP` right after the imports, but it raises a NameError at import time because `HuggingFaceRecognizerConfig` is referenced as a runtime value (not just a type annotation).
As an alternative, I can declare `CONFIG_MODEL_MAP` (empty) right after the imports with a short comment, and then populate it later via `CONFIG_MODEL_MAP.update({...})` after the config classes are defined. Would that work for you?

omri374: Thanks for the attempt here. No need, let's keep it here after all.

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
```diff
@@ -138,10 +143,10 @@ def get_recognizer_name(recognizer_conf: Union[Dict[str, Any], str]) -> str:
 
     @staticmethod
     def _convert_supported_entities_to_entity(conf: Dict[str, Any]) -> None:
-        if "supported_entities" in conf:
-            supported_entities = conf.pop("supported_entities")
-            if "supported_entity" not in conf and supported_entities:
-                conf["supported_entity"] = supported_entities[0]
+        if RecognizerListLoader.SUPPORTED_ENTITIES in conf:
```
omri374, line 157: Nice!

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
```diff
@@ -264,32 +258,60 @@ def _prepare_recognizer_kwargs(
...
-            if kwargs.get("supported_entity") is None:
-                kwargs.pop("supported_entity", None)
-        else:
-            # For non-PatternRecognizer classes, remove both fields
-            # as they may not accept these parameters
-            kwargs.pop("supported_entities", None)
-            kwargs.pop("supported_entity", None)
+        try:
+            params = inspect.signature(recognizer_cls.__init__).parameters
+        except (TypeError, ValueError):
```
omri374, line 302: Can you explain this change? custom (yaml based) pattern recognizers historically received the singular supported_entity param, while newer versions and predefined received the plural version. Is this new implementation supporting these two use cases?

omri374:
Hi @ultramancode, there are some linting issues and one comment I left on the recognizer loading utils (making sure this didn't change any previously working behavior. Other than that, looks great! Thanks

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
ultramancode: Thanks for the suggestion! I inlined the default values into the `__init__` parameters.

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
ultramancode: Yes, this implementation supports both use cases. The logic uses `inspect.signature()` to dynamically inspect each recognizer class's `__init__` parameters and adapts the kwargs accordingly:

1. Custom (YAML-based) Recognizers (singular)): If the class signature shows only `supported_entity` parameter, `supported_entities[0]` is extracted and passed as `supported_entity`.

2. Predefined/Modern Recognizers (plural): If the class signature includes `supported_entities` parameter, it is passed as-is.

3. Flexible Recognizers (`**kwargs`): Recognizers that accept `**kwargs` receive all fields without modification—whether the implementation uses them is up to the recognizer.

This allows the registry to handle various recognizer types dynamically.

ultramancode:
Hi @omri374, thanks for the feedback. 
I’ve replied to your comments—appreciate you taking the time!

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
ultramancode: Follow-up: I updated the loader logic to handle an edge case with `**kwargs` + inheritance/forwarding.

Issue: A recognizer may accept `**kwargs` but then forward them to a strict parent `__init__`. In that case, an unsupported key like `supported_entity` can still leak into the parent and raise a `TypeError`.

Change:
- `supported_entity` is now dropped unless explicitly accepted by the target recognizer’s `__init__` signature (safety).
- `supported_entities` is kept when `**kwargs` is present (compatibility for existing configs).

I also added `tests/test_recognizers_loader_utils.py` to cover the forwarding scenario and related cases.

If you think there are other edge cases worth covering, I’m happy to add them too.

omri374: Thanks! I will take a look later today

omri374 (approved):
Thanks! Sorry it took so long to finalize the review

ultramancode:
Thanks @omri374! Your feedback helped me a lot. I really appreciate you taking the time to review this so carefully. 
I learned a lot from your review.
Also, it looks like the approval may have been dismissed after I resolved the merge conflict in `CHANGELOG.md`.

omri374:
Thanks @ultramancode! This is a great addition to Presidio, and we'd love for you to continue helping us shape the tool!

Issue #1491, opened 2024-11-28
Presidio Medical Recognizer


RKapadia01:
**Is your feature request related to a problem? Please describe.**
I’m working with Presidio in a context where users may input medical and dietary information into a chatbot.

Currently, Presidio does not have built-in support for detecting medical entities such as diseases, medications, and clinical procedures. This limitation required me to implement a custom recognizer to address the need for medical PII detection.

Below is an example of the custom recognizer I’ve built:

```python
from presidio_analyzer import EntityRecognizer, RecognizerResult
from transformers import AutoTokenizer, AutoModelForTokenClassification, pipeline

class ClinicalBERTRecognizer(EntityRecognizer):
    def __init__(self):
        # Download the model from Hugging Face's model hub
        model_name = "blaze999/Medical-NER"

        # Load the model and tokenizer
        self.tokenizer = AutoTokenizer.from_pretrained(model_name)
        self.model = AutoModelForTokenClassification.from_pretrained(model_name)

        # Create a pipeline for named entity recognition
        self.ner_pipeline = pipeline("ner", model=self.model, tokenizer=self.tokenizer)

        # Define the supported entities
        self.supported_entities = [
            "BIOLOGICAL_ATTRIBUTE",
            "BIOLOGICAL_STRUCTURE",
            "CLINICAL_EVENT",
            "DISEASE_DISORDER",
            "FAMILY_HISTORY",
            "HISTORY",
            "MEDICATION",
            "THERAPEUTIC_PROCEDURE"
        ]

        super().__init__(supported_entities=self.supported_entities)


    def analyze(self, text, entities, nlp_artifacts=None):
        results = []

        # Perform named entity recognition on the input text
        ner_results = self.ner_pipeline(text)

        for entity in ner_results:
            entity_type = entity["entity"].replace("B-", "").replace("I-", "")

            # Check if the entity type is in  the list of supported entities
            if entity_type in self.supported_entities:
                recognizer_result = RecognizerResult(
                    entity_type=entity_type,
                    start=entity["start"],
                    end=entity["end"],
                    score=entity["score"]
                )

                # Create a RecognizerResult object for the entity
                results.append(recognizer_result)

        return results
```

**Describe the solution you'd like**
Would there be interest in incorporating a medical domain recognizer into Presidio? If so, I’d be happy to submit a PR with this implementation or a more generalized version.

[...]

PR #1853, opened 2026-02-12
Add transformer-based MedicalNERRecognizer for clinical entity detection

stevenelliottjr:
## Summary
- Adds a new `MedicalNERRecognizer` that uses a HuggingFace token-classification pipeline to detect clinical entities (diseases, medications, procedures, etc.) — addresses #1491
- Mirrors the existing `GLiNERRecognizer` pattern: extends `LocalRecognizer`, disabled by default (not in `default_recognizers.yaml`), graceful `ImportError` handling when `transformers` is not installed
- Default model is `blaze999/Medical-NER` (configurable), detects 8 entity types with `MEDICAL_` prefix, supports text chunking for long clinical documents

## Changes
- **New**: `presidio_analyzer/predefined_recognizers/ner/medical_ner_recognizer.py` — recognizer implementation
- **New**: `tests/test_medical_ner_recognizer.py` — 10 mock-based tests (no model download required)
- **Modified**: `ner/__init__.py`, `predefined_recognizers/__init__.py` — wired up imports
- **Modified**: `pyproject.toml` — added `medical-ner` optional dependency group

## Test plan
- [x] `pytest tests/test_medical_ner_recognizer.py -v` — 10/10 pass
- [x] `pytest tests/test_gliner_recognizer.py -v` — existing tests unaffected
- [x] Import verification: `from presidio_analyzer.predefined_recognizers import MedicalNERRecognizer` succeeds without `transformers` installed
- [ ] Upstream maintainer review for alignment with Presidio conventions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## [unreleased]
 ### Analyzer
 #### Added
+- US_NPI recognizer for National Provider Identifier with Luhn checksum validation and context support (#1847) (Thanks @stevenelliottjr)
 - UK Postcode (UK_POSTCODE) recognizer with pattern matching and context support
 
 ### Analyzer
--- a/docs/supported_entities.md
+++ b/docs/supported_entities.md
@@ -34,6 +34,7 @@ For more information, refer to the [adding new recognizers documentation](analyz
 |US_DRIVER_LICENSE|A US driver license according to <https://ntsi.com/drivers-license-format/>|Pattern match and context|
 |US_ITIN | US Individual Taxpayer Identification Number (ITIN). Nine digits that start with a "9" and contain a "7" or "8" as the 4 digit.|Pattern match and context|
 |US_MBI|A US Medicare Beneficiary Identifier (MBI) with 11 alphanumeric characters.|Pattern match and context|
+|US_NPI|A US National Provider Identifier (NPI) is a 10-digit number issued to healthcare providers by CMS under HIPAA.|Pattern match, context and checksum|
 |US_PASSPORT |A US passport number with 9 digits.|Pattern match and context|
 |US_SSN|A US Social Security Number (SSN) with 9 digits.|Pattern match and context|
 
--- a/presidio-analyzer/presidio_analyzer/conf/default_recognizers.yaml
+++ b/presidio-analyzer/presidio_analyzer/conf/default_recognizers.yaml
@@ -55,6 +55,12 @@ recognizers:
     type: predefined
     enabled: false
 
+  - name: UsNpiRecognizer
+    supported_languages:
+    - en
+    type: predefined
+    enabled: false
+
   - name: NhsRecognizer
     supported_languages: 
     - en
--- a/presidio-analyzer/presidio_analyzer/predefined_recognizers/__init__.py
+++ b/presidio-analyzer/presidio_analyzer/predefined_recognizers/__init__.py
@@ -69,6 +69,7 @@ from .country_specific.us.us_bank_recognizer import UsBankRecognizer
 from .country_specific.us.us_driver_license_recognizer import UsLicenseRecognizer
 from .country_specific.us.us_itin_recognizer import UsItinRecognizer
 from .country_specific.us.us_mbi_recognizer import UsMbiRecognizer
+from .country_specific.us.us_npi_recognizer import UsNpiRecognizer
 from .country_specific.us.us_passport_recognizer import UsPassportRecognizer
 from .country_specific.us.us_ssn_recognizer import UsSsnRecognizer
 
@@ -137,6 +138,7 @@ __all__ = [
     "UsItinRecognizer",
     "UsLicenseRecognizer",
     "UsMbiRecognizer",
+    "UsNpiRecognizer",
     "UsPassportRecognizer",
     "UsSsnRecognizer",
     "EsNifRecognizer",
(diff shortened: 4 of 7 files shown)


omri374:
Thanks! Are there major differences between this logic and the HuggingfaceNerRecognizer for PII NER? Maybe a simpler integration would be to inherit from the HuggingfaceNerRecognizer and adapt to clinical NER.

stevenelliottjr:
Thanks for the review, @omri374!

I took a close look at the existing `TransformersRecognizer`. The main architectural differences are:

1. **Inference model**: `TransformersRecognizer` inherits from `SpacyRecognizer` and doesn't run inference itself — it consumes pre-processed `NlpArtifacts` from an external NLP engine. `MedicalNERRecognizer` directly loads and runs a HuggingFace `token-classification` pipeline.

2. **Text chunking**: Clinical notes can be very long, so `MedicalNERRecognizer` includes built-in text chunking (via `CharacterBasedTextChunker`) with word-boundary-aware splitting, offset adjustment, and deduplication. `TransformersRecognizer` doesn't handle this.

3. **Entity mapping**: Medical NER models output domain-specific labels (e.g., `DISEASE_DISORDER`, `MEDICATION`) that need configurable mapping to Presidio entity types. The current recognizer has a hardcoded set of 8 generic PII types.

4. **Score threshold filtering**: `MedicalNERRecognizer` filters low-confidence predictions before returning results.

That said, I'm very open to simplifying the integration. A few options:

- **Option A**: Refactor `MedicalNERRecognizer` to inherit from `TransformersRecognizer`/`SpacyRecognizer` and override the methods that differ (though this may fight against the NlpArtifacts-based architecture).
- **Option B**: Extract a shared base class (e.g., `HuggingFaceNerRecognizer`) from the common logic, and have both recognizers inherit from it.
- **Option C**: Keep it as a standalone `LocalRecognizer` subclass but simplify where possible — essentially a purpose-built recognizer for clinical NER, similar to how other domain-specific recognizers exist.

Which approach would you prefer? Happy to refactor accordingly.

omri374:
We have a new addition which is not based on Spacy at all: https://github.com/microsoft/presidio/blob/main/presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py

Apologies if I confused you earlier. @copilot

stevenelliottjr:
Thanks for pointing me to `HuggingFaceNerRecognizer`, @omri374 — does this implementation work?

I've refactored `MedicalNERRecognizer` to be a thin subclass of the upstream `HuggingFaceNerRecognizer`. Here's a summary of changes:

**New file:** `ner/huggingface_ner_recognizer.py`
- Ported the upstream `HuggingFaceNerRecognizer` (from `main`) into this branch so `MedicalNERRecognizer` can subclass it directly.

**Rewritten:** `ner/medical_ner_recognizer.py`
- Now a thin subclass of `HuggingFaceNerRecognizer` (~70 lines) that only sets medical-specific defaults:
  - `model_name = "blaze999/Medical-NER"`
  - `label_mapping = DEFAULT_MEDICAL_ENTITY_MAPPING` (8 clinical entity types)
  - `aggregation_strategy = "simple"`
- All pipeline inference, text chunking, BIO label normalization, device detection, and threshold filtering are inherited from the base class.

**Removed:**
- `conf/medical_ner.yaml` (no longer needed — no spaCy/NLP engine config)
- `spacy_huggingface_pipelines` dependency from `medical-ner` extras
- Old standalone `LocalRecognizer` implementation

**Updated:**
- `pyproject.toml`: `medical-ner` extras now require `transformers`, `torch`, `huggingface_hub` (no spaCy)
- `__init__.py` files: exports updated, `HuggingFaceNerRecognizer` added to public API
- Tests: 16 tests covering inheritance, defaults, pipeline mocking, entity filtering, BIO prefix handling, threshold filtering

No spaCy dependency for this recognizer anymore — it uses `transformers.pipeline("token-classification")` directly via the base class.

presidio-analyzer/pyproject.toml
```diff
@@ -61,6 +61,12 @@ gliner = [
     "onnxruntime (>=1.19, <1.24.1) ; python_version == '3.10'",
     "onnxruntime (>=1.19) ; python_version > '3.10'"
 ]
+medical-ner = [
```
omri374: If a user uses the `transformers` extra, would this be sufficient? if yes, there's no need for another group of extra deps.

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/medical_ner_recognizer.py
```diff
@@ -0,0 +1,68 @@
...
+
+from typing import Dict, List, Optional, Union
+
+from presidio_analyzer.chunkers import BaseTextChunker
+from presidio_analyzer.predefined_recognizers.ner.huggingface_ner_recognizer import (
+    HuggingFaceNerRecognizer,
+)
+
+DEFAULT_MEDICAL_ENTITY_MAPPING: Dict[str, str] = {
+    "DISEASE_DISORDER": "MEDICAL_DISEASE_DISORDER",
```
omri374, line 11: please updated the supported_entities.md file with the new entities here.

omri374:
Thanks! great addition. I left some minor comments

presidio-analyzer/pyproject.toml
stevenelliottjr: Good call — the existing `transformers` extra covers everything needed. I've removed the separate `medical-ner` group. Thanks for pointing that out\!

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/medical_ner_recognizer.py
stevenelliottjr: Done\! I've added a new "Medical / Clinical" section to `supported_entities.md` with all 8 entity types, descriptions, and detection methods.

PR #1855, opened 2026-02-13
Fix _process_names unconditionally treating all metadata as PHI

### LINKED ISSUE #1726: _process_names in class DicomImageRedactorEngine unconditionally adds all DICOM metadata strings as PHI
**Describe the bug**
_process_names in class DicomImageRedactorEngine unconditionally adds all DICOM metadata strings as PHI

**To Reproduce**
Code inspection: first line is ```phi_list = text_metadata.copy()``` and then we just extend to it when is_name is 
True.

**Expected behavior**
_process_names should only add words where is_name is True.

**Additional context**
Too many words are being removed when all we should be removing is PHI.

Mr-Neutr0n:
## Summary

Fixes #1726

`_process_names` in `DicomImageRedactorEngine` has two bugs that cause incorrect PHI detection:

**Bug 1: All metadata treated as PHI regardless of flags**

The method initializes `phi_list` with `text_metadata.copy()`, which means every single DICOM metadata string gets included as PHI — not just the ones where `is_name` (or `is_patient`, since the method is called for both) is `True`. This causes false positives during redaction because non-PHI fields like institution names, study descriptions, and numeric IDs all get flagged.

**Fix:** Initialize `phi_list` as an empty list and only add entries where the corresponding flag is `True`.

**Bug 2: MultiValue DICOM fields stringified incorrectly**

When a metadata value is a `pydicom.multival.MultiValue` (e.g., `MultiValue(str, ['1000', '234566'])`), calling `str()` on it produces `"['1000', '234566']"` — a single string with brackets and quotes baked in. This mangled string then gets passed to `augment_word` and ends up in the PHI list, where it will never match anything useful during redaction.

**Fix:** Check if the value is a `MultiValue`, `list`, or `tuple`, and if so iterate over individual elements, adding and augmenting each one separately.

## Changes

- `presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py`: Rewrote `_process_names` to only include flagged items and properly handle iterable DICOM values
- `presidio-image-redactor/tests/test_dicom_image_redactor_engine.py`: Updated test expectations to reflect that non-flagged metadata should no longer appear in the output
- `presidio-image-redactor/tests/test_phi_list.py`: Adjusted test flags so the MultiValue entry is actually marked as a name (to properly test the MultiValue iteration path)

## Test plan

- [x] Verified `_process_names` with `is_name=[True, False, False]` only returns PHI for the flagged entry
- [x] Verified `MultiValue(str, ['1000', '234566'])` produces individual strings `'1000'` and `'234566'` instead of `"['1000', '234566']"`
- [x] Verified empty input returns empty list
- [x] Verified all-False flags returns empty list
- [x] Existing `test_make_phi_list_flattens_trims_and_dedupes_without_mutation` passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project will be documented in this file.
 
 ## [2.2.361] - 2026-02-12
 ### Analyzer
+#### Changed
+- Fixed context enhancement substring matching bug where context words were incorrectly matched as substrings (e.g., 'lic' matching 'duplicate'). Added configurable `context_matching_mode` parameter to `LemmaContextAwareEnhancer` with two options: "substring" (default, maintains backward compatibility for compound words like "creditcard"), and "whole_word" (prevents false positives like 'lic' matching 'duplicate') (#1061)
+
 #### Added
 - US_MBI recognizer for Medicare Beneficiary Identifier with pattern matching and context support (#1821) (Thanks @chrisvoncsefalvay)
 - MAC address recognizer for detecting MAC addresses in various formats (#1829) (Thanks @kyoungbinkim)
@@ -51,7 +54,6 @@ All notable changes to this project will be documented in this file.
 - Migrated CI workflows (lint, dependency review, release) to ubuntu-slim runners for improved efficiency (#1840) (Thanks @Copilot)
 - Updated actions/cache from v4 to v5 with Node.js 24 runtime support (#1817) (Thanks @dependabot)
 
-## [unreleased]
 ### Image Redactor
 #### Changed
 - DICOM: use_metadata will now use both is_patient and is_name to generate the PHI list of words via change to _make_phi_list.
--- a/docs/samples/python/streamlit/requirements.txt
+++ b/docs/samples/python/streamlit/requirements.txt
@@ -6,7 +6,7 @@ streamlit-tags
 pandas>=1.0.4
 python-dotenv
 st-annotated-text
-torch>=2.7.0
+torch>=2.8.0
 flair>=0.15.0
 openai
 azure-ai-textanalytics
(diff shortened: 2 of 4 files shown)


presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
```diff
@@ -646,18 +646,33 @@ def augment_word(word: str, case_sensitive: bool = False) -> list:
...
+                value = text_metadata[i]
+                # Handle MultiValue, list, and tuple by iterating individual
+                # elements rather than stringifying the entire collection
+                if isinstance(value, (MultiValue, list, tuple)):
+                    for item in value:
+                        text = str(item).strip()
+                        if text:
+                            phi_list.append(text)
+                            phi_list += cls.augment_word(text)
+                else:
```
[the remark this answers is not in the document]
omri374: @Mr-Neutr0n please review this comment

presidio-image-redactor/tests/test_phi_list.py
```diff
@@ -13,8 +13,8 @@ def test_make_phi_list_flattens_trims_and_dedupes_without_mutation():
         "",
         None,
     ]
-    # Only first entry is_name, second is_patient (typical split across tags)
-    is_name =    [True,  False, False, False, False]
+    # First entry is_name, second is_patient, third is_name (to test MultiValue)
+    is_name =    [True,  False, True,  False, False]
```
omri374, line 16: Consider adding more test cases for this specific issue

presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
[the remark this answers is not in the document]
Mr-Neutr0n: Good catch — I collapsed the two branches into one by wrapping scalar values in a list, so now there's a single loop body. Also moved the `MultiValue` import to the top of the module since pydicom is already a top-level dep.

presidio-image-redactor/tests/test_phi_list.py
Mr-Neutr0n: Added a handful of focused tests: non-flagged entries excluded, all-false returns empty, MultiValue iteration produces individual strings, empty/whitespace items skipped, and an end-to-end `_make_phi_list` filtering check. Also dropped the unused `import pydicom` while I was in here.

omri374: Thanks!

PR #1866, opened 2026-02-17
Updated copilot instructions

omri374:
## Change Description

Updated Copilot instructions to better support agent implementations and especially PR review, which was a bit off lately.
@SharonHart @TamirKamara @RonShakutai please go over and see if there are any missing pieces or things that don't make sense. My goal is to have much better automated PR reviews that would reduce the review burden. This should also be used for agentic tasks on Github. We could further consider to create two instructions, one for code generation and one for code review.


omri374:

omri374:
@SharonHart would you like to review or should we merge?

PR #1917, opened 2026-03-22
Fix: config path resolution for PyPI installs (langextract recognizer)

RonShakutai:
## Change Description

LangExtract recognizers (e.g. AzureOpenAILangExtractRecognizer) fail with FileNotFoundError when presidio-analyzer is installed from PyPI. The config path resolver assumed a dev repo layout that doesn't exist in installed packages. Added a fallback that resolves config files relative to the package directory, matching the pattern used across the rest of presidio.
## Issue reference

first run failed, after the fix its passed. 
change in toml to use the branch
presidio-analyzer = {git = "https://github.com/microsoft/presidio.git", branch = "ronshakutai/fix-path-resolve", subdirectory = "presidio-analyzer", extras = ["langextract"]}

image

Fixes #XX

## Checklist

### DIFF
--- a/presidio-analyzer/presidio_analyzer/llm_utils/config_loader.py
+++ b/presidio-analyzer/presidio_analyzer/llm_utils/config_loader.py
@@ -32,7 +32,8 @@ def resolve_config_path(config_path: Union[str, Path]) -> Path:
     Handles paths in multiple formats (checked in order):
     1. Absolute paths: returned as-is
     2. Relative paths that exist from CWD: returned as-is
-    3. Relative paths resolved from repository root
+    3. Paths resolved from the package conf/ directory
+    4. Relative paths resolved from repository root
 
     :param config_path: Configuration file path (string or Path object).
     :return: Resolved absolute path.
@@ -44,6 +45,10 @@ def resolve_config_path(config_path: Union[str, Path]) -> Path:
 
     if config_path_obj.exists():
         return config_path_obj
+
+    conf_resolved = get_conf_path(str(config_path))
+    if conf_resolved.exists():
+        return conf_resolved
 
     presidio_analyzer_root = Path(__file__).parent.parent
     repo_root = presidio_analyzer_root.parent.parent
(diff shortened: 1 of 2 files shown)


omri374:
Thanks @RonShakutai, do you know why it works in other cases, like this one?
https://github.com/microsoft/presidio/blob/9a6cfb42ac8dab328fc615da27e474bb0dd603b8/presidio-analyzer/presidio_analyzer/analyzer_engine_provider.py#L161

And why it wasn't caught in the e2e tests which are installing the package using pip?

presidio-analyzer/presidio_analyzer/llm_utils/config_loader.py
```diff
@@ -46,8 +47,26 @@ def resolve_config_path(config_path: Union[str, Path]) -> Path:
...
+    # This handles PyPI-installed packages where the repo root doesn't exist.
+    path_parts = config_path_obj.parts
+    try:
+        conf_index = path_parts.index("conf")
+        conf_relative = Path(*path_parts[conf_index:])
+        package_resolved = presidio_analyzer_root / conf_relative
+        if package_resolved.exists():
+            return package_resolved
+    except ValueError:
+        pass
```
dorlugasigal: should we add a warning/error log here?

RonShakutai: fixed. remove the except section

RonShakutai:
> Thanks @RonShakutai, do you know why it works in other cases, like this one?
> 
> https://github.com/microsoft/presidio/blob/9a6cfb42ac8dab328fc615da27e474bb0dd603b8/presidio-analyzer/presidio_analyzer/analyzer_engine_provider.py#L161
> 
> And why it wasn't caught in the e2e tests which are installing the package using pip?

`_get_full_conf_path` works because it uses `Path(__file__).parent` which resolves to the installed package location. The langextract configs stored full repo-relative paths that only exist in dev/editable installs — e2e tests use `-e` so the bug was masked. The fix:
- Shortened YAML paths to conf-relative (e.g. `langextract_prompts/default_pii_phi_prompt.j2`)
- Added `get_conf_path()` fallback in `resolve_config_path` (same `Path(__file__) / "conf"` pattern) for bundled defaults, while keeping absolute/CWD-relative resolution first for custom user configs

PR #1943, opened 2026-04-10
fix(analyzer): honour language_model_params in BasicLangExtractRecognizer

### LINKED ISSUE #1942: BasicLangExtractRecognizer silently drops provider.language_model_params (timeout, num_ctx, ...)
## Summary

`BasicLangExtractRecognizer` loads `langextract.model.provider.language_model_params` from the yaml into `self._language_model_params` and forwards it to `langextract.extract(..., language_model_params=...)`. But because the same call also passes `config=ModelConfig(...)` (built from `provider.kwargs` only), `langextract/extraction.py` takes the `elif config:` branch at line 240, which **does not read `language_model_params`** — only the `else:` branch at line 255 does.

Net effect: any value under `provider.language_model_params` in the yaml — including `timeout` and `num_ctx` — is **silently dropped**. Ollama falls back to its 120 s default, which is the source of the confusing

```
InferenceRuntimeError: Ollama Model timed out (timeout=120, num_threads=None)
```

users see even after setting `timeout: 1800` (or any other value) in the yaml. The default bundled config at `presidio-analyzer/presidio_analyzer/conf/langextract_config_basic.yaml` is also affected — the `timeout: 240` and `num_ctx: 8192` it ships with have no effect at runtime.

## Reproduction

```yaml
# repro_config.yaml
lm_recognizer:
  supported_entities: [PERSON]
langextract:
  prompt_file: "langextract_prompts/default_pii_phi_prompt.j2"
  examples_file: "langextract_prompts/default_pii_phi_examples.yaml"
  model:
    model_id: "gemma3:27b"
    provider:
      name: "ollama"
      kwargs:
        model_url: "http://localhost:11434"
      language_model_params:
        timeout: 1800
        num_ctx: 8192
  entity_mappings:
    person: PERSON
```

```python
from presidio_analyzer.predefined_recognizers.third_party.basic_langextract_recognizer import BasicLangExtractRecognizer

r = BasicLangExtractRecognizer(config_path="repro_config.yaml")
cfg = r._get_provider_params()["config"]
print(cfg.provider_kwargs)
# {'model_url': 'http://localhost:11434'}
# ^^ timeout and num_ctx are missing — they are stored on r._language_model_params
#    but never reach the provider.
```

Running any slow local model (e.g. `gemma3:27b` or any model whose first inference takes longer than 120 s on the host) then fails with:

```
langextract.core.exceptions.InferenceRuntimeError: Ollama Model timed out (timeout=120, num_threads=None)
```

## Root cause

`BasicLangExtractRecognizer._get_provider_params()` returns `{"config": self.lx_model_config}`. In `langextract/extraction.py`:

```python
elif config:                                       # taken for BasicLangExtractRecognizer
    language_model = factory.create_model(
        config=config,
        examples=...,
        use_schema_constraints=use_schema_constraints,
        fence_output=fence_output,
    )
else:                                              # only here is language_model_params read
    base_lm_kwargs = {...}
    base_lm_kwargs.update(language_model_params or {})
    filtered_kwargs = {k: v for k, v in base_lm_kwargs.items() if v is not None}
    config = factory.ModelConfig(model_id=model_id, provider_kwargs=filtered_kwargs)
    language_model = factory.create_model(config=config, ...)
```

Only `provider.kwargs` survives as `ModelConfig.provider_kwargs` and reaches `OllamaLanguageModel(**provider_kwargs)`, whose constructor accepts `timeout=` and forwards `num_ctx` via `**kwargs` to every `_ollama_query()` call.

## Why `AzureOpenAILangExtractRecognizer` is not affected

Its `_get_provider_params()` returns `{"model_id": ..., "language_model_params": ...}` — no `config`, so `lx.extract()` takes the `else:` branch where `language_model_params` is actually applied. The bug is specific to `BasicLangExtractRecognizer`.

## Why existing unit tests did not catch this

`presidio-analyzer/tests/test_basic_langextract_recognizer.py:568-570` asserts:

```python
assert "language_model_params" in call_kwargs
assert call_kwargs["language_model_params"]["timeout"] == 180
assert call_kwargs["language_model_params"]["num_ctx"] == 8192
```

i.e. it verifies that `language_model_params` is *passed* to `lx.extract()` — but never that it reaches the `ModelConfig` or the provider. The assertions are satisfied while the effective behaviour is broken. This issue adds regression tests that assert values arrive on `ModelConfig.provider_kwargs` as well.

## Proposed fix

Merge `self._language_model_params` into `self.provider_kwargs` before building `lx_factory.ModelConfig`, using `setdefault` so explicit `provider.kwargs:` entries still win. One-line behaviour change in `basic_langextract_recognizer.py`. PR forthcoming.

## Environment

- `presidio-analyzer` 2.2.362
- `langextract` 1.x
- Python 3.12, macOS 15.5 (Darwin 25.5.0)
- Ollama 0.x, model `gemma3:27b` / `gemma4:31b`

lsternlicht:
## Summary

Fixes #1942.

`BasicLangExtractRecognizer` loads `langextract.model.provider.language_model_params` from the yaml into `self._language_model_params` and forwards it to `lx.extract(..., language_model_params=...)`. But `langextract/extraction.py` ignores the `language_model_params` argument when a pre-built `ModelConfig` is passed via `config=` — it takes the `elif config:` branch at line 240, and only the `else:` branch at line 255 reads `language_model_params`. Because `BasicLangExtractRecognizer._get_provider_params()` always returns `{"config": ModelConfig(...)}`, values like `timeout` and `num_ctx` were silently dropped and Ollama fell back to langextract's 120 s default regardless of what the yaml said.

This reliably breaks any Ollama model whose first inference takes longer than 120 s, with the confusing error:

```
InferenceRuntimeError: Ollama Model timed out (timeout=120, num_threads=None)
```

even when the user explicitly set `timeout: 1800` in the yaml. The default bundled config at `presidio-analyzer/presidio_analyzer/conf/langextract_config_basic.yaml` is also affected — its `timeout: 240` and `num_ctx: 8192` have no effect at runtime.

`AzureOpenAILangExtractRecognizer` is **not** affected, because its `_get_provider_params()` returns `{"model_id": ..., "language_model_params": ...}` with no `config=` key, so `lx.extract()` takes the `else:` branch where `language_model_params` is actually applied.

## Fix

Merge `self._language_model_params` into `self.provider_kwargs` before building `lx_factory.ModelConfig` in `basic_langextract_recognizer.py`. `setdefault` preserves precedence for explicit `provider.kwargs:` entries, so configs that already work by placing `timeout` under `kwargs:` (as a workaround) continue to work unchanged.

```python
for key, value in self._language_model_params.items():
    self.provider_kwargs.setdefault(key, value)
```

## Tests

Before the fix, existing tests were green while the behaviour was broken. `test_basic_langextract_recognizer.py:568-570` only asserted that `language_model_params` was *passed* to `lx.extract()`, but never that it reached the `ModelConfig` or the provider.

This PR:
1. Strengthens `test_when_analyze_called_then_params_passed_to_langextract` to also assert that `timeout` and `num_ctx` land on `call_kwargs["config"].provider_kwargs`.
2. Adds `test_language_model_params_reach_provider_kwargs` — direct regression test for the reported behaviour, without going through the full `analyze()` flow.
3. Adds `test_provider_kwargs_take_precedence_over_language_model_params` — asserts backwards compatibility for configs that already placed values under `kwargs:`.

All 83 LangExtract-related tests pass locally:

```
cd presidio-analyzer
pytest tests/test_basic_langextract_recognizer.py \
       tests/test_azure_openai_langextract_recognizer.py \
       tests/test_langextract_helper.py \
       tests/test_lm_recognizer.py
...
83 passed in 0.59s
```

## Test plan

- [x] All existing `test_basic_langextract_recognizer.py` tests pass (21 tests).
- [x] All LangExtract-related tests across the analyzer pass (83 tests).
- [x] New regression test asserts `ModelConfig.provider_kwargs` contains `timeout` and `num_ctx` from yaml's `language_model_params` section.
- [x] New backwards-compat test asserts `provider.kwargs:` values win over `provider.language_model_params:` values of the same name.
- [x] Verified end-to-end against an actual Ollama `gemma4:31b` run: with the fix, the configured `timeout: 3600` is passed to `requests.post(..., timeout=3600)` inside `langextract/providers/ollama.py`, replacing the hardcoded 120 s default.


presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/basic_langextract_recognizer.py
```diff
@@ -62,14 +62,25 @@ def __init__(
 
         self.model_id = model_config.get("model_id")
         self.provider = provider_config.get("name")
-        self.provider_kwargs = provider_config.get("kwargs", {})
+        self.provider_kwargs = dict(provider_config.get("kwargs", {}))
```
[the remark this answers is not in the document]
RonShakutai: Agree with Copilot.

presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/basic_langextract_recognizer.py
```diff
@@ -62,14 +62,25 @@ def __init__(
...
 
         # Not ideal, but update _extract_params now that self.config is fully loaded.
         self._extract_params.update(provider_config.get("extract_params", {}))
         self._language_model_params.update(
             provider_config.get("language_model_params", {})
         )
```
RonShakutai, line 71: `.update(None)` crashes if the YAML key is present but empty.
so  we should add the or {} like copilot suggested above.

presidio-analyzer/tests/test_basic_langextract_recognizer.py
```diff
@@ -574,3 +574,68 @@ def test_when_analyze_called_then_params_passed_to_langextract(self, tmp_path):
             assert call_kwargs["config"].provider == "ollama"
```
RonShakutai, line 574: Should we add a test where `kwargs: null`  in the yaml and check if the yaml doesnt crash ?

RonShakutai (approved):
LGTM, 
Great PR!!
Left minor comments.

omri374:
Hi @lsternlicht would you be interested in merging this PR? There are a few minor changes to make

RonShakutai:
Hey @lsternlicht, do you plan with proceeding or should we pick it up ? 
CC: @omri374

RonShakutai:
Adopting this PR, therefore closing this one.

PR #1974, opened 2026-04-19
Fix/kvnr checksum 1972

MvdB:
## Change Description
                                                                                                                                                 
Corrects the KVNR (`DE_HEALTH_INSURANCE`) Prüfziffer algorithm introduced in #1909. The previous implementation applied weights `[2,9,8,7,6,5,4,3,2,1]`, which is not the algorithm defined in § 290 SGB V Anlage 1 (GKV-Spitzenverband, Stand 02.01.2023). The spec uses alternating factors `[1,2,1,2,1,2,1,2,1,2]` on the 10 effective digits (2-digit letter ordinal + 8 data digits), takes the Quersumme of any product ≥ 10, and compares `sum mod 10` against the check digit at position 10.                                                                  
                                                                                                                                                 
Worked examples from Anlage 1 (both now in the fixtures):                                                                                        
- `A000500015` → A=01, expanded `0100050001`, products `0,2,0,0,0,10,0,0,0,2`, cross-sums `0,2,0,0,0,1,0,0,0,2` = 5 → PZ 5 ✓
- `C000500021` → C=03, expanded `0300050002`, products `0,6,0,0,0,10,0,0,0,4`, cross-sums `0,6,0,0,0,1,0,0,0,4` = 11 → 11 mod 10 = 1 → PZ 1 ✓

The three previous "valid" fixtures (`A123456787`, `M123456789`, `B123456787`) were generated with the wrong algorithm and never satisfied the spec; they are moved to the negative-case set and replaced with the two Anlage 1 samples plus three recomputed fixtures (`A123456780`, `B123456782`, `M123456785`) and two edge-case fixtures (`Z000000005` for the upper-bound letter ordinal, `Z999999997` for worst-case Quersumme density).

Files changed:
- `presidio-analyzer/.../germany/de_health_insurance_recognizer.py` — algorithm + docstring
- `presidio-analyzer/tests/test_de_health_insurance_recognizer.py` — fixtures + 4 new edge-case parameters
- `docs/recipes/german-language-support/README.md` — sample text updated to a valid KVNR
- `CHANGELOG.md` — entry under `[unreleased] → Analyzer → Fixed`

### Verification
- `ruff check .` clean
- `poetry run pytest` in `presidio-analyzer` (Python 3.13, all extras): **1952 passed, 12 skipped, 0 failed**
- `tests/test_de_health_insurance_recognizer.py`: **29/29 pass** (all branches of `validate_result` covered, plus Z-letter upper bound and worst-case Quersumme density)
[...]

PR #1909, opened 2026-03-15
feat(analyzer): add German PII recognizers (DE_*)

MvdB:
## Summary

Adds 9 new pattern-based recognizers for German-language PII, covering the most relevant entity types under DSGVO/BDSG:

| Entity | Description | Legal Basis | Validation |
|---|---|---|---|
| `DE_TAX_ID` | Steueridentifikationsnummer (11 digits) | § 139b AO | ISO 7064 Mod 11,10 checksum |
| `DE_TAX_NUMBER` | Steuernummer / ELSTER (13 digits + slash formats) | § 139a AO | Regex (state code range 01–16) |
| `DE_PASSPORT` | Reisepass (ICAO Doc 9303, 9 chars) | PassG § 4, PassV | Strict + relaxed ICAO charset |
| `DE_ID_CARD` | Personalausweis — nPA (9 ICAO chars) + old T+8 format | PAuswG, PAuswV | Regex |
| `DE_SOCIAL_SECURITY` | Rentenversicherungsnummer (12 chars) | § 147 SGB VI | DRV checksum algorithm |
| `DE_HEALTH_INSURANCE` | Krankenversicherungsnummer / KVNR (10 chars) | § 290 SGB V | GKV-Spitzenverband checksum |
| `DE_KFZ` | KFZ-Kennzeichen / vehicle registration plate | FZV § 8 | Regex (district + letter-id + digits) |
| `DE_HANDELSREGISTER` | Handelsregisternummer (HRA/HRB prefix) | §§ 9, 14 HGB | Regex |
| `DE_PLZ` | Postleitzahl / postal code (5 digits) | — | Regex (range 01001–99998), score=0.05 |

## Design decisions

- All recognizers follow the existing `PatternRecognizer` pattern and are placed in `predefined_recognizers/country_specific/germany/`
- All are **disabled by default** (`enabled: false`) following the project convention for country-specific recognizers
- Three recognizers with checksum algorithms (`DE_TAX_ID`, `DE_SOCIAL_SECURITY`, `DE_HEALTH_INSURANCE`) implement `validate_result()` returning `True`/`False` to promote/suppress matches — not `None`, to ensure invalid numbers are filtered out rather than kept at base score
- `DE_PLZ` has intentionally low base confidence (0.05) due to high false-positive risk; documented in docstring
- `DE_TAX_NUMBER` includes both ELSTER 13-digit and regional slash-format patterns with graduated confidence scores
[...]

PR #1990, opened 2026-04-19
refactor(de-recognizers): complete Prüfziffer validation per primary sources (fixes #1972, supersedes #1974)

### LINKED ISSUE #1972: Bug: DE_HEALTH_INSURANCE (KVNR) checksum rejects valid values (factor sequence mismatch vs §290 SGB V Anlage 1)
**Describe the bug**
The checksum validation for `DE_HEALTH_INSURANCE` (KVNR) appears to use a factor sequence that is inconsistent with the official §290 SGB V Anlage 1 method, causing valid KVNR examples to be rejected.

**To Reproduce**
1. Go to `presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/germany/de_health_insurance_recognizer.py`.
2. Run the current `validate_result` checksum logic for these sample KVNRs:
   - `A000500015`
   - `C000500021`
3. Observe validation result.

**Expected behavior**
Both sample values should validate as true according to the official KVNR Prüfziffer method (letter -> `01..26`, alternating factors `1,2,1,2,...`, cross-sum, `sum % 10` equals check digit).

**Screenshots**
N/A.

**Additional context**
Current implementation uses factors `[2,9,8,7,6,5,4,3,2,1]`.
Official references:
- https://www.gkv-datenaustausch.de/kvnr/kvnr.jsp
- https://www.gkv-datenaustausch.de/media/dokumente/kvnr/Anlage_1_20230102_Pruefziffernberechnung-KVNR.pdf

This was noticed while validating German health insurance letter extraction after PR #1909.

MvdB:
## Change Description

Extends the pattern-focused work of #1909 by completing the Prüfziffer validation layer of the German recognizers. #1909 brought in all 13 DE recognizers with correct regex patterns and context words, but the checksum / structural-validation side was not comprehensively handled: several algorithms were implemented incorrectly and several were missing entirely. Issue #1972 was the visible symptom of this broader gap. During the follow-up audit I verified each recognizer against its primary source and corrected or added the missing Prüfziffer logic; each new test fixture is derived from a primary-source worked example to prevent the spec-vs-implementation drift from silently recurring.

### Corrected checksum algorithms

- **`DE_HEALTH_INSURANCE`** — factors per § 290 SGB V Anlage 1 (`[1,2,1,2,…]`, Quersumme on products ≥ 10, sum mod 10). Verified against Anlage 1 worked examples `A000500015` and `C000500021`.
- **`DE_SOCIAL_SECURITY` (RVNR)** — weights per VKVV § 4 (`[2,1,2,5,7,1,2,1,2,1,2,1]`). Verified against DRV canonical example `15070649C103`.
- **`DE_LANR`** — weights per KBV Arztnummern-Richtlinie (`[4,9,4,9,4,9]`), without the spurious Quersumme step, using the complement-to-10 formula `(10 − sum mod 10) mod 10`. Verified against physician digits `123456 → 6`.

### Added missing structural checksums

- **`DE_VAT_ID`** — ISO 7064 Mod 11,10, identical to `DE_TAX_ID`. Widely used by community validators (`python-stdnum`, VIES-adjacent); BZSt does not publish it in a Merkblatt but it is the in-practice-authoritative algorithm.
- **`DE_PASSPORT`**, **`DE_ID_CARD`** — ICAO Doc 9303 MRZ check digit (weights 7, 3, 1 repeating; letters A=10 … Z=35; sum mod 10).
- **`DE_ID_CARD`** regex last character tightened to `[0-9]` (the ICAO check digit is always numeric).
- **`DE_BSNR`** — no public Prüfziffer exists (confirmed against KBV Arztnummern-Richtlinie and HL7-Wiki), so added a KV-regional-code whitelist per Arztnummern-Richtlinie Anlage 1 as defense-in-depth.

### Minor

- **`DE_TAX_ID`** — enforce the post-2016 BZSt repetition rule (max 3 repetitions in positions 1–10), replacing the lenient "all-10 must not be identical" check.
- Registered `DeLanrRecognizer`, `DeBsnrRecognizer`, `DeVatIdRecognizer` and `DeFuehrerscheinRecognizer` in `default_recognizers.yaml` — these four were imported in code but missing from the registry configuration, so they were unreachable via the default registry.
- Recipe `docs/recipes/german-language-support/README.md` sample data updated to identifiers that pass the corrected validators.

## Issue reference

Fixes #1972. Supersedes #1974.

## Checklist

- [x] My code includes unit tests (each recognizer's fixtures are derived directly from a primary-source worked example cited in the test docstring)
- [x] All unit tests and lint checks pass locally (2239 passed, 12 skipped on the full analyzer suite; ruff clean)
- [x] My PR contains documentation updates / additions if required (CHANGELOG + recipe sample)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,17 @@ All notable changes to this project will be documented in this file.
 
 - Added recognizer for Swedish Organisationsnummer, ID number for all Swedish oragnisations.
 
+#### Fixed
+- Fixed incorrect Prüfziffer algorithm in `DeHealthInsuranceRecognizer` (KVNR); now uses alternating factors [1,2,…,1,2] per § 290 SGB V Anlage 1 (#1972).
+- Fixed incorrect check-digit weights in `DeSocialSecurityRecognizer` (RVNR); now uses VKVV § 4 weights [2,1,2,5,7,1,2,1,2,1,2,1]. Previous weights diverged from the Deutsche Rentenversicherung specification and rejected the canonical DRV example 15070649C103.
+- Fixed incorrect check-digit algorithm in `DeLanrRecognizer`; now uses KBV Arztnummern-Richtlinie weights [4,9,4,9,4,9] without the spurious Quersumme step, and the complement-to-10 formula `(10 − sum mod 10) mod 10`. Previous weights and formula were internally self-consistent only.
+- Enforced post-2016 BZSt repetition rule in `DeTaxIdRecognizer` (no digit may appear more than three times in positions 1–10).
+
+#### Added
+- ISO 7064 Mod 11,10 structural checksum in `DeVatIdRecognizer`. Algorithm identical to `DeTaxIdRecognizer`; widely used by community validators (python-stdnum, VIES-adjacent).
+- ICAO Doc 9303 MRZ checksum validation in `DePassportRecognizer` and `DeIdCardRecognizer` (weights 7, 3, 1 repeating; letters A=10…Z=35; sum mod 10).
+- KV regional-code whitelist defense-in-depth check in `DeBsnrRecognizer` per KBV Arztnummern-Richtlinie Anlage 1 (no public checksum exists for BSNR).
+
 ## [2.2.362] - 2026-03-15
 ### General
 #### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ All notable changes to this project will be documented in this file.
 - Fixed incorrect check-digit weights in `DeSocialSecurityRecognizer` (RVNR); now uses VKVV § 4 weights [2,1,2,5,7,1,2,1,2,1,2,1]. Previous weights diverged from the Deutsche Rentenversicherung specification and rejected the canonical DRV example 15070649C103.
 - Fixed incorrect check-digit algorithm in `DeLanrRecognizer`; now uses KBV Arztnummern-Richtlinie weights [4,9,4,9,4,9] without the spurious Quersumme step, and the complement-to-10 formula `(10 − sum mod 10) mod 10`. Previous weights and formula were internally self-consistent only.
 - Enforced post-2016 BZSt repetition rule in `DeTaxIdRecognizer` (no digit may appear more than three times in positions 1–10).
+- Registered `DeLanrRecognizer`, `DeBsnrRecognizer`, `DeVatIdRecognizer` and `DeFuehrerscheinRecognizer` in the default registry (previously imported but missing from `conf/default_recognizers.yaml`, so they were unreachable via the default registry).
 
 #### Added
 - ISO 7064 Mod 11,10 structural checksum in `DeVatIdRecognizer`. Algorithm identical to `DeTaxIdRecognizer`; widely used by community validators (python-stdnum, VIES-adjacent).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ All notable changes to this project will be documented in this file.
 #### Added
 - ISO 7064 Mod 11,10 structural checksum in `DeVatIdRecognizer`. Algorithm identical to `DeTaxIdRecognizer`; widely used by community validators (python-stdnum, VIES-adjacent).
 - ICAO Doc 9303 MRZ checksum validation in `DePassportRecognizer` and `DeIdCardRecognizer` (weights 7, 3, 1 repeating; letters A=10…Z=35; sum mod 10).
-- KV regional-code whitelist defense-in-depth check in `DeBsnrRecognizer` per KBV Arztnummern-Richtlinie Anlage 1 (no public checksum exists for BSNR).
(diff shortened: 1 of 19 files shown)


MvdB:
Follow-up after an external review round focused on enterprise-grade correctness, specifically for `DE_VAT_ID` where the BZSt does not publish the Prüfziffer algorithm.

**Problem:** the previous `validate_result` returned `False` on checksum mismatch, which Presidio treats as "drop the match". Given that the ISO 7064 Mod 11,10 we run is community consensus (python-stdnum, VIES-adjacent) rather than a BZSt-normative rule, a real USt-IdNr that happens to diverge from the heuristic was silently lost.

**Changes in `072e8b8`:**

1. `validate_result` is now tri-state. Structural failures (wrong prefix, wrong length after normalisation, non-digit body) still return `False`. Checksum pass returns `True`. **Checksum fail returns `None`** in the default mode — the match keeps its pattern score instead of being dropped.
2. Opt-in strict mode via `strict_checksum=True` constructor parameter for callers who prefer precision over recall (e.g. compliance-adjacent scans).
3. Real-world input normalisation: a second lenient pattern allows whitespace, dashes and dots between digit groups; `validate_result` strips them before structural validation. Handles `DE 123 456 789`, `DE-123-456-789`, `DE.123.456.789`, lowercase and mixed forms.
4. The heuristic nature of the checksum is made explicit in the class docstring and inline so readers of the code see the rejection policy rationale immediately.

**No other DE checksums changed.** KVNR (§ 290 SGB V Anlage 1), RVNR (VKVV § 4), LANR (KBV Arztnummern-Richtlinie), Steuer-IdNr and the ICAO Doc 9303 MRZ checks all have officially documented algorithms and continue to reject on mismatch.

For legally authoritative DE VAT existence verification, callers must still cross-check VIES — this recognizer performs detection, not compliance validation. The class docstring says so explicitly.

Test suite extended from 25 to 56 `DE_VAT_ID` tests covering default-mode tri-state behaviour, strict-mode enforcement, and the real-world formatting variants. Full DE suite: 303 passing. Ruff clean.

MvdB:
Addressed all five Copilot comments — commit mapping for the record:

- **Comment 1** (`test_de_id_card_recognizer.py`): `31341e7` — added per-case `expected_score` to the parametrize and now actually calls `assert_result()` in the loop. nPA with valid ICAO → `MAX_SCORE`; legacy "T + 8 digits" → pattern score 0.5.
- **Comment 2** (`test_de_bsnr_recognizer.py`): `9051ffb` — same parametrize shape; matches the new BSNR scoring behaviour from comment 4.
- **Comment 3** (`de_social_security_recognizer.py` docstring): `2854583` — clarified that the letter expansion yields 12 *data* digits weighted, separate from the check digit at position 12.
- **Comment 4** (`de_bsnr_recognizer.py`): `9051ffb` — `validate_result` no longer promotes whitelisted-prefix matches to `MAX_SCORE`. On a `\b\d{9}\b` base pattern that upgrade was too strong for a 2-digit-prefix check. Returns `None` for every structurally-valid input; context words drive final confidence. `VALID_KV_CODES` is retained as documentation of the expected KV range per KBV Arztnummern-Richtlinie Anlage 1.
- **Comment 5** (`de_passport_recognizer.py`): `def5267` — removed the relaxed `\b[A-Z][A-Z0-9]{7}[0-9]\b` pattern entirely (the strict ICAO charset covers every legitimate German passport) and added a defense-in-depth check in `validate_result` that rejects ICAO-forbidden letters (A, B, D, E, I, O, Q, S, U) outright. Two negative tests added.

Plus `7f9526b` — minor ruff D301 follow-up after the BSNR docstring update (raw docstring since it contains `\b\d{9}\b`).

Full DE suite: 306 passing. Ruff clean.

MvdB:
Second Copilot review round addressed — commit mapping:

- **CHANGELOG.md** (`#3126014708`): `ad16bca` — reworded the `DeBsnrRecognizer` entry to match the actual behaviour. The KV regional-code list is retained for documentation / defense-in-depth purposes per KBV Arztnummern-Richtlinie Anlage 1, but `validate_result` does not reject unknown prefixes (`9051ffb` from the previous round made that deliberate — BSNR has no public checksum, so a 2-digit-prefix drop on a `\b\d{9}\b` base pattern would be too aggressive).
- **`de_social_security_recognizer.py`** (`#3126014784`): `6a9ba8b` — `validate_result` now enforces the VKVV § 4 structural invariants on the birth-day (01–31, or 51–81 with Ergänzungsmerkmal) and month (01–12) *before* running the checksum. Previously, the relaxed `\b\d{8}[A-Z]\d{3}\b` pattern could surface strings with impossible dates and, on a ~10% random chance, have them pass the checksum and be promoted to `MAX_SCORE`. The date check closes that false-positive vector. Four explicit impossible-date cases added to the `validate_result` parametrize.

Full DE suite: 310 passing. Ruff clean.

SharonHart (dismissed):
Amazing contribution, thanks @MvdB!

SharonHart (approved):

omri374 (approved):

PR #1998, opened 2026-04-23
Fix incorrect PESEL checksum validation in PlPeselRecognizer

sienioApius:
## Change Description

The Polish PESEL check-digit algorithm is `check = (10 - weighted_sum % 10) % 10`
(see https://en.wikipedia.org/wiki/PESEL#Check_digit), but
`PlPeselRecognizer.validate_result` currently compares the raw
`weighted_sum % 10` to the check digit, which rejects real valid PESEL numbers.

Example — `44051401458` is the canonical example used in official Polish
documentation and is algorithmically valid, but today:

```python
>>> from presidio_analyzer.predefined_recognizers import PlPeselRecognizer
>>> PlPeselRecognizer().validate_result("44051401458")
False   # expected True
```

This PR:

- Corrects the check-digit formula.
- Guards against non-11-digit / non-numeric inputs (previously would raise
  `IndexError` on a regex-produced string shorter than 11).
- Replaces the existing test fixtures — which were generated using the buggy
  formula and would not pass real-world PESEL validators — with PESEL numbers
  that are valid under the true algorithm, plus negative cases covering wrong
  check digits, wrong length, and non-digit input.

The previous "valid" fixture `11111111114` has a real check digit of `6`, not
`4`. The updated fixtures (`44051401458`, `02070803628`, `11111111116`) verify
against the actual PESEL standard.

## Issue reference

This completes the work started in #1520 (closed on 2025-06-01 due to missing
tests). Attribution to the original author @BlaiseCz for the analysis.

## Checklist

- [x] I have signed the CLA (if required)   *(will sign when CLA bot prompts)*
- [x] All unit tests and lint checks pass locally (`pytest tests/test_pl_pesel_recognizer.py` → 18 passed)

### DIFF
--- a/presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/poland/pl_pesel_recognizer.py
+++ b/presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/poland/pl_pesel_recognizer.py
@@ -44,10 +44,15 @@ class PlPeselRecognizer(PatternRecognizer):
         )
 
     def validate_result(self, pattern_text: str) -> bool:  # noqa: D102
+        if len(pattern_text) != 11 or not pattern_text.isdigit():
+            return False
+
         digits = [int(digit) for digit in pattern_text]
         weights = [1, 3, 7, 9, 1, 3, 7, 9, 1, 3]
 
-        checksum = sum(digit * weight for digit, weight in zip(digits[:10], weights))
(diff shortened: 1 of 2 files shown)


sienioApius:
@microsoft-github-policy-service agree company="APIUS Technologies S.A."

omri374 (approved):

PR #2000, opened 2026-04-24
feat(analyzer): add optional country filter to load_predefined_recogn…

### LINKED ISSUE #1328: Filter recognizers based on locale/country
The number of recognizers in Presidio are growing which is great, but it also means that recognizers are more likely to be irrelevant to some users. Running all recognizers means more processing time and more false positives. 

The ability to provide the `RecognizerRegistry` (or `AnalyzerEngine`) with a country code to either include or filter will allow users to use only a subset of recognizers that is relevant to their tasks

ynachiket:
## Change Description

Adds an optional `countries` parameter to `RecognizerRegistry.load_predefined_recognizers()` that lets callers restrict which country-specific predefined recognizers are loaded, alongside the existing `languages` filter. Locale-agnostic recognizers (credit cards, emails, URLs, crypto, IBAN, etc.) are always preserved regardless of the filter.

**Motivation**

Today, `load_predefined_recognizers()` either loads every country-specific recognizer or none. A US-only or EU-only deployment has to either accept the noise of every country's recognizer, or manually enumerate every recognizer class to reconstruct the registry. This change makes the common "I only care about these locales" case a one-liner.

**Approach**

- Country is inferred from the recognizer's **module path** — the segment directly beneath `country_specific/` (e.g. `us`, `uk`, `es`, `in`, `pl`, `fi`, `sg`, `au`, `it`). No changes to any individual recognizer class are required; the existing directory layout is the source of truth.
- Comparison is **case-insensitive** (`US`, `Us`, `us` all work).
- `countries=[]` keeps only the locale-agnostic recognizers.
- `countries=None` (the default) preserves **exactly** today's behavior — fully backwards compatible. No existing call site needs to change.

Implementation lives in:
- `presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py` — new `RecognizerListLoader._get_recognizer_country()` and `filter_by_countries()` helper.
- `presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry.py` — the new `countries` kwarg threaded through to the loader.

**Example**

```python
from presidio_analyzer import RecognizerRegistry

registry.load_predefined_recognizers(languages=["en"], countries=["us", "uk"])
# -> only US + UK country-specific recognizers, plus all locale-agnostic ones
```

**Tests**

Added 5 unit tests in presidio-analyzer/tests/test_recognizer_registry.py:

default behavior unchanged when countries is not passed
single-country filtering keeps that country + all agnostic recognizers
case-insensitive matching ("US" == "us")
countries=[] keeps only agnostic recognizers
multi-country filtering (["us", "uk"])
All tests in tests/test_recognizer_registry.py pass locally.

Issue reference
Fixes #1328


ynachiket:

omri374:
Thanks! this is great. One comment, which might not be easily solvable- if we're using the module path, it wouldn't catch any custom recognizer that a user would add and filtering on those would not work. Do you have any suggestions on how to include custom country recognizers as well?

ynachiket:
Great call out. 
you're right that module-path inference is a closed-world heuristic that won't catch user-added custom recognizers. I'd like to propose making `country` a first-class property of a recognizer rather than keeping it as an inference rule. Three ways to do that, in rough order of invasiveness:

### Option A — Opt-in `country` attribute on `EntityRecognizer` (recommended)

Add an optional `country: Optional[str] = None` kwarg to `EntityRecognizer.__init__`, default `None` (= locale-agnostic). Predefined country-specific recognizers set it explicitly in their own `__init__` (or we can batch-set it for the `country_specific/*` tree in one commit). Custom recognizers opt in the same way:

```python
class MyBrazilPassportRecognizer(PatternRecognizer):
    def __init__(self):
        super().__init__(
            supported_entity="BR_PASSPORT",
            patterns=[...],
            country="br",
        )
```
The filter then uses getattr(recognizer, "country", None) — no module-path tricks. Module-path inference in this PR becomes a transitional fallback (or gets removed entirely once the predefined recognizers are migrated).

Pros: explicit, discoverable, works uniformly for built-in and custom, plays well with introspection/serialization. Cons: touches the base class (small additive change, fully backwards compatible); predefined recognizers need a one-time sweep to set the attribute.

Option B — Declarative country in the YAML / dict config
Extend PatternRecognizer.from_dict (and the YAML schema under conf/example_recognizers.yaml) to accept a country field, which sets the attribute from Option A:
```python
recognizers:
  - name: "BR Passport"
    supported_language: en
    supported_entity: BR_PASSPORT
    country: br
    patterns: [...]
  ```
   Composes naturally with Option A — same attribute, just a declarative path for ops folks who wire recognizers via config.

Option C — Pluggable country resolver callback
Keep the filter but let callers pass a country_of: Callable[[EntityRecognizer], Optional[str]] so teams with custom recognizers can supply their own classifier without touching the base class:

```python
registry.load_predefined_recognizers(countries=["us", "br"])
registry.filter_by_countries(["us", "br"], country_of=my_resolver)
  ```
Pros: zero changes to recognizer classes. Cons: less discoverable; every user has to write glue code.

What do yo think?

**Comment by omri374** [comment]
Thanks. 1 and 2 fit together (I don't see them as competing alternatives). Once `country_code` is part of the EntityRecognizer class, you don't have to use `getattr(recognizer, "country_code", None)`, you could simply use `recognizer.country_code`. Still, there's are a few open questions:
1. How do we handle generic recognizers? (e.g. credit card, date)
2. There's still the question of discoverabilit. If someone creates a new `br` recognizer but doesn't know about the `country_code` field, they might not get the full list of recognizers they're looking for
3. How do we do this in a backgrward compatible way? i.e. users already have their own list of custom recognizers but did not populate the `country_code` field. Do we assume they don't filter by country anyway? 

Option 3 is interesting, but I agree with you that this requires the user to figure out the mechanism (callable is a generic interface) and use it.

I'm not sure there are great answers to those questions, just wanted to get your perspective and make sure we think about this from multiple angles.

**Comment by ynachiket** [comment]
Some options.. 

### 1. Generic recognizers (credit card, date, email, URL, IBAN, crypto)

**Proposal: `country_code = None` means "locale-agnostic, always included regardless of the filter."**

So `None` is the default in `EntityRecognizer.__init__`, and the filter's rule becomes:
- `country_code is None` → always kept (agnostic).
- `country_code is not None` → kept only if in the requested set.

This keeps the filter's behavior intuitive ("filter country-specific recognizers; leave the generic ones alone") and matches what this PR already does for agnostic built-ins. It also avoids magic sentinels like `"agnostic"` / `"*"`, which tend to leak into configs and get misspelled.

Edge case worth flagging: a few built-ins are *geographically skewed but not country-bound* — IBAN is the usual example (European-ish but not one country), crypto addresses are global-ish. I'd keep those as `country_code=None` (the status quo) rather than introduce `country_code: List[str]`, because:

- A list-valued `country_code` doubles the complexity of the filter logic for one or two edge cases.
- Users who truly want to drop IBAN in a US-only deployment already have a cleaner lever: the `entities=[...]` filter at analyze time, or `registry.remove_recognizer("IbanRecognizer")`.

If that trade-off bites enough users later, promoting `country_code` to `Optional[Union[str, List[str]]]` is an additive, backwards-compatible change for a future PR.

### 2. Discoverability

You're right that this is the weakest link of any opt-in metadata field. I don't think it's fully solvable — we can only *reduce* the failure mode. Concretely, I'd layer four things:

1. **Clear docstring + type hint** on `EntityRecognizer.__init__` explaining the field and linking to the "filter by country" doc section.
2. **Migrate `country_specific/*` built-ins to set `country_code` explicitly**, so every predefined country-specific recognizer is a worked example. Anyone reading those for inspiration sees the pattern.
3. **A `registry.get_country_codes() -> Set[str]`** helper so users can `print(registry.get_country_codes())` in a REPL and immediately see what's tagged. Cheap to add, big debugging win.
4. **Runtime feedback when the filter "misses"**: if a user passes `countries=["br"]` and the filter removes zero recognizers from the BR bucket (i.e. no recognizer had `country_code="br"`), log a WARNING listing what *was* found, plus a one-line hint: *"If you have custom BR recognizers, set `country_code='br'` on them to include in filtering."*

That combination won't be perfect, but it'll catch the common footgun where someone filters, sees a short list, and silently wonders why their custom recognizer is missing.

### 3. Backwards compatibility for untagged custom recognizers

This is the most consequential question, and I think the cleanest answer is the one implied by the rule in (1):

**An untagged recognizer (`country_code is None`) is treated as locale-agnostic and is always kept**, regardless of the requested countries. Same rule as a built-in generic recognizer.

To your phrasing — *"do we assume they don't filter by country anyway?"* — yes, exactly. A user who hasn't adopted the field hasn't opted into the filtering contract, and the safest interpretation is "this recognizer applies everywhere." That has the nice property of giving us **one rule** for both generics and untagged custom recognizers — easy to document, easy to reason about.

The obvious failure mode: a user has a custom `BrPassportRecognizer` without `country_code` set, calls `countries=["us"]`, and their BR recognizer incorrectly sticks around. That's a real but **visible** problem — the recognizer fires on the wrong input, they notice, they grep the docs for "country", and they find the field. It's a much less dangerous failure mode than the alternative (strict: drop all untagged recognizers), which would silently break existing registries the moment someone turns on the filter.

If we want to nudge harder, option 2.4 above (runtime warning when the filter returns zero hits for a requested country) covers the common case where they pass a country they expected to match.

### Summary of proposed defaults

| Scenario | `country_code` | Filter behavior |
|---|---|---|
| Built-in generic (CC, date, email, URL, IBAN, crypto) | `None` (unchanged) | Always included |
| Built-in country-specific (after migration) | `"us"`, `"uk"`, etc. | Included iff in `countries` |
| Custom recognizer with no opt-in | `None` | Always included (same as generic) |
| Custom recognizer that opts in | user-set | Included iff in `countries` |
| `countries=None` (default) | any | All included (identical to today) |
| `countries=[]` | any | Only `country_code is None` kept |

Net effect:
- Zero breaking change for anyone not passing `countries=`.
- Zero breaking change for anyone passing `countries=` who hasn't tagged custom recognizers (they just continue to run).
- Gradual opt-in: as folks tag their custom recognizers, filtering gets more precise.

Happy to do this as a single follow-up PR once this one lands (keeping this PR as the "module-path inference for the built-ins" step, which is then superseded by the attribute in PR #2), or collapse both into this PR if you'd prefer one bigger change. Slight lean toward the two-PR path because PR #2 will touch every file under `country_specific/` and I'd rather keep the review surfaces separable. Your call.

**Comment by omri374** [comment]
Thanks! 
So calling with `None` returns all recognizers, country specific or not, and calling with a country code returns all recognizers with this country value + all recognizers with None. Right? 

Please also add a doc section + update the recognizers yaml + contributing.md to make sure we set this rule properly from now on.

**Comment by ynachiket** [comment]
Confirmed — that's exactly the rule:

- `countries=None` (default) → everything is kept (country-specific + agnostic / untagged). Identical to today's behavior.
- `countries=["us"]` → recognizers with `country_code="us"` **plus** recognizers with `country_code=None` (i.e. generic built-ins *and* custom recognizers that haven't opted in).
- `countries=[]` → only `country_code=None` kept.

The one-line invariant is "untagged = always included," which is what keeps the gradual migration backwards-compatible.

On the three asks:

**Doc section** — a new "Filtering recognizers by country" page (or subsection of analyzer customization). Will cover the `country_code` attribute on `EntityRecognizer`, the three `countries=` cases above, and a short troubleshooting note for "I filtered and my custom recognizer disappeared" (answer: set `country_code` on it).

**Recognizers YAML** — annotate the predefined entries under `country_specific/*` with `country_code: <iso>`, plus one example of a custom recognizer declaring `country_code` so the YAML doubles as a worked example.

**CONTRIBUTING.md** — short subsection under "adding a new recognizer": any new country-specific recognizer must declare `country_code` (matching its `country_specific/<iso>/` directory); generic recognizers leave it as `None`.

One scope question before I start: the doc / YAML / CONTRIBUTING work only really makes sense once `country_code` is a first-class attribute on `EntityRecognizer` (currently this PR is the module-path heuristic only). Two ways to handle that:

- **A — single PR**: add the `country_code` attribute + migrate the predefined recognizers + all three docs into this PR.
- **B — two PRs**: land this one as-is, follow-up PR adds the attribute + migration + the three docs.

Slight lean toward **B** for review hygiene (the migration touches every file under `country_specific/`, which I'd rather keep separable from the filter logic). Happy with **A** if you'd rather have everything together. Your call.

**Comment by omri374** [comment]
Thanks! Sounds like a plan! let's go with A to not introduce changes and then remove them. It's ok if the PR is slightly big because it touches every predefined country recognizer.

**Comment by ynachiket** [comment]
Thanks for the steer! Pushed `54d0644` implementing **plan A** end-to-end.

**`EntityRecognizer`**
- Added optional `country_code` constructor arg + instance attribute (lower-cased ISO-3166 alpha-2 by convention; `None` = generic).
- Added `COUNTRY_CODE: ClassVar` so subclasses can declare country once, without overriding `__init__`.
- `to_dict()` now serializes `country_code` when set.

**`PatternRecognizer`**
- Forwards a `country_code` kwarg to the base class so YAML-defined custom recognizers can declare it too.

**Predefined recognizers**
- All 60 country-specific predefined recognizers now declare `COUNTRY_CODE` as a class-level attribute (au, ca, de, es, fi, in, it, kr, ng, pl, se, sg, th, tr, uk, us). Generic recognizers (credit card, email, phone, IP, IBAN, dates, …) intentionally remain unset.

**`RecognizerRegistry`**
- `_get_recognizer_country` now prefers `recognizer.country_code`; legacy module-path inference is kept as a fallback for any user-defined recognizer that hasn't adopted the attribute, so existing custom registries keep working.
- `_prepare_recognizer_kwargs` drops `country_code` from kwargs when the target class's `__init__` doesn't accept it (avoids `TypeError` on predefined recognizers that don't take it through their constructor).
- `filter_by_countries` now keeps generic (`None`) recognizers and filters only country-tagged ones; emits a `WARNING` log when a requested country matches zero recognizers (discoverability).
- New `RecognizerRegistry.get_country_codes()` helper returns the set of country codes currently loaded.

**Behavior summary**
- `load_predefined_recognizers(countries=None)` → all recognizers (country-tagged + generic). Unchanged.
- `load_predefined_recognizers(countries=["us"])` → US-tagged recognizers + all generic recognizers.
- Backwards compatible: any recognizer without `country_code` is treated as generic and is always returned.

**Config / YAML**
- `example_recognizers.yaml` documents the new `country_code` field and adds a `BR CPF` recognizer that sets `country_code: "br"` to demonstrate custom usage.

**Docs**
- New page: `docs/analyzer/filtering_by_country.md` covering how the filter works, declaring `country_code` on custom recognizers (class attribute, constructor arg, YAML), when to leave it unset, backwards-compat, and debugging tips.
- `docs/analyzer/index.md` and `mkdocs.yml` link to the new page.
- `docs/analyzer/adding_recognizers.md` documents the `COUNTRY_CODE` convention for predefined recognizers.
- `CONTRIBUTING.md` adds a "Contributing a New Predefined Recognizer" rule requiring `COUNTRY_CODE` for country-specific additions.

**Tests**
- Extended `tests/test_recognizer_registry.py` with coverage for attribute-based filtering, untagged/generic recognizers, the zero-match warning path, and `get_country_codes()`.

**CHANGELOG**
- Entry under `[unreleased]` → `Analyzer` → `Added`.

Local verification: `2126 passed, 60 skipped` on `presidio-analyzer/tests` (the skipped/ignored modules are gated on optional dependencies — `azure_ai_language`, `azure_auth_helper`, `ahds`, `langextract`, `nlp_engine_provider`, `slim_spacy_nlp_engine` — and aren't touched by this change). `ruff check` and `ruff format --check` are clean on the modified files.

Diff is ~73 files / +695/-26 as you anticipated. Happy to split it (e.g., base attribute + filter logic in one commit, predefined-recognizer migration in a follow-up) if that helps review — just say the word.

File: docs/analyzer/filtering_by_country.md [hunk 1]
```diff
@@ -0,0 +1,182 @@
+# Filtering recognizers by country
```
omri374, line 1: Great doc!!

presidio-analyzer/presidio_analyzer/entity_recognizer.py
```diff
@@ -29,10 +29,22 @@ class EntityRecognizer:
...
+    ``COUNTRY_CODE`` attribute (e.g. ``COUNTRY_CODE = "us"``); the constructor
+    argument exists for ad-hoc / custom recognizers and overrides the class
+    attribute when both are provided. Codes are compared case-insensitively
+    and should follow ISO 3166-1 alpha-2 (with ``"uk"`` accepted as a synonym
+    of ``"gb"`` for historical consistency with the directory layout).
     """
 
     MIN_SCORE = 0
     MAX_SCORE = 1.0
+    COUNTRY_CODE: ClassVar[Optional[str]] = None
```
omri374, line 46: Having a class var is definitely the right way here, but how do we make this discoverable? having half the recognizers implement it and have keep it as None doesn't take us very far. Any suggestions?
I was thinking along those lines maybe:
```py
from typing import ClassVar

class BaseRecognizer:
    COUNTRY_CODE: ClassVar[str | None] = None

    @classmethod
    def country_code(cls) -> str | None:
        return cls.COUNTRY_CODE

    @classmethod
    def is_country_specific(cls) -> bool:
        return cls.COUNTRY_CODE is not None
```

WDYT?

@SharonHart, how do you see this?

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry.py
```diff
@@ -110,6 +129,11 @@ def load_predefined_recognizers(
         )
         recognizers = RecognizerListLoader.get(**configuration)
 
+        if countries is not None:
+            recognizers = RecognizerListLoader.filter_by_countries(
```
omri374: instead of introducing this logic here, can we inject the country codes to the configuration and have the RecognizerListLoader do all the work, like we do with languages? 
This essentially means that the country filter is now global in the yaml, which I feel is a step in the right direction, similar to how languages are filtering recognizers.

omri374 (commented):
Added a few comments, thanks!

presidio-analyzer/presidio_analyzer/entity_recognizer.py
ynachiket: - EntityRecognizer now exposes COUNTRY_CODE: ClassVar[Optional[str]] = None as the single canonical declaration.
- Added @classmethod country_code(cls) -> Optional[str] (lower-cases the class attr) and @classmethod is_country_specific(cls) -> bool.
- Removed the country_code constructor kwarg, the self.country_code instance attribute, the country_code field in to_dict(), and the _prepare_recognizer_kwargs shim that swallowed the YAML kwarg.
- Country tagging is now intentionally a class-level fact: instance overrides are a no-op, so COUNTRY_CODE is hard to miss when reading the registry / docs / a recognizer class definition.

ynachiket: Will wait for Sharon's POV on the YAML aspect

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry.py
ynachiket: - RecognizerListLoader.get(...) gained supported_countries: Optional[Iterable[str]] = None. Filtering happens inline next to _is_language_supported_globally.
- RecognizerRegistry.load_predefined_recognizers(countries=...) simply forwards into registry_configuration["supported_countries"]. The post-hoc filter_by_countries call site in recognizer_registry.py is gone.
- The filter is now also driveable via a top-level supported_countries: ["us", "uk"] field in the registry YAML, mirroring supported_languages. Documented in docs/analyzer/filtering_by_country.md and conf/example_recognizers.yaml.

ynachiket:
Thanks for the review @omri374 — `1375fbb` addresses both threads. Summary:

**1. Discoverability — went with the classmethod shape, but made it the *only* path.**

Rather than `COUNTRY_CODE` ClassVar + a parallel `country_code` instance attribute (which would have given two competing APIs), `COUNTRY_CODE` is now the single source of truth at the class level. The `country_code` constructor kwarg, instance attribute, `to_dict` field, and the `_prepare_recognizer_kwargs` shim that swallowed the YAML kwarg are all removed.

`EntityRecognizer` exposes:

```python
COUNTRY_CODE: ClassVar[Optional[str]] = None

@classmethod
def country_code(cls) -> Optional[str]:
    code = cls.COUNTRY_CODE
    return code.lower() if isinstance(code, str) else code

@classmethod
def is_country_specific(cls) -> bool:
    return cls.country_code() is not None
    
The filter logic now reads recognizer.country_code() everywhere; instance-level overrides are intentionally a no-op so country tagging is unambiguously a fact about the recognizer class. That pushes the field's discoverability up the stack — every code path that cares about country lands on is_country_specific() / country_code(), which makes the missing-attribute case much harder to overlook in review and at runtime.

2. Filter is now threaded through RecognizerListLoader exactly like supported_languages.

RecognizerListLoader.get(...) gained supported_countries: Optional[Iterable[str]] = None. The filter runs inline alongside _is_language_supported_globally, no separate code path.
The configuration loader threads it automatically, so the same filter is also driveable from the top-level YAML (supported_countries: ["us", "uk"]) the same way supported_languages is.
RecognizerRegistry.load_predefined_recognizers(countries=...) is now just a thin pass-through into registry_configuration["supported_countries"]. The post-hoc filter_by_countries(...) call site in recognizer_registry.py is gone.

File: presidio-analyzer/presidio_analyzer/entity_recognizer.py (hunk 2 above)
Comment (ynachiket) [reply presidio-analyzer/presidio_analyzer/entity_recognizer.py:46]: Let me know how you want to proceed.

**Comment by omri374** [comment]
Thanks. busy week, we'll review asap

File: presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry.py [hunk 4]
```diff
@@ -91,19 +91,44 @@ def add_nlp_recognizer(self, nlp_engine: NlpEngine) -> None:
...
 
         registry_configuration = {"global_regex_flags": self.global_regex_flags}
         if languages is not None:
             registry_configuration["supported_languages"] = languages
+        if countries is not None:
+            # Threaded through the configuration the same way as
+            # ``supported_languages`` so the filter is applied inside
+            # ``RecognizerListLoader.get(...)`` and behaves uniformly
+            # whether driven from Python or from a YAML config file.
+            registry_configuration["supported_countries"] = countries
```
omri374, line 131: Please update default_recognizers.yaml with the country tag too. This would make this more explicit for those using the no-code yaml approach

presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/italy/it_passport_recognizer.py
```diff
@@ -13,6 +13,8 @@ class ItPassportRecognizer(PatternRecognizer):
     :param supported_entity: The entity this recognizer can detect
     """
 
+    COUNTRY_CODE = "it"
```
omri374, line 16: please add a country code for IT Vat code too (https://github.com/ynachiket/presidio/blob/main/presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/italy/it_vat_code.py)

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry.py
omri374: Also make sure that there is input validation. What if the user inputs two countries? (actually, in some cases this could be applicable, like a phone number recognizer) but for now let's put it aside. Worst case, the user could define two recognizers one for each country.

omri374:
@ynachiket apologies for the delay. I've added a few minor comments but it's mostly ready. Thanks!

ynachiket:
Thanks @omri374! Pushed 47e01c4 addressing all three threads.

**1. `default_recognizers.yaml` country tags.** Every predefined country-specific entry now carries `country_code: <iso>` next to `enabled:` / `supported_languages:`. The field is advisory metadata for no-code YAML readers — the class-level `COUNTRY_CODE` ClassVar remains the single source of truth. The loader strips the YAML field from instantiation kwargs and **cross-checks it against the class attribute** in a new `RecognizerListLoader._validate_yaml_country_code` helper. Mismatches (YAML says `us`, class says `uk`, or YAML declares `country_code` on a class that has no `COUNTRY_CODE`) raise `ValueError` at load time naming both values, so drift is impossible to introduce silently.

**2. Input validation.** `RecognizerListLoader.filter_by_countries` (and transitively `RecognizerRegistry.load_predefined_recognizers(countries=...)` and `RecognizerListLoader.get(supported_countries=...)`) now reject the common footguns up front through a new `_normalize_countries` helper:
- Bare `str` (`countries="us"`) → `TypeError` (would otherwise iterate over chars and silently match nothing).
- Non-iterable scalars → `TypeError`.
- Non-string elements → `TypeError` naming the offender.
- Blank / whitespace-only codes → `ValueError`.
- Whitespace stripped, codes lower-cased.

I left multi-country-per-recognizer aside per your note; happy to revisit that as a follow-up if `phone_recognizer` ever wants it (the natural shape would be promoting `COUNTRY_CODE` to `Optional[Union[str, FrozenSet[str]]]`, fully additive).

**3. `COUNTRY_CODE` for `ItVatCodeRecognizer`** — added (`"it"`). While I was there I also caught `EsPassportRecognizer`, which was added to main during this PR's review cycle and never went through the original migration sweep — set to `"es"`.

**Tests.** 7 new tests in `test_recognizer_registry.py`:
- 5 for input validation (bare `str`, non-iterable, non-string element, blank element, normalization, end-to-end through `load_predefined_recognizers`).
- 4 for YAML cross-validation (including a sanity test that walks every `country_code:` in the shipped `default_recognizers.yaml` and asserts it matches the corresponding class — guards against future drift).

**Docs.** New "Annotating predefined recognizers in YAML" section in `docs/analyzer/filtering_by_country.md`; CONTRIBUTING.md predefined-recognizer checklist now includes the YAML `country_code:` step; CHANGELOG entry extended.

Local: 169 tests pass across `recognizer_registry`, `recognizer_registry_provider`, `analyzer_engine`, `pattern_recognizer`, `it_vat_code_recognizer`, `es_passport_recognizer`. `ruff format --check` clean on touched production files; `ruff check` is net-improved vs. baseline (only pre-existing D103/D205/E501 in the test file remain).

omri374:
Hey @ynachiket, first off- thank you so much for your patience and for sticking with this through a longer-than-expected design discussion (this is also PR 2000 for the repo! Congrats!). The work you've done here is solid and has been really valuable in helping us think through the right approach. Truly appreciated. 🙏
 
I'd like to revisit the decision on country_code as a constructor parameter again (in addition, not instead of the existing approach).

Your implementation correctly handles the YAML metadata case for predefined recognizers, and we'd like to build on top of it to unlock one more scenario before merging: letting users tag their own custom recognizers with a country code, whether they're defining them in YAML, in Python, or via `from_dict`,  without having to subclass and set `COUNTRY_CODE` on a class.

#### The gap

Today, a user who writes a regex for an Armenian national ID and wants it filtered by supported_countries has two options:
1. Subclass `PatternRecognizer`, set `COUNTRY_CODE = "am"` on the class works, but requires Python code.
Add `country_code: am` to their YAML custom recognizer entry: currently stripped at load time and has no effect on filtering.
We'd like both to work. The YAML path (option 2) is the main motivation.

#### The proposed addition

Add `country_code: Optional[str] = None` to `EntityRecognizer.__init__` (threaded through `PatternRecognizer`), with the following rules at construction time:
- COUNTRY_CODE ClassVar is set and the passed value conflicts → ValueError. A Polish tax ID recognizer is always Polish; passing country_code="uk" is a misconfiguration and should fail loudly.
- COUNTRY_CODE ClassVar is set and the passed value matches → fine (redundant but harmless).
- COUNTRY_CODE ClassVar is None (locale-agnostic class) → the passed value is assigned to the instance. This is the new capability.
`country_code()`changes from a `@classmethod` to a regular instance method reading `self._country_code`. The COUNTRY_CODE ClassVar stays exactly as-is as the canonical class-level declaration. This is an addition, not a replacement.

The key benefit of threading it through the constructor: `PatternRecognizer.from_dict({"country_code": "am", ...})` just works, which means YAML country_code: on a custom recognizer flows through naturally. The `_tag_recognizer_with_country` helper and the extract-before-strip logic in the loader can be removed in favor of this cleaner path.

**Pros**

- Uniform API: YAML users, Python users, and from_dict users all go through the same constructor path — no special-casing in the loader.
- Predefined recognizers stay protected: a class with COUNTRY_CODE = "pl" cannot be silently re-tagged with a conflicting value.
- Less loader complexity: removes the dynamic subclass patch and the pre-strip extraction.
Serialisation works correctly: to_dict() can emit country_code and from_dict round-trips without data loss.

**Cons / tradeoffs**
- country_code() is no longer callable on an uninstantiated class (RecognizerClass.country_code() → TypeError). The one internal call site that does this (_validate_yaml_country_code) needs to switch to reading. 
- RecognizerClass.COUNTRY_CODE directly, and any external callers would face the same. The ClassVar itself remains accessible as always.
- Minor test updates: a couple of tests that explicitly assert the class-level-call contract need adjusting.
 
I realize this extends the scope of your PR a bit, and I'm genuinely happy to discuss. Does this direction make sense to you? Are there aspects of the current approach you'd prefer to keep? Would love your thoughts before we decide how to proceed.

ynachiket:
Thanks @omri374 — and back at you on the patience! And yeah, the gap you're flagging is real: subclass-or-nothing for the `type: custom` path was a cliff, and threading `country_code` through the constructor cleanly removes it. Happy to extend scope.

Here's my read of the design before I touch code, so we can lock alignment in one round-trip:

**Construction-time rules** (`country_code: Optional[str] = None` on `EntityRecognizer.__init__`, threaded through `PatternRecognizer`):

| `cls.COUNTRY_CODE` | `country_code=` kwarg | Result |
| --- | --- | --- |
| `"us"` | `None` | `self._country_code = "us"` (today) |
| `"us"` | `"us"` (match, case-insensitive) | `self._country_code = "us"` (fine) |
| `"us"` | `"uk"` (conflict) | `ValueError` naming both values |
| `None` | `"am"` | `self._country_code = "am"` — **new** |
| `None` | `None` | `self._country_code = None` (locale-agnostic) |

**API shape:**
- `country_code()` and `is_country_specific()` become regular instance methods reading `self._country_code`.
- `to_dict()` emits `country_code` whenever resolved non-None, so `from_dict(to_dict(rec))` is lossless for both ClassVar-tagged and constructor-tagged recognizers.
- `COUNTRY_CODE` ClassVar stays exactly as is — the canonical class-level declaration.
- `_validate_yaml_country_code` switches to reading `recognizer_cls.COUNTRY_CODE` directly (per your note).

**Three specifics I'd like to confirm before I commit, so we don't bounce:**

1. **Predefined recognizer `__init__`s.** They don't currently accept `country_code` (e.g. `UsSsnRecognizer.__init__(self, patterns=…, supported_language=…)`), and their tag is unambiguously class-level. Two paths:
   - **(a) minimal**: keep the existing strip + `_validate_yaml_country_code` cross-check in the loader for `type: predefined` entries; **stop stripping** for `type: custom` so the constructor kwarg path picks them up naturally. The custom YAML route is now fully constructor-driven; predefined stays loader-driven (which it has to be, since `supported_languages` etc. work the same way).
   - **(b) uniform**: thread `country_code` through every predefined `__init__` so the kwarg path is the single one. Touches ~60 files for no behavioral gain — predefined classes can't legally accept a non-matching code anyway, the constructor would just `ValueError`.
   
   I'd lean **(a)** unless you want the uniformity badly enough to absorb the diff. Either way, custom recognizers go entirely through the constructor.

2. **`to_dict()` policy.** Emit `country_code` for **any** instance where it resolved non-None (uniform across predefined-via-ClassVar and custom-via-kwarg)? Or only when set explicitly via the constructor (special-cased)? I'd argue uniform — predefined classes reconstruct the same value from their ClassVar regardless, so emitting it is redundant-but-harmless and avoids a code-path split. `from_dict` round-trip is lossless either way for predefined; for custom, uniform is the one that actually matters.

3. **Constructor kwarg normalization.** Mirror the YAML / `countries=` validation we just added: strip + lower-case + reject blank/non-string. So `country_code=" AM "` is stored as `"am"`, `country_code=""` raises `ValueError`. Sound right?

**Test/doc deltas I'm planning** (calling them out so nothing surprises you in the diff):
- Replace `test_country_code_is_a_class_level_fact_not_instance` (asserts instance overrides are a no-op — that contract is changing) and update `test_country_code_classmethod_lowercases_class_attribute` to call on an instance.
- Add tests for the construction-time matrix, `from_dict` round-trip, custom `PatternRecognizer(country_code="am", …)`, custom YAML (`type: custom`) flowing through the filter end-to-end, and the `_country_code` normalization edge cases.
- `docs/analyzer/filtering_by_country.md`: "Declaring a country" section gets a third path (constructor kwarg / YAML on custom) with an Armenian worked example.
- Extend the CHANGELOG entry; CONTRIBUTING.md is already correct (it points at `COUNTRY_CODE` for predefined, which doesn't change).

If (a), uniform `to_dict`, and constructor normalization all match what you had in mind, I'll get a commit up. Otherwise let me know which knobs you want flipped and I'll adjust.

omri374:
@ynachiket I agree with all your comments and decisions. Thanks!

ynachiket:
Pushed d0bbcd2 implementing the constructor-kwarg path end-to-end, exactly per the plan we agreed on. Tagging now reconciles cleanly:

| `cls.COUNTRY_CODE` | `country_code=` kwarg | result |
| --- | --- | --- |
| `None` | `None` | locale-agnostic |
| `None` | `"am"` | `self._country_code = "am"` (the new capability — used by YAML `type: custom`) |
| `"us"` | `None` | `self._country_code = "us"` (today) |
| `"us"` | `"us"` (case-insensitive match) | `self._country_code = "us"` (redundant but fine) |
| `"us"` | `"uk"` | `ValueError` naming both values |

`country_code()` and `is_country_specific()` are now instance methods reading `self._country_code` (resolved once at construction time and cached). The class-level `COUNTRY_CODE` ClassVar stays accessible as `cls.COUNTRY_CODE` for callers that need to introspect without instantiating. `to_dict()` emits `country_code` whenever the resolved tag is non-None, so `from_dict(to_dict(rec))` round-trips losslessly for both class-tagged and constructor-tagged recognizers.

**Loader.** Per option (a) in the plan: `predefined_to_exclude` keeps stripping `country_code` (predefined `__init__`s don't accept the kwarg), `_validate_yaml_country_code` reads `recognizer_cls.COUNTRY_CODE` directly. `custom_to_exclude` no longer strips the field, so YAML `type: custom` entries with `country_code:` flow straight through `from_dict` → `PatternRecognizer.__init__` → `EntityRecognizer.__init__` and land on `self._country_code`.

**Tests.** Three contracts flipped, so I replaced those tests:
- `test_country_code_classmethod_lowercases_class_attribute` → `test_country_code_lowercases_class_attribute` (instance-method shape).
- `test_predefined_country_specific_recognizer_carries_class_country_code` now reads `UsSsnRecognizer.COUNTRY_CODE` for the no-instantiation path and asserts the instance-method form on a constructed instance.
- `test_country_code_is_a_class_level_fact_not_instance` (asserted instance overrides were a no-op — that contract is gone) → `test_country_code_class_attribute_is_authoritative_over_instance_attr`, which locks in "resolved once at construction, cached on the instance, post-construction `rec.COUNTRY_CODE = ...` is a no-op."
- `test_to_dict_does_not_carry_country_code` → split into `test_to_dict_omits_country_code_when_unset` and `test_to_dict_emits_class_level_country_code`.

New tests for the construction matrix (7 cases including blank/non-string rejection, case-insensitive match, conflict-raises), `from_dict` round-trip preserves the tag, and an end-to-end YAML test that loads a `type: custom` Armenian recognizer through `RecognizerConfigurationLoader` / `RecognizerListLoader.get` and asserts it survives `supported_countries=["am"]` and is dropped under `supported_countries=["uk"]`.

**Docs.** `filtering_by_country.md` now has three numbered subsections under "Declaring a country on your own recognizers" — subclass + ClassVar (Brazilian CPF), constructor kwarg (Armenian one-off Python), and YAML `type: custom` (Armenian no-code) — plus a "Combining paths" note covering the conflict-raises invariant. The "How the filter works" section is rewritten to describe both tagging paths and the reconciliation matrix. CHANGELOG entry extended.

Local: 234 tests pass across recognizer_registry, recognizer_registry_provider, analyzer_engine, pattern_recognizer, pattern, it_vat_code, es_passport, us_ssn, pl_pesel, uk_nino, es_nif. `ruff check` and `ruff format --check` clean on all touched production files; test-file lint is net-improved vs. the prior commit.

Ready for another look — let me know if anything else.

presidio-analyzer/presidio_analyzer/entity_recognizer.py
```diff
@@ -29,10 +29,45 @@ class EntityRecognizer:
     :param version: the recognizer current version
     :param context: a list of words which can help boost confidence score
     when they appear in context of the matched entity
+
+    To declare a recognizer as country-specific there are two paths:
+
+    - **Class-level (preferred for predefined recognizers):** override the
```
omri374: Please shorten this, and put it under :param country_code and not above.

presidio-analyzer/tests/test_recognizer_registry.py
```diff
@@ -245,3 +244,715 @@ def test_remove_recognizer_when_multiple_instances_exist():
...
+
+
+def test_predefined_country_specific_recognizer_carries_class_country_code():
+    """Predefined country recognizers expose their class-level ``COUNTRY_CODE``.
+
+    The country tag is reconciled at construction time and surfaced via
+    the instance's ``country_code()`` method; the ClassVar itself remains
+    accessible for class-level introspection without instantiating.
+    """
+    from presidio_analyzer.predefined_recognizers import UsSsnRecognizer
```
omri374: Please put imports on top

presidio-analyzer/tests/test_recognizer_registry.py
```diff
@@ -245,3 +244,715 @@ def test_remove_recognizer_when_multiple_instances_exist():
...
+
+def test_yaml_country_code_on_locale_agnostic_class_raises():
+    """YAML ``country_code`` on a class without ``COUNTRY_CODE`` raises.
+
+    The filter has no class-level fact to anchor on. The error message
+    points at the fix (set ``COUNTRY_CODE`` on the class, or remove the
+    YAML field).
+    """
+    from presidio_analyzer.predefined_recognizers import CreditCardRecognizer
+    from presidio_analyzer.recognizer_registry.recognizers_loader_utils import (
```
omri374: please move all the tests testing RecognizerLoaderUtils and RecognizerListLoader until test_recognizer_loader_utils.py. or test_configuration_validator. They focus on testing the yaml loading and not the recognizer registry logic. Thanks!

omri374 (commented):
Thanks! Most of the logic seems solid, but there are still tests missing for the YAML definitions. 

What happens if a user enters the wrong country code in the yaml? what if they enter multiple? 
Please see my comments on tests and add more tests to validate that the YAML path works as expected. For example, I think you need to add country_code to the CustomRecognizerConfig so that it is set properly.

ynachiket:
Thanks @omri374 — pushed `e6312af4` addressing the YAML `country_code` validation gap.

Summary:
- Added `country_code` to `CustomRecognizerConfig`, normalized to a single stripped/lower-cased string.
- Blank values now fail validation instead of being ignored.
- Multiple values like `country_code: [us, uk]` fail validation; the field remains intentionally single-country for now.
- Added registry-model coverage proving validated custom recognizer YAML preserves `country_code` through `model_dump()`, so it can flow into the loader/from_dict path.

Validation:
- `python -m pytest presidio-analyzer/tests/test_yaml_recognizer_models.py presidio-analyzer/tests/test_recognizers_loader_utils.py presidio-analyzer/tests/test_recognizer_registry.py presidio-analyzer/tests/test_recognizer_registry_provider.py presidio-analyzer/tests/test_pattern_recognizer.py -q` → 175 passed
- `git diff --check` → passed

I also checked `ruff check` on the touched files; it still reports pre-existing lint in `test_yaml_recognizer_models.py` outside this diff, so I did not broaden the PR to clean unrelated style issues.

omri374 (approved):
Looks great, thanks!!

ynachiket:
Need anything from me (squash?) before we can merge?

omri374:
No, thanks! I approved but also asked @SharonHart to take a look. If he doesn't have the capacity, we'll merge later today or tomorrow.

PR #2007, opened 2026-04-28
fix: Preserve GLiNER config fields through Pydantic validation

yuriihavrylko:
## Change Description

Add GLiNERRecognizerConfig to the YAML recognizer config model and register it in `CONFIG_MODEL_MAP`, so that model_name and other GLiNER-specific fields (`flat_ner`, `threshold`, `multi_label`, `map_location`, `load_onnx_model`, `onnx_model_file`, `entity_mapping`) are preserved through Pydantic validation when loading recognizer configuration from YAML.                                            
                                                                                                                                    
**Root cause**: When parsing a GLiNERRecognizer entry from YAML, parse_recognizers looked up "GLiNERRecognizer" in `CONFIG_MODEL_MAP` and found nothing, falling back to PredefinedRecognizerConfig. Since that class has no model_name field and Pydantic's default is to ignore extra fields, `model_name` (and all other GLiNER params) were silently dropped. The recognizer would then always load the hardcoded default model (`urchade/gliner_multi_pii-v1`) regardless of what was specified in config.  HuggingFaceNerRecognizer already had this handled correctly via HuggingFaceRecognizerConfig.

## Issue reference

Fixes #XX

## Checklist

### DIFF
--- a/presidio-analyzer/presidio_analyzer/input_validation/yaml_recognizer_models.py
+++ b/presidio-analyzer/presidio_analyzer/input_validation/yaml_recognizer_models.py
@@ -175,6 +175,16 @@ class HuggingFaceRecognizerConfig(PredefinedRecognizerConfig):
         default=None, description="Prefixes to strip from labels (e.g. B-, I-)"
     )
 
+    def model_dump(self, *args, **kwargs) -> Dict[str, Any]:
+        """Serialize the config without None values by default.
+
+        HuggingFace recognizer kwargs are passed directly to the recognizer constructor.
+        Excluding None values preserves constructor defaults for omitted YAML fields
+        instead of overriding them with explicit None.
+        """
+        kwargs.setdefault("exclude_none", True)
+        return super().model_dump(*args, **kwargs)
+
 
 class GLiNERRecognizerConfig(PredefinedRecognizerConfig):
     """Configuration specifically for GLiNER models."""
--- a/presidio-analyzer/presidio_analyzer/input_validation/yaml_recognizer_models.py
+++ b/presidio-analyzer/presidio_analyzer/input_validation/yaml_recognizer_models.py
@@ -157,7 +157,7 @@ class PredefinedRecognizerConfig(BaseRecognizerConfig):
 class HuggingFaceRecognizerConfig(PredefinedRecognizerConfig):
     """Configuration specifically for HuggingFace NER models."""
 
-    model_config = ConfigDict(extra="forbid")
+    model_config = ConfigDict(extra="allow")
 
     model_name: Optional[str] = Field(None, description="HuggingFace model name")
     tokenizer_name: Optional[str] = Field(
@@ -189,7 +189,7 @@ class HuggingFaceRecognizerConfig(PredefinedRecognizerConfig):
 class GLiNERRecognizerConfig(PredefinedRecognizerConfig):
     """Configuration specifically for GLiNER models."""
 
-    model_config = ConfigDict(extra="forbid")
+    model_config = ConfigDict(extra="allow")
 
     model_name: Optional[str] = Field(None, description="GLiNER model name")
     flat_ner: Optional[bool] = Field(None, description="Use flat NER")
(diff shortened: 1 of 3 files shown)


yuriihavrylko:

presidio-analyzer/presidio_analyzer/input_validation/yaml_recognizer_models.py
```diff
@@ -176,6 +176,23 @@ class HuggingFaceRecognizerConfig(PredefinedRecognizerConfig):
...
+    multi_label: Optional[bool] = Field(
+        None, description="Use multi-label classification"
+    )
+    threshold: Optional[float] = Field(None, description="Confidence threshold")
+    map_location: Optional[str] = Field(None, description="Device (cpu/gpu/etc.)")
+    load_onnx_model: Optional[bool] = Field(None, description="Load ONNX model")
+    onnx_model_file: Optional[str] = Field(None, description="ONNX model file name")
+    entity_mapping: Optional[Dict[str, str]] = Field(None, description="Entity mapping")
+
+
```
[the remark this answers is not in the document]
yuriihavrylko: Added `model_dump` override on `GLiNERRecognizerConfig` with `exclude_none=True` as default, so omitted yaml fields don't override constructor defaults like `model_name="urchade/gliner_multi_pii-v1"` with explicit `None`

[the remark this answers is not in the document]
yuriihavrylko: Agreed. Added a `model_validator` to reject configs that set both `entity_mapping` and `supported_entities`, surfacing the error at  config parse time with a clear message rather than at recognizer instantiation

presidio-analyzer/presidio_analyzer/input_validation/yaml_recognizer_models.py
```diff
@@ -176,6 +176,44 @@ class HuggingFaceRecognizerConfig(PredefinedRecognizerConfig):
     )
 
 
+class GLiNERRecognizerConfig(PredefinedRecognizerConfig):
+    """Configuration specifically for GLiNER models."""
+
+    model_config = ConfigDict(extra="forbid")
```
omri374: why are we setting extra="forbid"? wouldn't this prevent users from adding new kwargs to be passed to gliner? should it be extra="allow"?

presidio-analyzer/presidio_analyzer/input_validation/yaml_recognizer_models.py
```diff
@@ -176,6 +176,44 @@ class HuggingFaceRecognizerConfig(PredefinedRecognizerConfig):
...
+        return self
+
+    def model_dump(self, *args, **kwargs) -> Dict[str, Any]:
+        """Serialize the config without None values by default.
+
+        GLiNER recognizer kwargs are passed directly to the recognizer constructor.
+        Excluding None values preserves constructor defaults for omitted YAML fields
+        instead of overriding them with explicit None.
+        """
+        kwargs.setdefault("exclude_none", True)
```
omri374, line 223: That's a good catch. Could we add this to the HuggingFace recognizer config too? to make them aligned and correct?

presidio-analyzer/presidio_analyzer/input_validation/yaml_recognizer_models.py
yuriihavrylko: You're right. I defined all params explicitly available in gliner as of now, but `extra="forbid"` also blocks the existing `**model_kwargs` that get forwarded to `GLiNER.from_pretrained()`. Changed to `extra="allow"` to support both. Applied the same to HuggingFace recognizer for consistency

presidio-analyzer/presidio_analyzer/input_validation/yaml_recognizer_models.py
yuriihavrylko: Done, applied the same fix to `HuggingFaceRecognizerConfig` as well
But there's a small duplication between `HuggingFaceRecognizer` and `GLiNERRecognizer` - strong signal to move it into `PredefinedRecognizer`. Not sure if future subclasses would always want `exclude_none=True` by default

omri374 (approved):

PR #2025, opened 2026-05-15
fix: Custom operator validate() must not invoke the lambda

### LINKED ISSUE #2024: Custom operator lambda called with dummy "PII" value during validation, causing side effects in stateful lambdas
In presidio-anonymizer/presidio_anonymizer/operators/custom.py, the validate() method calls the user-supplied lambda with the literal string "PII" to assert it returns a str. For stateless lambdas this is harmless, but for stateful lambdas — the most common real-world use case, where the lambda builds a token→original-value map for later de-anonymization — this call silently inserts a spurious entry into the map before any real anonymization has occurred.

Minimal Reproducible Example:

```from presidio_anonymizer import AnonymizerEngine
from presidio_anonymizer.entities import OperatorConfig

analyzer = AnalyzerEngine()
engine = AnonymizerEngine()

token_map = {}
counter = {"n": 0}

def build_map(original_value: str) -> str:
    counter["n"] += 1
    token = f"PERSON_{counter['n']}"
    token_map[token] = original_value
    return f"[{token}]"

text = "Alice met Bob at the office."
results = analyzer.analyze(text=text, language="en", entities=["PERSON"])

engine.anonymize(
    text=text,
    analyzer_results=results,
    operators={"PERSON": OperatorConfig("custom", {"lambda": build_map})},
)

print(token_map)
```

Actual: {'PERSON_1': 'PII', 'PERSON_2': 'Alice', 'PERSON_3': 'Bob'}
Expected: {'PERSON_1': 'Alice', 'PERSON_2': 'Bob'}

The "PII" entry appears because validate() calls new_val("PII") at setup time. It also skews the counter — all real tokens start at _2 instead of _1.

Root Cause:
custom.py, validate() ~line 27:
```
result = new_val("PII")  # ← triggers side effects in stateful lambdas
if not isinstance(result, str):
    raise InvalidParamException(...)
```

Suggested Fix:
Remove the probe call entirely. callable() is sufficient — return-type is already enforced in operate() when the lambda runs on real data.

HammadSiddiqui:
## Change Description
`Custom.validate()` was calling `new_val("PII")` to check the return type of the lambda. This causes a spurious invocation with a dummy string, which breaks stateful lambdas — e.g. anonymizers that build a token-to-original-value map for de-anonymization will insert a bogus `{"TOKEN_1": "PII"}` entry and shift all subsequent token counters by one.

**Fix:**
- Remove the `new_val("PII")` probe call from `validate()` — only check `callable(new_val)`
- Move the return-type contract (`isinstance(result, str)`) to `operate()`, where the lambda runs on real data
**Tests added:**
- `test_given_non_str_lambda_then_ipe_raised_at_operate_time` — updated to reflect that the check now happens at operate-time
- `test_stateful_lambda_not_called_during_validate` — regression: asserts validate() never calls the lambda
- `test_stateful_token_map_not_corrupted_by_validate` — regression: asserts no spurious "PII" entry in token map after validate()
- 
## Issue reference

Fixes #2024

## Checklist


HammadSiddiqui:

omri374 (dismissed):

omri374:
Thanks, good change. Note that you don't have to use the custom operator for non-traditional operations. You can create your own operator. see more here: https://microsoft.github.io/presidio/samples/python/pseudonymization/

HammadSiddiqui:
Updated the Error assertion based on the copilot's review.

omri374 (approved):

PR #2033, opened 2026-05-19
feat(analyzer): add memory_zone support to SpacyNlpEngine for long-running services

Idan-Or:
## Summary

When Presidio runs as a long-lived service processing diverse text (e.g., PII masking in a web service), spaCy's `Vocab`/`StringStore` grows unboundedly — every unique token encountered becomes a permanent lexeme. In production this causes memory to grow from ~5 GiB to 11+ GiB over time until the process is OOMKilled.

spaCy 3.7+ introduced [`nlp.memory_zone()`](https://spacy.io/api/language#memory_zone) which allocates lexemes in a temporary pool that gets freed when the context manager exits, with zero performance impact.

This PR adds an opt-in `use_memory_zone` parameter to `SpacyNlpEngine` that wraps NLP processing in `memory_zone`, preventing unbounded vocab growth.

## Changes

- Added `use_memory_zone: bool = False` parameter to `SpacyNlpEngine.__init__()`
- When enabled, `process_text()` and `process_batch()` are wrapped in `nlp.memory_zone()`
- Graceful fallback via `hasattr(nlp, 'memory_zone')` — no-op on spaCy < 3.7
- 7 unit tests covering default value, enable/disable, invocation, and fallback

## Design Decisions

- **Defaults to `False`** — opt-in to avoid changing behavior for existing consumers. Users running Presidio as a long-lived service can enable it explicitly:
  ```python
  engine = SpacyNlpEngine(use_memory_zone=True)
  ```
- **Zero performance impact** — `memory_zone` uses the same Cython `Pool` allocator, just in a temporary scope
- **Backwards compatible** — gracefully falls back to `contextlib.nullcontext()` when spaCy < 3.7 or the attribute is missing

## Motivation

We run Presidio as a Kubernetes service processing PII masking requests across 4 languages (en, fr, pt, es with `_lg` models). Over ~3 hours of diverse traffic, per-pod memory grows from 5 GiB to 11+ GiB, triggering OOMKills. The root cause is spaCy's vocab permanently storing every unique token (names, emails, addresses) it encounters. Wrapping in `memory_zone` keeps memory flat at baseline.


Idan-Or:
@microsoft-github-policy-service agree company="monday.com"

PR #2044, opened 2026-05-29
fix(analyzer): prevent UrlRecognizer matching partial TLD labels (#1498)

### LINKED ISSUE #1498: UrlRecognizer causes false positives when analyzing code snippets
**Describe the bug**
The `BASE_URL_REGEX` pattern of `presidio_analyzer.predefined_recognizers.url_recognizer.UrlRecognizer` generates several false positives when analyzing text that contains code snippets, filenames, or any non-URL token in the format `<first>.<second>`. In particular, regex patterns like `(?:sy)`, `(?:mt)`, `(?:py)` etc. match with `os.system` (Python module), `zeus.mtia.local` (hostname), `rpc.py` (Python file) and others, causing unintended false positives like `os.sy`, `zeus.mt`, `rpc.py`, etc.

**To Reproduce**
Install `presidio_analyzer` and run this:
```

text = '''
# Exploit Title: - rpc.py: Remote Code Execution (RCE)
# Google Dork: N/A
# Date: 2022-07-12
# Exploit Author: Elias Hohl
# Vendor Homepage: https://github.com/abersheeran
# Software Link: https://github.com/abersheeran/rpc.py
# Version: v0.4.2 - v0.6.0
# Tested on: Debian 11, Ubuntu 20.04
# CVE : CVE-2022-35411
import requests
import pickle
HOST = "zeus.mtia.local:65432"
HEADERS = {
"serializer": "pickle"
}
def generate_payload(cmd):
  class PickleRce(object):
    def __reduce__(self):
      import os
      return os.system, (cmd,)
  payload = pickle.dumps(PickleRce())
  print(payload)
  return payload
'''

analyzer = AnalyzerEngine()
results = analyzer.analyze(text=text, entities=['URL'], language='en')
piis = []
for result in results:
    pii = text[result.start:result.end]
    piis.append(pii)

print(piis)
```
**Expected behavior**
`['https://github.com/abersheeran/rpc.py', 'https://github.com/abersheeran']`

**Actual behavior**
`['https://github.com/abersheeran/rpc.py', 'https://github.com/abersheeran', 'rpc.py', 'zeus.mt', 'os.sy']`

**Screenshots**
`UrlRecognizer.BASE_URL_REGEX`:
Image

**Additional context**
If this behavior is intended, adding `(?![a-z])` to the expression at least prevents false positives like `os.system` (see the screenshots below). Cases like `rpc.py` (Python file) and _rpc.py_ (valid Paraguayan URL) don't seem to have a good solution though.

Image
Image

Lawson-Darrow:
### Problem
`UrlRecognizer.BASE_URL_REGEX` matches a TLD even when it is only the *prefix* of a longer label. The TLD alternation has no trailing boundary, so schema-less `<word>.<word>` tokens — common in source code, module paths, and filenames — produce false positives:

| Input | Matched (before) |
|-------|------------------|
| `os.system` | `os.sy` (`.sy` = Syria) |
| `zeus.mtia.local` | `zeus.mt` (`.mt` = Malta) |

### Fix
Add a negative lookahead `(?![a-z0-9-])` immediately after the TLD alternation, requiring the matched TLD to be a complete domain label (i.e. not immediately followed by another label character).

- Genuine bare domains (`example.sy`, `microsoft.com`, `google.co.il`) still match.
- URLs with schemes/paths are unaffected — a path begins with `/`, which the lookahead permits.

### Tests
Added regression cases to `test_url_recognizer.py`:
- The reported false positives (`os.system`, `zeus.mtia.local`, and a code-snippet line) now yield zero matches.
- A guard case (`example.sy`) ensures a real ccTLD used as a complete label still matches.

All cases in `test_url_recognizer.py` pass; `ruff check` reports no new findings on the changed files. CHANGELOG updated under Analyzer → Fixed.

### Note on scope
Tokens that are *syntactically* valid bare domains remain ambiguous by design — e.g. `rpc.py` (a Python file) is indistinguishable from the Paraguay domain `rpc.py` without surrounding context, so it intentionally still matches. This PR removes only the unambiguous partial-label matches.


Lawson-Darrow:

PR #2046, opened 2026-05-29
feat(analyzer): merge adjacent same-type entities separated by whitespace

### LINKED ISSUE #1090: Merge two entities from the same type with whitespace between them
Currently, Presidio does not merge adjacent results having the same entity type. This causes the output to have multiple placeholders when replacing. 

For example: `My name is Dave Jones` could with some NER models produce `My name is <PERSON> <PERSON>`.
It is better if the output would be `My name is <PERSON>`, especially when replacing tokens with fake values or using OpenAI for fake data generation.

AlexanderSanin:
## Summary

- Fixes the long-standing issue where NER models that tokenise multi-word entities (e.g. spaCy detecting "Dave" and "Jones" as two separate `PERSON` spans) would produce redundant placeholders like `<PERSON> <PERSON>` instead of a single `<PERSON>`.
- Adds `EntityRecognizer.merge_adjacent_text_entities(results, text)` — a static method that sorts results by start offset and greedily fuses consecutive spans of the **same entity type** when the text between them is purely whitespace. The merged span takes `max(score_a, score_b)`.
- Wires the new method into `AnalyzerEngine.analyze()` immediately after `remove_duplicates()`, so it is applied transparently to every analysis call.

## Test plan

- [x] `test_merge_adjacent_same_type_entities` — two PERSON spans `"Dave" + "Jones"` are merged into one
- [x] `test_merge_adjacent_preserves_max_score` — merged entity uses the higher of the two scores
- [x] `test_merge_adjacent_three_tokens` — chain of three same-type spans (`"Jean Luc Picard"`) collapses to one
- [x] `test_no_merge_when_different_entity_types` — PERSON + LOCATION are not merged
- [x] `test_no_merge_when_gap_has_non_whitespace` — spans separated by punctuation are not merged
- [x] `test_merge_empty_results` — empty input returns empty output without error
- [x] All pre-existing tests continue to pass (13/13 green)

Closes #1090

### DIFF
--- a/presidio-analyzer/presidio_analyzer/analyzer_engine.py
+++ b/presidio-analyzer/presidio_analyzer/analyzer_engine.py
@@ -256,6 +256,7 @@ class AnalyzerEngine:
 
         # Remove duplicates or low score results
         results = EntityRecognizer.remove_duplicates(results)
+        results = EntityRecognizer.merge_adjacent_text_entities(results, text)
         results = self.__remove_low_scores(results, score_threshold)
 
         if allow_list:
--- a/presidio-analyzer/presidio_analyzer/entity_recognizer.py
+++ b/presidio-analyzer/presidio_analyzer/entity_recognizer.py
@@ -290,6 +290,48 @@ class EntityRecognizer:
         return filtered_results
 
     @staticmethod
+    def merge_adjacent_text_entities(
+        results: List[RecognizerResult], text: str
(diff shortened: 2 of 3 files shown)


AlexanderSanin:
Hey @Surya-5555 @omri374 @SharonHart. Could you, please, have a look at this?

PR #2061, opened 2026-06-09
fix(analyzer): ignore empty allow_list terms in regex match

uwezkhan:
`_remove_allow_list` builds the regex-mode allow list by joining the terms with `|` and calling `search` on each detected span. A single empty string in the list makes that an empty alternation, which matches every span, so all detected entities get treated as allowed. Before: `allow_list=["foo", ""]` with `allow_list_match="regex"` returns zero results and every entity passes through as not-PII, so a blank entry coming out of upstream string handling flips the filter into allow-all.

After: empty terms are dropped before the pattern is compiled, and if nothing is left the results are returned unchanged. Keeping the guard in the callee means the allow list stays fail-closed regardless of how the caller assembled it, instead of each caller having to strip blanks first. The tradeoff is that a literal empty regex term can no longer match everything, which is not a meaningful allow-list entry.


SharonHart:
@uwezkhan can you take a look at the test coverage? I think that there is a missing test for a fully empty string list case

uwezkhan:
Good catch. The existing test passed `["unrelated", ""]`, which filters down to `["unrelated"]` and never reaches the early-return guard, so the all-empty path was uncovered. Added `test_when_regex_allow_list_is_all_empty_entries_then_results_are_kept` with `allow_list=["", ""]` to hit that branch directly. Both results are kept and all 10 `-k allow_list` tests pass.

SharonHart (approved):

Issue #1476, opened 2024-10-28
IP Recognizer has bugs when ipv6 contains double colon `::`


yungyuchen521:
# Describe the bug
The [regex](https://github.com/microsoft/presidio/blob/main/presidio-analyzer/presidio_analyzer/predefined_recognizers/ip_recognizer.py#L25) doesn't work correctly on ipv6 with double colons `::`

| Case | Example | Matches | Expected |
| - | - | - | - |
| Valid address with `::` | A099::09C0:876A:130B | ["A099::"] | ["A099::09C0:876A:130B"] |
| Invalid address with multiple `::` | 2001::25de::cade | ["2001::", "25de::"] | [] |

# To Reproduce
```python
import re

PATTERN = re.compile(r"\b(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))\b")

VALID_IP_WITH_DOUBLE_COLON = "A099::09C0:876A:130B"
[...]

PR #2062, opened 2026-06-11
test(ip-recognizer): add regression tests for IPv6 double-colon bugs (#1476)

AlexanderSanin:
## Summary

Issue #1476 reported two bugs in the `IpRecognizer` IPv6 pattern:

1. **Valid address with single `::`** — `A099::09C0:876A:130B` was matched as `A099::` instead of the full address (the `{1,7}::` alternative fired before the `{1}::{1,6}` one).
2. **Invalid address with multiple `::`** — `2001::25de::cade` produced two spurious matches (`2001::` and `25de::`) instead of none, because `\b` treats `:` as a non-word character.

Both bugs have since been fixed in the regex (the lookbehind/lookahead guards were updated), but the specific test cases from the issue were never added to the test suite, leaving the door open for silent regressions.

This PR adds the two exact examples from the issue report as explicit parametrized test cases:

- `A099::09C0:876A:130B` → 1 match at `(0, 20)` in `test_when_ipv6_compression_then_succeed`
- `2001::25de::cade` → 0 matches in `test_when_invalid_ip_then_no_match`

## Test plan

- [x] Run `pytest presidio-analyzer/tests/test_ip_recognizer.py` — all **115 tests pass** (113 pre-existing + 2 new)
- [x] New tests specifically exercise the two failure modes described in #1476
- [x] No production code changed; change is test-only

### DIFF
--- a/presidio-analyzer/tests/test_ip_recognizer.py
+++ b/presidio-analyzer/tests/test_ip_recognizer.py
@@ -81,6 +81,8 @@ def test_when_all_ips_then_succeed(
         ("2001:db8:85a3::8a2e:370", 1, ((0, 23),), ((0.6, 0.81),),),
         ("2001:db8::1", 1, ((0, 11),), ((0.6, 0.81),),),
         ("fe80::1%eth0", 1, ((0, 12),), ((0.6, 0.81),),),
+        # Regression for issue #1476: 1 group before :: must match fully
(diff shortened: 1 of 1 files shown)


AlexanderSanin:
Hey @SharonHart @Surya-5555. Could you, please, have a look at this?

SharonHart (approved):

PR #2063, opened 2026-06-13
fix(analyzer): cap url recognizer host to stop quadratic backtracking

uwezkhan:
## Change Description

The URL recognizer matches the host with `[a-z0-9.\-]+[.]`, where the dot sits both inside the repeated character class and as the required separator right after it. The non-schema pattern runs against every analyzed string, so a short run of dots backtracks against the large TLD alternation and the match time grows with the square of the input length. This is a regular-expression denial of service on a path that runs before any entity-specific logic, bounded today only by `REGEX_TIMEOUT_SECONDS`.

Before, the host run was unbounded and the cost was quadratic, so a few kilobytes of crafted text consumed the whole regex timeout budget per request. After, the run is bounded to the 253 character DNS hostname limit and the cost is linear in input length, while every valid URL still matches at the same span. The bound lives in the shared `BASE_URL_REGEX`, so all four URL patterns inherit it. The tradeoff is a hard cap on host length, which only excludes inputs that are already invalid as hostnames.

Benchmark (non-schema pattern over `"." * n`, `regex` engine):

```
n       before    after
1000    ~1.9s     ~0.9s
2000    ~7.8s     ~1.9s
4000    ~31s      ~3.9s
```

## Issue reference

N/A

## Checklist


uwezkhan:

SharonHart (approved):

Issue #2012, opened 2026-04-30
Question about `HuggingFaceNerRecognizer`


tcsmaster:
Hey!

I am really glad that the `HuggingFaceNerRecognizer` came to be, becasue its fits neatly in my use case.  However, I have a remark, and a question about it.

In the docstring, the example is creating an empty `AnalyzerEngine` and adds this recognizer into its registry. However, this instantiates an `NlpEngine`, which still tries to download spacy models, which is why this recognizer was built in the first place. Nonetheless, I tried the recognizer on it's own with my own ner mdoel, and it worked very well.

My question is this (becuase I haven't been using Presidio for that long): is there a drawback to using the recognizer on its own, without an engine? What do I lose not using an `AnalyzerEngine`?

Thanks

SharonHart:
Thanks for raising this @tcsmaster!
You’re right. If you use only HuggingFaceNerRecognizer, you don’t strictly need spaCy, or even the slim NLP engine (downloads small spacy model), for the recognition itself. This recognizer runs the Hugging Face pipeline directly and does not rely on spaCy token alignment or nlp_artifacts.

The reason AnalyzerEngine still initializes an NlpEngine is that the analyzer is a general orchestration layer. Other Presidio recognizers and features rely on NLP artifacts such as tokens, lemmas, stop-word/punctuation information, NLP entities, and context words. These are used for things like context-aware score enhancement, built-in NLP recognizers, and multi-recognizer flows.

So the tradeoff is:

* Using HuggingFaceNerRecognizer directly is fine if you only need that model’s results.
* Using it through AnalyzerEngine gives you Presidio features such as registry orchestration, multiple recognizers, allow-list filtering, score thresholding, duplicate removal, explanations, and batch/dict integrations.

For the AnalyzerEngine case, the current recommendation would be to use the slim NLP engine, since it avoids spaCy NER and only provides lightweight linguistic artifacts. That said, for self-contained recognizers like HuggingFaceNerRecognizer, a no-op/empty NLP engine would also make sense. We should consider adding one, so users can keep AnalyzerEngine orchestration without loading any NLP model when their recognizers don’t need NLP artifacts.

cc: @omri374

PR #2071, opened 2026-06-17
feat: add no-op NLP engine

ultramancode:
## Change Description

Adds a `NoOpNlpEngine` for analyzer configurations where the active recognizers do not need artifacts produced by an NLP engine.

This is mainly useful for recognizers such as `HuggingFaceNerRecognizer`, which runs model inference directly and does not need NLP artifacts from spaCy. In that setup, loading a spaCy model only adds startup cost.

With this change, users can select `nlp_engine_name: no_op` in analyzer configuration instead of configuring a real NLP model that will not be used.

Includes regression tests and an updated `HuggingFaceNerRecognizer` example.

## Issue reference

Refs #2012

## Checklist


presidio-analyzer/presidio_analyzer/nlp_engine/no_op_nlp_engine.py
```diff
@@ -0,0 +1,166 @@
...
+    def _create_empty_nlp_artifacts(self, language: str) -> NlpArtifacts:
+        return NlpArtifacts(
+            entities=[],
+            tokens=Doc(self._vocab, words=[]),
+            tokens_indices=[],
+            lemmas=[],
+            nlp_engine=self,
+            language=language,
+            scores=[],
+        )
```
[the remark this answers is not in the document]
ultramancode: `NlpArtifacts` does not accept a `keywords` constructor argument. Keywords are derived internally from `lemmas` in `NlpArtifacts.__init__`, so passing `lemmas=[]` keeps `artifacts.keywords` as an empty list.

omri374:
Thanks! Have you seen the slim nlp engine? https://github.com/microsoft/presidio/blob/main/presidio-analyzer/presidio_analyzer/nlp_engine/slim_spacy_nlp_engine.py

Can we use this as is / adapt it?

ultramancode:
Thanks for the pointer! I looked at `SlimSpacyNlpEngine`, and I agree it is relevant here.

The main difference is that `slim` is still spaCy-backed: it disables NER, but still initializes a spaCy pipeline for token/lemma artifacts. The use case here is a bit stricter: recognizers that do not need those artifacts at all, so we can skip initializing a spaCy model or blank pipeline.

We could technically adapt `slim` with a no-op mode, but that would broaden its contract: depending on configuration, `slim` would either provide token/lemma artifacts or return empty artifacts.

I went with a separate `no_op` engine to make the path clearer for cases where the recognizers themselves don’t need NLP artifacts, but I can adapt the PR if you’d prefer to keep that under `slim`.

presidio-analyzer/tests/test_no_op_nlp_engine.py
```diff
@@ -0,0 +1,389 @@
...
+        results = no_op_nlp_engine.process_batch(["one", "two"], language="en")
+
+        assert isinstance(results, Iterator)
+        result_list = list(results)
+        assert [text for text, _ in result_list] == ["one", "two"]
+        assert all(artifacts.entities == [] for _, artifacts in result_list)
+        assert all(isinstance(artifacts.tokens, Doc) for _, artifacts in result_list)
+        assert all(len(artifacts.tokens) == 0 for _, artifacts in result_list)
+
+    def test_when_process_batch_as_tuples_then_context_preserved(
```
omri374, line 158: This test reads as "context awareness works with the no-op engine," but it only exercises the explicitly-passed context=["zip"] path. Because `NoOpNlpEngine` returns empty tokens, `LemmaContextAwareEnhancer._extract_surrounding_words` short-circuits to [""], so context words that appear in the analyzed text (the far more common case, e.g. analyze("my zip code is 10023", ...) with no context= arg) will not boost the score.

That's a legitimate limitation of a truly no-op engine, but right now the test obscures it. Could we either:

1. Rename this to something like ..._context_enhancer_handles_explicit_context, and
add a companion test asserting that an in-text context word does not change the score (locking in the known behavior so it's documented, not accidental)?
2. A one-line note in the `NoOpNlpEngine` docstring ("context enhancement only applies to context passed explicitly via context=, not words found in the text, since no tokens are produced") would also save users a confusing debugging session.
3. If the user passes both the `NoOpNlpEngine`, and the `LemmaContextAwareEnhancer`, a warning should be raised that the context enhancer will not work without proper tokenization and/or lemmatization.

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry.py
```diff
@@ -76,6 +77,10 @@ def add_nlp_recognizer(self, nlp_engine: NlpEngine) -> None:
         :return: None
         """
 
+        if isinstance(nlp_engine, NoOpNlpEngine):
```
omri374, line 98: The guard here silently skips for a no-op engine, but the parallel guard in `recognizer_registry_provider.py::__update_based_on_nlp_recognizer_conf` raises when a `SpacyRecognizer` is present with a no-op engine.

So the two construction paths disagree: `AnalyzerEngineProvider(...)` with a no-op engine + a `SpacyRecognizer` raises a helpful error, but `AnalyzerEngine(nlp_engine=NoOpNlpEngine(...))` with a manually `registry.add_recognizer(SpacyRecognizer())` silently registers a recognizer that then runs against empty artifacts and finds nothing.

Could we make these consistent? Either detect an explicitly-added NLP recognizer in the direct path and warn (or raise) the same way, or at least add a comment noting the divergence is intentional. As is, the direct path is a silent-failure trap.

presidio-analyzer/presidio_analyzer/nlp_engine/no_op_nlp_engine.py
```diff
@@ -0,0 +1,165 @@
...
+        texts: Iterable[Union[str, Tuple[str, Any]]],
+        language: str,
+        batch_size: int = 1,
+        n_process: int = 1,
+        as_tuples: bool = False,
+        **kwargs,
+    ) -> Iterator[Union[Tuple[str, NlpArtifacts], Tuple[str, NlpArtifacts, Any]]]:
+        """Return empty NLP artifacts for each text in a batch."""
+        if kwargs:
+            raise ValueError(
```
omri374: I'd consider changing this to a warning. The user might not be the one triggering this call, and replacing a flow that uses a spaCy engine with this would result in a value error without the user being able to fix it easily.

omri374 (commented):
Thanks! Apologies for the delay, left a few comments but it looks good!

presidio-analyzer/tests/test_no_op_nlp_engine.py
ultramancode: I renamed the context-enhancement test to clarify that it covers explicitly provided context, added a companion test for context words in the analyzed text, and documented the limitation.

I also added a warning when a user passes `LemmaContextAwareEnhancer` with `NoOpNlpEngine`.

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry.py
ultramancode: I moved the existing compatibility check into the registry and now run it for both construction paths.
Both paths raise the same `ValueError` when an incompatible NLP recognizer is registered.

presidio-analyzer/presidio_analyzer/nlp_engine/no_op_nlp_engine.py
ultramancode: Changed this to log a warning and continue processing instead of raising a `ValueError`.

ultramancode:
Thanks for taking the time to review this and leave detailed comments!
I've pushed the updates and replied to each one.

omri374 (approved):

PR #2074, opened 2026-06-18
fix(analyzer): stop UsSsnRecognizer from over-blocking 987-65-432X SSNs

AUTHENSOR:
## Change Description

`UsSsnRecognizer.invalidate_result()` drops detected SSN candidates that match a hardcoded "sample SSN" blocklist. The check uses `str.startswith()`:

```python
for sample_ssn in ("000", "666", "123456789", "98765432", "078051120"):
    if only_digits.startswith(sample_ssn):
        return True
```

The mixed list conflates two different intents — **never-issued area numbers** (`000`, `666`, which are legitimately a 3-digit *prefix*/area check) and **full 9-digit canonical sample SSNs** — and matches both with a prefix test. The defect is that one of the 9-digit literals, `"98765432"`, is **only 8 digits**: an apparent truncation of the canonical fake `"987654320"`. Because the test is `startswith`, that truncated literal over-blocks the **entire `987-65-4320` … `987-65-4329` family — ten distinct SSN-shaped values** — instead of the single intended sample.

The practical impact is a **false negative that leaks PII**: real SSN-shaped strings such as `987-65-4321` (a canonical example printed on countless real-world forms) are silently dropped to **no result** and pass through **unredacted**. This is a confidentiality failure, not a cosmetic one: the recognizer's whole job is to catch SSNs so a downstream anonymizer can redact them, and here it actively suppresses real ones.

### Fix

Split the two concerns that the single `startswith` loop had merged:

- **Area-number check (kept as a prefix/area check):** an SSN whose **area number** (first group) is `000` or `666` is invalid — the SSA never issues those. This stays a first-three-digits check: `only_digits[:3] in ("000", "666")`.
- **Sample-SSN check (now exact equality):** the full 9-digit canonical samples are matched by exact equality instead of prefix: `only_digits in ("123456789", "987654320", "078051120")`. The truncated `"98765432"` is corrected to its intended 9-digit canonical form `"987654320"`, so it now invalidates only itself.

The regex and scores are unchanged. SSNs the recognizer is *meant* to drop are still dropped: `000`/`666`-area SSNs, the all-same-digit case, the all-zero-group case, and the three canonical sample SSNs (`987-65-4320`, `078-05-1120`, `123-45-6789`) all remain invalidated.

### Verification

A standalone repro instantiates the real `UsSsnRecognizer` and runs `.analyze(..., ["US_SSN"], nlp_artifacts=None)`.

Before the fix — the whole family leaks:

```
987-65-4320 .. 987-65-4329: LEAKED (no result)   # all ten
219-09-9999 (control):       DETECTED
```

After the fix:

```
987-65-4321 .. 987-65-4329:  DETECTED            # nine real SSNs recovered
987-65-4320:                 invalidated         # the actual canonical sample, now by exact match
078-05-1120 / 123-45-6789:   invalidated         # other canonical samples, still dropped
000-12-3456 / 666-12-3456:   invalidated         # never-issued area numbers, still dropped
219-09-9999 (control):       DETECTED
```

`ruff check` / `ruff format --check` on the changed source and the targeted SSN test file both pass:

```
$ ruff check --force-exclude .../us_ssn_recognizer.py
All checks passed!
$ ruff format --check --force-exclude .../us_ssn_recognizer.py
1 file already formatted

$ PYTHONPATH=presidio-analyzer python -m pytest presidio-analyzer/tests/test_us_ssn_recognizer.py -q
............................                                             [100%]
28 passed in 0.03s
```

The test file gains: `987-65-4321` … `987-65-4329` detected; `987-65-4320` / `078-05-1120` / `123-45-6789` still invalidated; `000`/`666`-area SSNs still invalidated; and a plain valid SSN (`219-09-9999`) still detected.

**Downstream blast radius:** NVIDIA NeMo Guardrails' PII rail inherits this exact Presidio recognizer, so the same SSN leak reached that deployed guardrail too — real SSNs in the `987-65-432X` family flowed through a configured PII guardrail unredacted.

## Note on CHANGELOG

CHANGELOG entry omitted from this patch to avoid merge conflicts with sibling PRs that all insert at the same `#### Fixed` anchor. Happy to add the entry once the merge order is known, or maintainers can squash it in.

## Checklist


AUTHENSOR:

SharonHart (approved):

SharonHart (approved):

PR #2075, opened 2026-06-18
fix(analyzer): detect Mastercard 2-series and 18-19 digit credit cards

AUTHENSOR:
## Change Description

`CreditCardRecognizer` was missing two whole families of real, current, Luhn-valid cards. Because the PAN regex never matched them, the recognizer returned **no** `CREDIT_CARD` result, so these card numbers passed Presidio as "clean" and would leak unredacted downstream (analyze → anonymize):

1. **Mastercard 2-series** — BIN range **2221-2720** (the first four digits), issued since 2017 (billions of live cards). The regex only had branches for leading digits 4/5/6/1/3, so e.g. `2221000000000009` (Luhn-valid) returned
   `[]`.
2. **18-19 digit PANs** — ISO/IEC 7812 allows PANs up to 19 digits
   (UnionPay, Maestro, some Visa). The old length window
   (`\d{3,4} \d{3,4} \d{3,5}` after a 4-digit prefix) capped matches at 17
   digits, so a Luhn-valid 19-digit card returned `[]`.

A leading-4 Visa such as `4111111111111111` was, and still is, detected at
score `1.0`, so this is purely a **false-negative** fix on the leak surface.

### Fix

Widen the PAN regex in
`presidio_analyzer/predefined_recognizers/generic/credit_card_recognizer.py`:

- Add a Mastercard 2-series prefix branch,
  `2(22[1-9]|2[3-9]\d|[3-6]\d\d|7[01]\d|720)`, which matches exactly the
  four-digit values 2221-2720 (same 4-digit shape as the existing
  `4\d{3}` / `5[0-5]\d{2}` / `6\d{3}` branches).
- Raise the final group from `\d{3,5}` to `\d{3,7}` so the pattern spans
  13-19 digit PANs.

The change is conservative: **Luhn (`validate_result`) still gates every
match**, and the existing negative lookahead that rejects 13-digit Unix
timestamps starting with `1` (`(?!1\d{12}(?!\d))`) is untouched. Widening the
range therefore does not flag non-card numbers — a 19-digit run that is not
Luhn-valid is still rejected, exactly as today.

### Verification

Targeted tests (run from the `presidio-analyzer` source dir):

```
$ python -m pytest tests/test_credit_card_recognizer.py -q
...........................                                              [100%]
27 passed in 0.03s
```

Lint on the changed source file:

```
$ ruff check  .../generic/credit_card_recognizer.py   -> All checks passed!
$ ruff format --check .../generic/credit_card_recognizer.py -> 1 file already formatted
```

New parametrized cases assert:
- Luhn-valid Mastercard 2-series PANs (`2221000000000009`, `2720990000000007`,
  `2223000048400011`) are detected as `CREDIT_CARD` at score `1.0`, including
  with context (`my credit card: ...`).
- Luhn-valid 18- and 19-digit PANs (`675919345145061238`,
  `4109906958483040118`, `6298036494205552661`) are detected.
- Regression guards: every previously detected card is still detected, the
  Unix-timestamp case (`1748503543012`) is still **not** flagged, the
  Luhn-invalid Visa/Discover cases stay rejected, and a new Luhn-**invalid**
  2-series PAN (`2221000000000001`) is rejected.

## Checklist


SharonHart:
@AUTHENSOR can you please take a look at the comments?

PR #2086, opened 2026-06-20
feat(analyzer): Add ONNX Runtime backend to HuggingFaceNerRecognizer

yuriihavrylko:
## Change Description

Adds an optional ONNX Runtime inference backend to `HuggingFaceNerRecognizer`, selected via a new `backend` parameter (`"torch"` - default, unchanged - or `"ort"`). The ort backend loads token-classification models through `optimum.onnxruntime`, enabling:

- Pre-quantized ONNX variants (FP16/INT8/4-bit) published on the HF Hub
- Hardware selection via ONNX Runtime execution providers (CUDA, TensorRT, OpenVINO, CoreML, ROCm) without code changes

### How

- **`backend` parameter** on `HuggingFaceNerRecognizer`. For `"ort"`, the recognizer pre-loads `ORTModelForTokenClassification.from_pretrained()` explicitly and hands the model object to the pipeline. This scopes loader kwargs (`subfolder`, `file_name`, `provider`, …) to the model loader only - required for mixed-layout repos (`onnx-community/*`, etc) that keep ONNX files under `onnx/` while config/tokenizer live at the repo root. Passing these at the pipeline level breaks on such repos.
- **`**model_kwargs` pass-through** (mirrors `GLiNERRecognizer`): extra kwargs — from Python or YAML — are forwarded to the active backend's loader. No recognizer changes needed for future loader options.
- **`onnxruntime` extra** in `pyproject.toml`: `pip install 'presidio-analyzer[onnxruntime]'` (CPU build). GPU/accelerator builds are installed directly instead of the extra (see docs) because the `onnxruntime*` packages ship the same Python module and conflict.

### Usage

```yaml
- name: "HF NER (ONNX)"
  type: predefined
  class_name: HuggingFaceNerRecognizer
  model_name: onnx-community/stanford-deidentifier-base-ONNX
  backend: ort
  subfolder: onnx
  file_name: model_quantized.onnx
  label_mapping:
    PATIENT: PERSON
    HCW: PERSON
    PHONE: PHONE_NUMBER
```

### Breaking change

Unknown constructor kwargs were previously logged and **dropped**; they are now forwarded to the model loader and raise `TypeError` if the loader rejects them. Typos in YAML recognizer configs that loaded silently before will now fail loudly at startup. This is intentional (silent misconfiguration of a PII detector is worse than a startup error), but existing configs with stray keys need cleanup.

## Issue reference

None

## Checklist


presidio-analyzer/tests/test_huggingface_ner_recognizer.py
```diff
@@ -472,14 +473,28 @@ def test_hf_recognizer_device_fallback_and_validation():
...
-    assert "Ignoring unsupported kwargs" in caplog.text
+        assert rec.model_kwargs == {
+            "revision": "main",
+            "cache_dir": "/tmp/cache",
+        }
+        mock_hf_pipeline.assert_called_once_with(
+            "token-classification",
+            model="test-model",
+            tokenizer="test-model",
+            aggregation_strategy="simple",
```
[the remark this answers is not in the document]
yuriihavrylko: These are false positives -  `EntityRecognizer.__init__` calls `self.load()`, so constructing the recognizer already triggers the pipeline/ORT model creation - the assertions run against that. 

Verified the test passes as written. Adding an explicit `rec.load()` would be a redundant no-op (load() early-returns when the pipeline is already built)

.github/workflows/ci.yml
```diff
@@ -56,6 +56,10 @@ jobs:
         with:
           fail-on-severity: low
           allow-licenses: MIT, Apache-2.0, BSD-3-Clause, 0BSD, 0BSD AND Apache-2.0 AND BSD-3-Clause AND MIT
+          # ml-dtypes (transitive via optimum-onnx -> onnx) is Apache-2.0 with
+          # MPL-2.0-licensed Eigen headers; exempt it rather than allowing
+          # MPL-2.0 globally.
+          allow-dependencies-licenses: pkg:pypi/ml-dtypes
```
omri374, line 62: Is the CI failing due to licensing issues?

omri374:
Looks great! Thanks and apologies for the delayed review. I left one minor comment.

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
```diff
@@ -137,30 +143,63 @@ def __init__(
...
-            raise ImportError(
-                "torch is not installed. Please install it "
-                "(pip install torch) to use this recognizer."
+                "(pip install transformers) to use this recognizer."
             )
+        if self.backend == "torch":
+            if torch is None:
+                raise ImportError(
+                    "torch is not installed. Please install it "
+                    "(pip install torch) to use the 'torch' backend."
```
omri374: why not refer the users to "pip install presidio-analyzer[transformers]"?

.github/workflows/ci.yml
yuriihavrylko: Yes. The `Dependency Review` job enforces the allow-licenses list, and the new `onnxruntime` extra pulls in `ml-dtypes` transitively (`optimum-onnx` -> `onnx` -> `ml-dtypes`), which is licensed MPL-2.0 **AND** Apache-2.0. The MPL-2.0 part isn't in the allow-list, so the job fails.

I scoped the exemption to just that one package rather than adding MPL-2.0 to allow-licenses globally, so the policy stays strict for everything else.

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
```diff
@@ -292,20 +360,50 @@ def load(self) -> None:
                 tokenizer=self.tokenizer_name,
                 aggregation_strategy=self.aggregation_strategy,
                 device=device,
+                model_kwargs=self.model_kwargs or None,
```
omri374: On the torch backend this spreads `model_kwargs` next to transformers' own hub kwargs. In transformers 4.57 `pipeline()` does `AutoConfig.from_pretrained(model, **hub_kwargs, **model_kwargs)` where `hub_kwargs` already contains `revision`, `token` and `trust_remote_code`. So `HuggingFaceNerRecognizer(model_name=..., revision="main")` (or `revision: main` in YAML), which the docstring and docs recommend, fails at load with "got multiple values for keyword argument 'revision'".

The unit test `test_hf_recognizer_forwards_extra_kwargs_as_model_kwargs` only asserts the mock received `{"revision": "main", ...}`, so it does not catch this.

Suggested fix: pop `revision` / `token` / `trust_remote_code` out of `model_kwargs` and pass them as top-level `pipeline()` arguments, or restrict the documented torch kwargs to ones that only `from_pretrained` accepts.

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
```diff
@@ -154,19 +161,53 @@ def __init__(
...
+              For NVIDIA GPU, install `onnxruntime-gpu` and pass
+              `provider="CUDAExecutionProvider"` via model_kwargs.
+        :param model_kwargs: Additional keyword arguments forwarded to the
+            underlying model loader.
+            For the "torch" backend, passed to `transformers.pipeline` as
+            `model_kwargs=`. For "ort", passed to
+            `ORTModel.from_pretrained` directly (so they scope to the model
+            loader only — important for mixed-layout repos where ONNX is
+            under `onnx/` but the tokenizer/config are at the repo root).
+            Use this for e.g. `file_name`, `subfolder`, `revision`,
```
omri374: This list mixes ort-only kwargs (`file_name`, `provider`, `provider_options`, `session_options`) with generic ones. On the default torch backend they reach `PreTrainedModel.from_pretrained` and raise `TypeError`, where before this PR they were logged and dropped. A user who copies the ort YAML sample but forgets `backend: ort` now gets an opaque transformers error with no hint that the key is ort-specific.

Please qualify each kwarg by backend here and in the docs page, and consider a clearer error (or at least a log line naming the backend) when a known ort-only key is passed with `backend="torch"`.

presidio-analyzer/tests/test_huggingface_ner_recognizer.py
```diff
@@ -602,6 +618,248 @@ def test_hf_recognizer_analyze_handles_malformed_pipeline_output(
...
+            "presidio_analyzer.predefined_recognizers.ner."
+            "huggingface_ner_recognizer.torch",
+            None,
+        ):
+            with pytest.raises(ImportError, match="torch is not installed"):
+                HuggingFaceNerRecognizer(model_name="test-model", backend="torch")
+
+
+@pytest.mark.usefixtures("mock_torch_installed")
+def test_hf_recognizer_ort_backend_no_torch_ok():
```
omri374: This test asserts something that is not true outside the mock. `optimum` 2.1.0 declares `torch` as a hard dependency and `optimum/pipelines/__init__.py` does `import torch` at module top, and `optimum-onnx` 0.1.0 `modeling.py` does the same. So the ort backend cannot run without torch; with torch missing, the module-level import sets `optimum_pipeline = None` and the user sees the misleading "optimum is not installed" error.

Suggest dropping this test, stating torch as a requirement of the ort backend in the docstring ("Requires: optimum, optimum-onnx[onnxruntime]" understates it), and mentioning torch in the ImportError message.

presidio-analyzer/presidio_analyzer/conf/hf_ner_onnx.yaml
```diff
@@ -0,0 +1,40 @@
...
+  recognizers:
+    - name: SpacyRecognizer
+      type: predefined
+      enabled: false
+
+    - name: "HuggingFace NER (ONNX)"
+      type: predefined
+      class_name: HuggingFaceNerRecognizer
+      model_name: onnx-community/stanford-deidentifier-base-ONNX
+      backend: ort
```
omri374: There is no test that loads `backend` / `subfolder` / `file_name` from YAML through `RecognizerRegistryProvider`; the ort YAML path is only exercised via direct constructor calls in the e2e file, and this shipped config is not loaded by any test. The YAML path currently works only because `HuggingFaceRecognizerConfig` is `extra="allow"` and `_prepare_recognizer_kwargs` forwards unknown keys; a change to either would silently drop `backend` and fall back to torch with nothing failing.

Please add a (mocked) config-path test: enable a HuggingFaceNerRecognizer entry with `backend: ort`, `subfolder`, `file_name` in a test YAML, load it through `RecognizerRegistryProvider`, and assert `rec.backend == "ort"` and `rec.model_kwargs == {"subfolder": ..., "file_name": ...}`. The existing `test_hf_recognizer_with_chunker.yaml` test in `test_recognizer_registry_provider.py` is a good template.

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
omri374, line 1: Thanks, the ort backend is a good addition and pre-loading the ORT model so `subfolder` / `file_name` stay scoped to the loader is the right call for mixed-layout repos.

One small request beyond the inline comments: please declare `backend` on `HuggingFaceRecognizerConfig` in `input_validation/yaml_recognizer_models.py`, e.g. `backend: Optional[Literal["torch", "ort"]] = Field(None, ...)`. Today it passes only via `extra="allow"`, so `backend: onnx` in YAML fails late with a bare `ValueError` from the constructor instead of a parse-time error, and the rule in `.github/instructions/yaml-config.instructions.md` is that constructor parameters and schema fields do not drift apart.

For awareness: we are reworking how recognizer YAML keys map to constructor kwargs before the next release, and library pass-through options will likely move into a nested `model_kwargs:` block rather than flat top-level keys. No need to change the shape in this PR; we will adapt it as part of that work.

If you would rather not pick this up again after the gap, say so and I will push the fixes onto the branch myself.

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
omri374, line 363: On the torch backend this spreads `model_kwargs` next to transformers' own hub kwargs. In transformers 4.57 `pipeline()` does `AutoConfig.from_pretrained(model, **hub_kwargs, **model_kwargs)` where `hub_kwargs` already contains `revision`, `token` and `trust_remote_code`. So `HuggingFaceNerRecognizer(model_name=..., revision="main")` (or `revision: main` in YAML), which the docstring and docs recommend, fails at load with "got multiple values for keyword argument 'revision'".

`test_hf_recognizer_forwards_extra_kwargs_as_model_kwargs` only asserts the mock received `{"revision": "main", ...}`, so it does not catch this.

Suggested fix: pop `revision` / `token` / `trust_remote_code` out of `model_kwargs` and pass them as top-level `pipeline()` arguments, or restrict the documented torch kwargs to ones that only `from_pretrained` accepts.

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
omri374, line 178: This list mixes ort-only kwargs (`file_name`, `provider`, `provider_options`, `session_options`) with generic ones. On the default torch backend they reach `PreTrainedModel.from_pretrained` and raise `TypeError`, where before this PR they were logged and dropped. A user who copies the ort YAML sample but forgets `backend: ort` now gets an opaque transformers error with no hint that the key is ort-specific.

Please qualify each kwarg by backend here and in the docs page, and consider a clearer error (or at least a log line naming the backend) when a known ort-only key is passed with `backend="torch"`.

presidio-analyzer/tests/test_huggingface_ner_recognizer.py
omri374, line 800: This test asserts something that is not true outside the mock. `optimum` 2.1.0 declares `torch` as a hard dependency and `optimum/pipelines/__init__.py` does `import torch` at module top, and `optimum-onnx` 0.1.0 `modeling.py` does the same. So the ort backend cannot run without torch; with torch missing, the module-level import sets `optimum_pipeline = None` and the user sees the misleading "optimum is not installed" error.

Suggest dropping this test, stating torch as a requirement of the ort backend in the docstring ("Requires: optimum, optimum-onnx[onnxruntime]" understates it), and mentioning torch in the ImportError message.

presidio-analyzer/presidio_analyzer/conf/hf_ner_onnx.yaml
omri374, line 20: There is no test that loads `backend` / `subfolder` / `file_name` from YAML through `RecognizerRegistryProvider`; the ort YAML path is only exercised via direct constructor calls in the e2e file, and this shipped config is not loaded by any test. The YAML path currently works only because `HuggingFaceRecognizerConfig` is `extra="allow"` and `_prepare_recognizer_kwargs` forwards unknown keys; a change to either would silently drop `backend` and fall back to torch with nothing failing.

Please add a (mocked) config-path test: enable a HuggingFaceNerRecognizer entry with `backend: ort` and the loader kwargs in a test YAML, load it through `RecognizerRegistryProvider`, and assert `rec.backend == "ort"` and that the kwargs reached `rec.model_kwargs`. The existing `test_hf_recognizer_with_chunker.yaml` test in `test_recognizer_registry_provider.py` is a good template.

presidio-analyzer/tests/test_huggingface_ner_recognizer_e2e.py
```diff
@@ -0,0 +1,165 @@
...
+    for r in results:
+        assert isinstance(r, RecognizerResult)
+        assert r.entity_type in ("PERSON", "LOCATION")
+        assert 0 <= r.start < r.end <= len(text)
+        assert 0.0 <= r.score <= 1.0
+
+
+def test_hf_recognizer_e2e_torch_backend():
+    """Real torch pipeline end-to-end on a tiny random model."""
+    pytest.importorskip("torch", reason="torch is not installed")
```
omri374, line 44: These tests are the only coverage of the real optimum load path, so I want them to keep running in CI. The analyzer suite already downloads the Stanford model in three existing tests, so the Hub dependency itself is not new. Three small changes would make this file cheap and stable:

1. **Cache the Hub directory in the analyzer test job.** Add an `actions/cache` step next to the existing uv cache in `.github/workflows/ci.yml`, with `path: ~/.cache/huggingface` and a key like `hf-hub-${{ runner.os }}-v1`. The matrix shares it across Python versions, and it also speeds up the existing Stanford-based tests.

2. **Pin model revisions.** The two Stanford tests assert exact detections from a mutable Hub repo. Pass `revision=<commit sha>` for both. For the ort test that already works through `model_kwargs`; for the torch test it needs the `revision` collision fixed first (see the comment on `_load_torch_pipeline`), which is a good reason to fix it in this PR.

3. **Add a `hub` marker without gating.** Register a `hub` marker in `pyproject.toml`, set `pytestmark = pytest.mark.hub` at the top of this file, and mention `-m "not hub"` in the test docs for offline local runs. Please do not make them opt-in only; that would mean they never run.

No need for local model fixtures; vendoring the ONNX file would be worse than the download.

presidio-analyzer/tests/test_huggingface_ner_recognizer_e2e.py
```diff
@@ -0,0 +1,165 @@
+"""End-to-end tests for HuggingFaceNerRecognizer with real model loading.
```
omri374, line 1: Can we please put those tests under test_huggingface_ner_recognizer?

omri374 (commented):
Thanks! and apologies for the extremely late reply. Left some comments but happy to hear your thoughts.

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
omri374: Addressed in dd14734e: the docstring and the docs page now list which keys apply to which backend, and `revision` / `token` / `trust_remote_code` are lifted to top-level `pipeline()` arguments on the torch backend.

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
```diff
@@ -103,6 +108,12 @@ class HuggingFaceNerRecognizer(LocalRecognizer):
         "DATE_TIME": "DATE_TIME",
     }
     DEFAULT_HF_TASK = "token-classification"
+    # Keys that transformers.pipeline() accepts as top-level arguments and also
+    # forwards itself to from_pretrained(); they must not be inside model_kwargs.
+    TORCH_PIPELINE_HUB_KWARGS = ("revision", "token", "trust_remote_code")
+    # Keys the ort path must also hand to optimum's pipeline() so the tokenizer
+    # is fetched from the same revision / with the same credentials as the model.
+    ORT_TOKENIZER_HUB_KWARGS = ("revision", "token")
```
RonShakutai, line 116: Could we pass `cache_dir`, `local_files_only`, and `trust_remote_code` to the tokenizer too? Currently, they only reach the model, so the tokenizer may try downloading files even when the user requested offline loading.

presidio-analyzer/tests/test_huggingface_ner_recognizer.py
```diff
@@ -602,6 +651,259 @@ def test_hf_recognizer_analyze_handles_malformed_pipeline_output(
...
+    mock_optimum = MagicMock()
+    mock_ort_model_cls = MagicMock()
+    with patch(HF_PIPELINE_PATH, new=MagicMock()):
+        with patch(
+            "presidio_analyzer.predefined_recognizers.ner."
+            "huggingface_ner_recognizer.optimum_pipeline",
+            mock_optimum,
+        ):
+            with patch(
+                "optimum.onnxruntime.ORTModelForTokenClassification",
```
RonShakutai, line 688: Could we mock Optimum without requiring it to be installed? These tests currently fail when the optional package is missing, even though they use mocks.

presidio-analyzer/tests/test_recognizer_registry_provider.py
```diff
@@ -384,6 +384,74 @@ def test_recognizer_registry_provider_yaml_character_chunker_config():
...
+    }
+    mock_ort_model_cls.from_pretrained.assert_called_once_with(
+        "onnx-community/stanford-deidentifier-base-ONNX",
+        subfolder="onnx",
+        file_name="model_quantized.onnx",
+    )
+    # Loader kwargs must not leak into the pipeline call.
+    _, pipeline_kwargs = mock_optimum_pipeline.call_args
+    assert "subfolder" not in pipeline_kwargs
+    assert "file_name" not in pipeline_kwargs
```
RonShakutai, line 429: Could we also check that the recognizer detects entities after loading it from YAML? The test currently checks only that it loads with the expected settings.

RonShakutai (changes requested):
Thanks for adding ONNX support. Left a few comments

presidio-analyzer/presidio_analyzer/predefined_recognizers/ner/huggingface_ner_recognizer.py
omri374: Done in d0a70903. `trust_remote_code` now goes top-level to optimum's `pipeline()`, and `cache_dir` / `local_files_only` are passed via its `model_kwargs`, which transformers copies into the tokenizer load (same path the torch backend already had). `subfolder` / `file_name` stay scoped to the model loader.

presidio-analyzer/tests/test_huggingface_ner_recognizer.py
omri374: Done in d0a70903. The tests now inject a fake `optimum` / `optimum.onnxruntime` into `sys.modules` (`patch_ort_model` helper) instead of patching the real package. Verified the ort tests pass with optimum blocked from import.

presidio-analyzer/tests/test_recognizer_registry_provider.py
omri374: Done in d0a70903. The mocked pipeline returns PATIENT and PHONE spans and the test asserts `analyze()` yields PERSON / PHONE_NUMBER at the expected offsets after loading from YAML.

PR #1735, opened 2025-09-26
fix: handle XA modality multi-frame grayscale dicom redaction

MRADULTRIPATHI:
## Change Description

Fixed handling of XA modality (X-Ray Angiography) DICOM images that were previously failing with  
`ValueError: Too many dimensions: 3 > 2` when passed to `Image.fromarray`.

- Added `_dicom_np_to_pil` helper to robustly handle conversion of multi-frame grayscale and RGB images.
- Updated `_get_most_common_pixel_value` to correctly handle multi-frame grayscale by taking the first frame.
- Updated `_set_bbox_color` to use `_dicom_np_to_pil` for consistency across grayscale and RGB images.
- Ensured redaction works as expected for XA modality and similar multi-frame images.
- Verified fix with local tests (`test_dicom_redactor.py`) → all tests passed.

## Issue reference

Fixes #1731

## Checklist

- [x] I have reviewed the [contribution guidelines](https://github.com/microsoft/presidio/blob/main/CONTRIBUTING.md)
- [x] I have signed the CLA (if required)
- [x] My code includes unit tests (local test added to validate XA modality fix)
- [x] All unit tests and lint checks pass locally
- [ x] My PR contains documentation updates / additions if required

PR #2094, opened 2026-06-23
feat(image-redactor): redact all frames of multi-frame DICOM images

### LINKED ISSUE #1731: Image Redactor - XA modality - ValueError: Too many dimensions: 3 > 2.
**Describe the bug**

Attempting to run image redaction on an XA modality image (X Ray Angiography) appears to be misinterpreting XA modality images as grayscale in some cases.

**To Reproduce**
You can obtain a test dicom file from [Aliza DICOM Conformance Tests] if you scroll down to "Enhanced XA Image Storage" and download the [Siemens, non-uniform, rotation](https://drive.google.com/file/d/1MtpbHAvVBX00Npb6K2vTyai0yt4jc1uY/view?usp=sharing) sample file.

When attempting to `redact_and_return_bbox` as described the presdio [getting started guide](https://microsoft.github.io/presidio/image-redactor/) using that file.

When attempting the above with the test file (and others that we've encountered in production) you will see an error similar to the following.

```python
  File ".../.venv/lib/python3.13/site-packages/presidio_image_redactor/dicom_image_redactor_engine.py", line 74, in redact_and_return_bbox
    image_pil = Image.fromarray(image_np, mode="L")
  File ".../.venv/lib/python3.13/site-packages/PIL/Image.py", line 3324, in fromarray
    raise ValueError(msg)
ValueError: Too many dimensions: 3 > 2.
```

**Expected behavior**
Image redaction should work as expected and described in the guide

**Screenshots**
See above error example

**Additional context**

Glad to offer more examples or more detail as needed.

### LINKED ISSUE #1737: Image Redactor - Support Multiframe Image Redaction
> I'm attempting to use presidio image redaction in cases where a dicom instance contains multiframe pixel data, an example of this would be a XA modality instance.  There are limitations with this library which prevent image redaction on multiframe instances, rather only one frame is used for PHI detection and the bboxes may not be applied properly or to all frames.  While it may not be common for multiframe instances to include PHI, it is possible for some PHI to be "burned in" to the pixel data on a multiframe instance.

> Rather than ignoring the extra frames on a multiframe instance, each frame that is included in a multiframe instance should be scanned for PHI and have that PHI redacted.

> I believe there are no workarounds in this case other than to not use this library for multiframe dicom instances

**Additional context**
> I'm including test files which were computer generated with fake pixel data and synthetic PHI 

[multiframe_test_images.zip](https://github.com/user-attachments/files/22619718/multiframe_test_images.zip)

Additionally, this was created as a spin off issue from #1735

Dashtid:
## Change Description

`DicomImageRedactorEngine` previously detected and redacted PII on only the first frame of a multi-frame DICOM instance, and raised `ValueError: Too many dimensions: 3 > 2` when redacting multi-frame (e.g. XA) grayscale instances, because the pixel array was passed to `Image.fromarray` with the frame axis still present.

This makes redaction frame-aware:

- Frame count is read from `NumberOfFrames` (0028,0008); single-frame behavior is unchanged.
- PII detection runs **per frame**. Multi-frame instances can carry burned-in PHI on any frame, and not necessarily in the same location (as raised in #1737), so each frame is scanned and redacted independently.
- `_add_redact_box` decodes the pixel data once, applies each frame's boxes to that frame, and writes it back as a single block (this also removes a fragile re-decode that relied on `pixel_array` caching in the previous in-place mutation).
- Per-frame rescaling preserves per-frame contrast.
- The DICOM verify engine visualizes a representative frame for multi-frame inputs instead of raising.

This also resolves the `Too many dimensions: 3 > 2` crash (#1731).

### Prior art

#1735 was an earlier attempt at the #1731 crash and was closed without merging. That approach stopped the exception by processing only the first frame; it was confirmed in-thread that the other frames were still not redacted. This PR instead scans and redacts every frame, which is the behavior requested in #1737.

### Tests

Added `tests/test_dicom_multiframe_redaction.py`:
- `_get_number_of_frames` (multi-frame, single-frame, missing element)
- per-frame rescaling returns a stacked uint8 array
- `_add_redact_box` redacts every frame for greyscale and color multi-frame instances, and leaves single-frame behavior unchanged
- end-to-end `redact()` applies redaction to every frame
- regression test for the #1731 crash

The image-redactor unit suite (200 tests) and `ruff` pass locally. Integration tests that require a Tesseract binary were not run locally and are left to CI.

## Issue reference

Fixes #1737
Fixes #1731

### Open questions
- Detection runs OCR per frame, which is faithful to #1737 but is N x OCR for large cine loops. Would you prefer an optional frame-deduplication step (skip re-OCR on identical frames) or an opt-out for very large series?
- For multi-frame inputs the verify engine currently visualizes the first frame rather than returning per-frame results. Is that acceptable, or would per-frame verification output be preferable?

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,7 @@ All notable changes to this project will be documented in this file.
 ### Image Redactor
 #### Added
 - Added Azure SDK credential support to `DocumentIntelligenceOCR` so callers can use Azure Identity credentials instead of API keys (#2085) (Thanks @mturac)
+- Multi-frame DICOM redaction support in `DicomImageRedactorEngine`. PII is now detected and redacted on every frame of a multi-frame instance (`NumberOfFrames > 1`), rather than only the first frame. This also resolves the `ValueError: Too many dimensions: 3 > 2` raised when redacting multi-frame (e.g., XA) grayscale instances (#2094). Fixes #1737, #1731.
 
 #### Changed
 - Updated image-redactor dependencies for `opencv-python`, `gunicorn`, `pytesseract`, and `azure-ai-formrecognizer` (#1978, #1977, #1980, #1986) (Thanks @dependabot)
--- a/presidio-image-redactor/presidio_image_redactor/dicom_image_pii_verify_engine.py
+++ b/presidio-image-redactor/presidio_image_redactor/dicom_image_pii_verify_engine.py
@@ -82,6 +82,12 @@ class DicomImagePiiVerifyEngine(ImagePiiVerifyEngine, DicomImageRedactorEngine):
 
         is_greyscale = self._check_if_greyscale(instance_copy)
         image = self._rescale_dcm_pixel_array(instance, is_greyscale)
+        num_frames = self._get_number_of_frames(instance)
+        if num_frames > 1 and image.shape[0] == num_frames:
+            # Verification visualizes a single representative frame; redaction
+            # (DicomImageRedactorEngine) handles every frame of a multi-frame
+            # instance.
+            image = image[0]
         if is_greyscale:
             # model L for grayscale, and has 8 bit-pixel to store the pixel value
             image_pil = Image.fromarray(image, mode="L")
--- a/presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
+++ b/presidio-image-redactor/presidio_image_redactor/dicom_image_redactor_engine.py
@@ -70,31 +70,52 @@ class DicomImageRedactorEngine(ImageRedactorEngine):
 
         is_greyscale = self._check_if_greyscale(instance)
         image_np = self._rescale_dcm_pixel_array(instance, is_greyscale)
-        if is_greyscale:
-            # model L for grayscale, and has 8 bit-pixel to store the pixel value
-            image_pil = Image.fromarray(image_np, mode="L")
-        else:
-            # model RGB, has 3x8 bit pixel available to store the value
-            image_pil = Image.fromarray(image_np, mode="RGB")
-        padded_image_pil = self._add_padding(image_pil, is_greyscale, padding_width)
+        num_frames = self._get_number_of_frames(instance)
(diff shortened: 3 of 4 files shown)


Dashtid:
Rebased onto current main to pick up the CI fixes from #2091/#2092.

Also addressed the two automated review points:

- `_normalize_bboxes_per_frame` now accepts the per-frame format (`[[...]]`) for single-frame instances as the docstring states; previously a nested list passed for a single-frame instance would have failed downstream in `_add_redact_box`.
- `_add_redact_box` now reuses the already-decoded frame for box-color selection instead of re-decoding `instance.pixel_array`, so the pixel data is genuinely decoded once.

Added a unit test for the single-frame per-frame-format case. Full image-redactor unit suite (201 tests) and ruff pass locally.

Dashtid:
This is rebased onto current `main` and CI is green across the matrix. Scope stays limited to `DicomImageRedactorEngine`: per-frame detection and redaction for multi-frame instances (fixes #1737 and the #1731 `Too many dimensions: 3 > 2` crash), with unit tests built on synthetic multi-frame fixtures.

Two design points from the description are open to your preference whenever someone picks this up:

1. Detection runs OCR per frame, which is faithful to #1737 but is N× on large cine loops. Would an optional frame-deduplication step (skip re-OCR on identical frames) or an opt-out be worth adding, or is per-frame the right default for now?
2. For multi-frame inputs the verify engine visualizes a representative frame rather than returning per-frame results. Is that acceptable, or would per-frame verification output be preferable?

Happy to adjust either way.

Dashtid:
I'm winding down my open-source work here to focus elsewhere, so I'm closing this rather than leaving it open indefinitely. The branch adds per-frame redaction for multi-frame DICOM, rebased with CI green; the two design questions still open in the thread — per-frame return vs. frame-stacking, and the rescale path — are the main things to settle before a merge. Anyone who wants the feature is welcome to build on the branch, and #1737 keeps the original context.

PR #2116, opened 2026-06-28
feat(analyzer): add recognizer-level threshold config

### LINKED ISSUE #1572: Recognizer-level thresholds
As a developer, I want to implement custom thresholds per recognizer in the Microsoft Presidio codebase so that different entities can have specific sensitivity levels for better accuracy and performance.

**Acceptance Criteria**

- Configuration: The system should allow setting custom thresholds for each recognizer and entity via a configuration file (e.g., YAML).
The configuration should be easily adjustable without modifying the core codebase.
- Threshold Application: The system should apply the specified custom thresholds during the recognition process. Each recognizer should use its respective threshold for evaluating entities.
- Validation: Ensure that the custom thresholds are correctly loaded and applied by running unit tests.
Validate that the recognition results vary according to the specified thresholds.
- Documentation: Update the documentation to include instructions on how to set and adjust custom thresholds for recognizers and entities. Provide examples of configuration files with custom thresholds.
- Error Handling: Implement error handling to manage cases where thresholds are not specified or incorrectly configured.
Ensure that meaningful error messages are provided to guide users in correcting configuration issues.
- Backward compatibility: The recognizer-level thresholding should be backward compatible, and work even if the user doesn't apply recognizer level thresholding.

rodboev:
## Change Description

Adds recognizer-level score thresholds in `presidio-analyzer`. Each recognizer entry can set a default threshold and entity-specific overrides, while `AnalyzerEngine` keeps final filtering after context enhancement and before duplicate removal.

```yaml
recognizer_registry:
  recognizers:
    - name: CreditCardRecognizer
      type: predefined
      score_thresholds:
        default: 0.4
        CREDIT_CARD: 0.7
```

`RecognizerRegistryProvider` validates and attaches normalized thresholds to each recognizer instance. The analyzer snapshots each selected recognizer's threshold mapping once per analysis request and resolves the producing recognizer by `recognizer_identifier`, so same-name recognizers, programmatic recognizers, and request-only ad hoc recognizers retain independent policies. When that identifier is missing or no longer matches a selected recognizer, the analyzer logs the fallback and uses `default_score_threshold`.

Supplying `analyze(score_threshold=...)` bypasses recognizer-level thresholds for that request and applies the supplied threshold to every result. When omitted, precedence is entity override, recognizer default, then `default_score_threshold`. The former top-level `recognizer_score_thresholds` key is removed because this PR has not merged.

This updates the analyzer and registry configuration docs, no-code example, default YAML examples, and changelog.

## Issue reference

Fixes #1572

## Tests

- `poetry run pytest tests/test_analyzer_engine.py tests/test_entity_recognizer.py -q` with 80 passed
- `poetry run ruff check presidio_analyzer/score_thresholds.py presidio_analyzer/entity_recognizer.py presidio_analyzer/analyzer_engine.py tests/test_analyzer_engine.py`

## Note on CHANGELOG

Updated `CHANGELOG.md` under `[unreleased]`, `Analyzer`, `Added`.

## Checklist


rodboev:

presidio-analyzer/presidio_analyzer/input_validation/schemas.py
```diff
@@ -135,6 +218,13 @@ def validate_analyzer_configuration(config: Dict[str, Any]) -> Dict[str, Any]:
                 config["default_score_threshold"]
             )
 
+        if "recognizer_score_thresholds" in config:
+            config["recognizer_score_thresholds"] = (
+                ConfigurationValidator.validate_recognizer_score_thresholds(
+                config["recognizer_score_thresholds"]
+            )
+            )
```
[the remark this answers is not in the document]
rodboev: Re-indented the constructor assignment so the nested normalization call is aligned and easier to edit next to the other constructor fields.

presidio-analyzer/presidio_analyzer/analyzer_engine.py
```diff
@@ -56,6 +59,7 @@ def __init__(
         default_score_threshold: float = 0,
         supported_languages: List[str] = None,
         context_aware_enhancer: Optional[ContextAwareEnhancer] = None,
+        recognizer_score_thresholds: Optional[Dict[str, Dict[str, float]]] = None,
```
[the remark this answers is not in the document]
rodboev: Widened the constructor annotation to accept per-recognizer numeric shorthand as `Dict[str, Union[float, Dict[str, float]]]`, which matches the documented API shape.

presidio-analyzer/presidio_analyzer/analyzer_engine.py
```diff
@@ -340,11 +348,103 @@ def __remove_low_scores(
...
+                if result.score >= self.__get_result_score_threshold(result)
+            ]
 
         new_results = [result for result in results if result.score >= score_threshold]
         return new_results
 
+    @staticmethod
+    def __normalize_recognizer_score_thresholds(
+        recognizer_score_thresholds: Optional[Dict[str, Dict[str, float]]]
+    ) -> Dict[str, Dict[str, float]]:
```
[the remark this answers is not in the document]
rodboev: Updated `__normalize_recognizer_score_thresholds()` to accept the same widened shorthand type, so the helper signature matches the supported config shape too.

presidio-analyzer/presidio_analyzer/analyzer_engine.py
```diff
@@ -340,11 +348,103 @@ def __remove_low_scores(
...
+        """Normalize shorthand threshold values into the nested mapping shape."""
```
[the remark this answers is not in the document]
rodboev: Dropped the duplicated constructor-side normalization logic and delegated directly to `ConfigurationValidator.validate_recognizer_score_thresholds()`, so YAML and direct-constructor paths now share one validation and normalization implementation.

presidio-analyzer/presidio_analyzer/input_validation/schemas.py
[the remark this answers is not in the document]
rodboev: Fixed in `adac24d5`: the `config["recognizer_score_thresholds"]` assignment now uses a normally aligned multi-line validation call, so the argument and closing parentheses are no longer offset.

presidio-analyzer/presidio_analyzer/analyzer_engine.py
```diff
@@ -15,6 +15,7 @@
     ContextAwareEnhancer,
     LemmaContextAwareEnhancer,
 )
+from presidio_analyzer.input_validation import ConfigurationValidator
```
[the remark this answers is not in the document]
rodboev: Fixed in `adac24d5`: `AnalyzerEngine` now imports `ConfigurationValidator` from `presidio_analyzer.input_validation.schemas`, while the configuration-boundary providers continue to use the public package-level validation API.

presidio-analyzer/presidio_analyzer/input_validation/schemas.py
```diff
@@ -44,6 +44,88 @@ def validate_score_threshold(threshold: float) -> float:
             )
         return threshold
```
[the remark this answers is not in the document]
rodboev: Fixed in `adac24d5`: `ConfigurationValidator.validate_score_threshold()` now rejects booleans and non-numeric values before the range check, so `default_score_threshold: true` fails validation instead of being treated as `1.0`. I also added direct helper and analyzer-config coverage for that path.

docs/analyzer/analyzer_engine_provider.md
```diff
@@ -70,10 +70,22 @@ recognizer_registry:
...
+- `supported_languages`: A list of supported languages that the analyzer will support.
+- `default_score_threshold`: A score that determines the minimal threshold for detection.
+- `recognizer_score_thresholds`: Optional per-recognizer thresholds keyed by the recognizer name from `RecognizerResult.RECOGNIZER_NAME_KEY`. Use `default` for a recognizer-wide fallback, then add entity names for overrides.
+- `nlp_configuration`: Configuration given to the NLP engine which will detect the PIIs and extract features for the downstream logic.
+- `recognizer_registry`: All the recognizers that will be used by the analyzer.
+
+Example:
+
+```yaml
+recognizer_score_thresholds:
```
omri374: Could you explain the motivation in creating a new section instead of defining thresholds at the recognizer level? 

Thanks! Apologies for the delay

rodboev: Sure. I put them at the analyzer level because `AnalyzerEngine` applies the cutoff after recognizer execution and context enhancement. Recognizers produce candidates; the analyzer owns final result filtering. The analyzer config already owns `default_score_threshold`, so the recognizer-specific map extends the same policy and keeps precedence in one place.

This also works across inline registries, separately loaded registry files, and programmatically created registries. Nesting the value in each recognizer entry would make `RecognizerRegistryProvider` carry analyzer policy back out of the registry-loading path. The map instead keys off `RecognizerResult.RECOGNIZER_NAME_KEY`, with entity overrides beneath it.

If you prefer colocation under `recognizer_registry.recognizers` for configuration discoverability, I can change the shape. I chose the separate section to match runtime ownership and the existing threshold config.

omri374: Yes, that's what I thought too. Before implementing, how would it look like if we did it at the recognizer level? Will the analyzer read this configuration from the recognizer registry?

rodboev: Yes. I’d keep filtering in `AnalyzerEngine`, but move the configuration into each recognizer entry:

```yaml
recognizer_registry:
  recognizers:
    - name: CreditCardRecognizer
      type: predefined
      score_thresholds:
        default: 0.4
        CREDIT_CARD: 0.7
```

`RecognizerRegistryProvider` would validate this block and attach the normalized thresholds to each instantiated recognizer. After context enhancement, `AnalyzerEngine` would use `recognizer_identifier` to find the producing recognizer and apply its entity override or default.

The precedence would remain: explicit `analyze(score_threshold=...)`, entity override, recognizer default, then `default_score_threshold`.

This would work with inline and separately loaded registry files. Programmatically created recognizers could expose the same threshold property. I think this is the cleaner shape because the configuration travels with the recognizer while the analyzer retains responsibility for final filtering. I’ll rework the PR around this design.

omri374: I think so too. Let's go with this approach. Thanks!

rodboev: I reworked this around the recognizer-level shape we discussed:

```yaml
recognizer_registry:
  recognizers:
    - name: CreditCardRecognizer
      type: predefined
      score_thresholds:
        default: 0.4
        CREDIT_CARD: 0.7
```

The registry provider pipeline now validates `score_thresholds` and attaches the normalized mapping while each config entry is still paired with the recognizer instances it creates. That also covers multi-language entries without matching instances back by name. Direct registry dict and YAML additions use the same strict normalizer.

`AnalyzerEngine` still performs final filtering after context enhancement. It uses `recognizer_identifier` to resolve the exact recognizer selected for that request, including ad hoc recognizers and otherwise indistinguishable same-name instances.

The precedence is: explicit `analyze(score_threshold=...)`, entity override, recognizer default, then `default_score_threshold`.

I removed the analyzer-level `recognizer_score_thresholds` map. The same recognizer property works for programmatically created recognizers, while inline and separately loaded registry files preserve raw values until strict validation, so booleans and numeric strings aren't coerced into valid thresholds. I also updated the registry docs and default configuration examples.

presidio-analyzer/presidio_analyzer/score_thresholds.py
```diff
@@ -0,0 +1,30 @@
+"""Validation helpers for recognizer score thresholds."""
+
+from collections.abc import Mapping
+from typing import Any, Dict
+
+
+def validate_score_threshold(threshold: Any) -> float:
+    """Validate a score threshold without coercing its input type."""
```
omri374: please align the docstring format to the rest of the repo (add `:param: threshold`)

omri374: Do we have to keep threshold as `Any`? can we use a more specific type?

presidio-analyzer/presidio_analyzer/score_thresholds.py
```diff
@@ -0,0 +1,30 @@
...
+    if not isinstance(threshold, (int, float)) or isinstance(threshold, bool):
+        raise ValueError(f"Score threshold must be numeric, got: {threshold}")
+    if not 0.0 <= threshold <= 1.0:
+        raise ValueError(
+            f"Score threshold must be between 0.0 and 1.0, got: {threshold}"
+        )
+    return threshold
+
+
+def normalize_score_thresholds(score_thresholds: Any) -> Dict[str, float]:
```
omri374: Same comments as above

presidio-analyzer/presidio_analyzer/entity_recognizer.py
```diff
@@ -53,6 +55,7 @@ def __init__(
         version: str = "0.0.1",
         context: Optional[List[str]] = None,
         country_code: Optional[str] = None,
+        score_thresholds: Any = None,
```
omri374: Can we provide a more specific type? I assume it's `dict[str, float]`?

presidio-analyzer/presidio_analyzer/analyzer_engine.py
```diff
@@ -330,21 +332,56 @@ def _enhance_using_context(
         return results
 
     def __remove_low_scores(
-        self, results: List[RecognizerResult], score_threshold: float = None
+        self,
+        results: List[RecognizerResult],
+        score_threshold: float = None,
```
omri374:
```suggestion
        score_threshold: Optional[float] = None,
```

docs/tutorial/08_no_code.md
```diff
@@ -119,6 +122,8 @@ recognizer_registry:
 """
 ```
 
+Each recognizer can set a `default` cutoff and entity-specific overrides in `score_thresholds`. An explicit request threshold takes priority, followed by an entity override, the recognizer default, and the analyzer's `default_score_threshold`.
```
omri374: This is an important implementation detail. If we pass `score_threshold`, then the recognizer-level thresholds are bypassed completely. Please make this more explicit ("explicit request threshold" might not be clear) and put this in the docs, not just the tutorial.

presidio-analyzer/presidio_analyzer/analyzer_engine.py
```diff
@@ -330,21 +332,56 @@ def _enhance_using_context(
...
         Remove results for which the confidence is lower than the threshold.
 
         :param results: List of RecognizerResult
         :param score_threshold: float value for minimum possible confidence
+        :param recognizers: recognizers selected for this request
         :return: List[RecognizerResult]
         """
         if score_threshold is None:
-            score_threshold = self.default_score_threshold
+            recognizers_by_id = {
```
omri374: if `id` is missing for some reason, it would not pick up the threshold. Consider adding a debug log

presidio-analyzer/presidio_analyzer/entity_recognizer.py
```diff
@@ -67,13 +70,24 @@ def __init__(
...
         self._country_code = self._resolve_country_code(country_code)
 
         self.load()
         logger.info("Loaded recognizer: %s", self.name)
         self.is_loaded = True
 
+    @property
+    def score_thresholds(self) -> Dict[str, float]:
+        """Return a defensive copy of this recognizer's score thresholds."""
+        return self._score_thresholds.copy()
```
omri374, line 84: This is copied on every result. Is it necessary?

docs/analyzer/recognizer_registry_provider.md
```diff
@@ -107,6 +110,7 @@ The recognizer list comprises of both the predefined and custom recognizers, for
   - `supported_entity`: the detected entity associated by the recognizer.
   - `deny_list`: A list of words to detect, in case the recognizer uses a predefined list of words.
   - `deny_list_score`: confidence score for a term identified using a deny-list.
+  - `score_thresholds`: optional score cutoffs for this recognizer. Use `default` as the recognizer-wide cutoff and entity names for overrides. An explicit request threshold takes priority, followed by an entity override, this recognizer default, and the analyzer's `default_score_threshold`.
```
omri374: please use "threshold" instead of "cutoff" to keep the jargon similar to the rest of the code base

presidio-analyzer/presidio_analyzer/analyzer_engine.py
omri374: please update this threshold with the new logic

docs/tutorial/08_no_code.md
omri374: here also, please use "threshold" instead of "cutoff"

presidio-analyzer/tests/test_analyzer_engine.py
```diff
@@ -558,6 +677,139 @@ def test_when_default_threshold_is_zero_then_all_results_pass(
...
+        language="en",
+        entities=["PERSON", "CREDIT_CARD", "URL", "DATE_TIME", "PHONE_NUMBER"],
+    )
+
+    scores = {result.entity_type: result.score for result in results}
+
+    assert scores == {"URL": 0.75, "DATE_TIME": 0.9}
+
+
+def test_explicit_score_threshold_overrides_config():
```
omri374: please make sure we are covering multiple competing thresholds scenarios.

omri374 (commented):
Thanks! Looks really good. A few minor and quickly fixable comments

rodboev:
Thanks for the detailed pass. I’ll tighten this as one recognizer-threshold consumer contract:

1. I’ll align the helper docstrings with the repository style and narrow the validation inputs from `Any` to `object`, which still lets the validators reject malformed runtime values without type coercion. The `EntityRecognizer` constructor will expose `Optional[Dict[str, float]]`, and the analyzer helper will use `Optional[float]` plus concrete lookup types.
2. I’ll keep the defensive `score_thresholds` property, but snapshot it once per selected recognizer for each analysis request. The resolver can then reuse that snapshot for every result instead of copying the mapping per result.
3. Missing or unmatched `recognizer_identifier` values will continue using `default_score_threshold`, with a debug log showing why recognizer-level lookup could not be applied.
4. I’ll update the API docstring, analyzer documentation, registry documentation, and tutorial to say directly that passing `analyze(score_threshold=...)` bypasses recognizer-level thresholds for that request. I’ll also use “threshold” consistently.
5. I’ll expand the analyzer coverage into competing-source cases where the explicit request, entity override, recognizer default, and engine default each determine the result.

This is the same consumer-integration invariant as the previous rework. The fresh enumeration covers the type and docstring surfaces, per-request state snapshot, producer lookup and fallback, permanent docs, tutorial wording, and competing-threshold tests. Existing registry construction and YAML attachment paths remain unchanged.

docs/analyzer/recognizer_registry_provider.md
```diff
@@ -107,6 +110,7 @@ The recognizer list comprises of both the predefined and custom recognizers, for
   - `supported_entity`: the detected entity associated by the recognizer.
   - `deny_list`: A list of words to detect, in case the recognizer uses a predefined list of words.
   - `deny_list_score`: confidence score for a term identified using a deny-list.
+  - `score_thresholds`: optional score thresholds for this recognizer. Use `default` as the recognizer-wide threshold and entity names for overrides. Supplying `analyze(score_threshold=...)` bypasses recognizer-level thresholds for that request. When omitted, precedence is an entity override, this recognizer default, then the analyzer's `default_score_threshold`.
```
omri374:
```suggestion
  - `score_thresholds`: optional score thresholds for this recognizer. Use `default` as the recognizer-wide threshold and entity names for overrides. Note that supplying `analyzer_engine.analyze(score_threshold=...)` bypasses recognizer-level thresholds for that request. The precedence is: Presidio Analyzer analyzer.analyze(score_threshold=...) > an entity specific threshold > a recognizer default threshold (`default`) > the Presidio Analyzer `default_score_threshold`.
```

docs/analyzer/analyzer_engine_provider.md
```diff
@@ -63,17 +63,22 @@ recognizer_registry:
...
-  - `supported_languages`: A list of supported languages that the analyzer will support.
-  - `default_score_threshold`: A score that determines the minimal threshold for detection.
-  - `nlp_configuration`: Configuration given to the NLP engine which will detect the PIIs and extract features for the downstream logic.
-  - `recognizer_registry`: All the recognizers that will be used by the analyzer. 
+- `supported_languages`: A list of supported languages that the analyzer will support.
+- `default_score_threshold`: A score that determines the minimal threshold for detection.
+- `nlp_configuration`: Configuration given to the NLP engine which will detect the PIIs and extract features for the downstream logic.
+- `recognizer_registry`: All the recognizers that will be used by the analyzer. Each recognizer entry can define `score_thresholds`, using `default` as its fallback and entity names for overrides.
+
+Supplying `analyze(score_threshold=...)` bypasses recognizer-level thresholds for that request and applies the supplied threshold to every result. When omitted, precedence is an entity-specific recognizer threshold, the recognizer's `default`, then `default_score_threshold`.
```
omri374:
```suggestion
!!! note "Defining confidence thresholds"
    Presidio supports multiple types of score thresholds: default ones (for the analyzer and each recognizer), per-request thresholds, and per-entity thresholds. Supplying `analyzer.analyze(score_threshold=...)` bypasses recognizer-level thresholds for that request and applies the supplied threshold to every result. The order of importance is: `analyzer.analyze(score_threshold=...)` > an entity specific threshold > a recognizer default threshold (`default`) > the Presidio Analyzer `default_score_threshold`.
```

omri374 (approved):

SharonHart (approved):

PR #2155, opened 2026-07-11
feat(analyzer): add Czech PII recognizers (CZ_*)

santiagotri:
## Summary

Adds 5 new pattern-based recognizers for Czech-language PII plus Czech `DATE_TIME`
date coverage, covering the most relevant entity types under GDPR / zákon č. 110/2019 Sb.,
and fixes the collision where Czech identifiers (rodné číslo, bank accounts) were
mis-detected as `PHONE_NUMBER`:

| Entity | Description | Legal Basis | Validation |
|---|---|---|---|
| CZ_BIRTH_NUMBER | Rodné číslo / birth number (YYMMDD/SSS or YYMMDD/SSSS, with or without slash) | § 13 z. č. 133/2000 Sb. | Mod 11 checksum, incl. the 1954–1985 historical exception; date semantics in regex (month ranges 01–12/21–32/51–62/71–82) |
| CZ_BANK_ACCOUNT | Domestic bank account (`[prefix-]number/bank code`) | vyhláška ČNB č. 169/2011 Sb. | Weighted mod 11 checksum on both prefix and account number |
| CZ_ID_CARD | Občanský průkaz / identity card (9 digits) | z. č. 269/2021 Sb. | Regex + structural checks, score=0.2 (context-driven) |
| CZ_PASSPORT | Cestovní pas (2 letters + 6–7 digits, or 7–8 digits) | z. č. 329/1999 Sb. | Regex + uppercase enforcement |
| CZ_DRIVER_LICENSE | Řidičský průkaz (2 letters + 6–9 digits) | § 103 z. č. 361/2000 Sb. | Regex + uppercase enforcement |
| DATE_TIME | Czech date notation ("12. dubna 1985", "12. 4. 1985") via `CzDateRecognizer` | — | Regex (day/month ranges, genitive + nominative month names) |

Also ships a Czech language support recipe (`docs/recipes/czech-language-support/`)
with a ready-to-use bilingual EN+CS NLP configuration (`spacy_en_cs.yaml`).

## Design decisions

- All recognizers follow the existing `PatternRecognizer` pattern and are placed in
  `predefined_recognizers/country_specific/czechia/`, mirroring the German
  implementation from #1909
- All are **disabled by default** (`enabled: false`) following the project convention
  for country-specific recognizers; registered for `supported_languages: [cs]` only,
  with `country_code: cz`
- **Checksum semantics deliberately differ from the German recognizers**: a checksum
  pass promotes to `MAX_SCORE`, but a checksum failure on an otherwise date-plausible
  number returns `None` (keep base score, let context words decide) rather than
  `False`. Rationale: legitimate rodná čísla exist that fail strict mod 11 — 9-digit
  pre-1954 numbers carry no check digit at all, and ~1000 numbers issued 1954–1985
  fall under the historical mod-10 exception — so a hard reject would drop real PII.
  `False` is reserved for structurally impossible values (all-zero suffix/account,
  bank code `0000`, wrong length)
- **PHONE_NUMBER collision fix**: `PhoneRecognizer` scores a flat 0.4 and previously
  won on `850412/0003` and on part of `19-123456789/0800`. Base scores plus context
  boosting put the Czech entities at 0.55–0.85 (1.0 on valid checksums), so
  score-based conflict resolution (as applied by presidio-anonymizer) picks the
  Czech entity; the genuine phone number `+420 601 234 567` is unaffected
- Context word lists contain **inflected Czech surface forms** ("účtu", "průkazu",
  "rodného", …) because the `LemmaContextAwareEnhancer` substring-matches lemmas and
  no Czech lemmatizer is available in the recommended pipeline
- Patterns use lookaround anchors (`(?<![\w/-])` …) instead of bare `\b` to prevent
  matches inside bank accounts, IBANs, and longer digit runs
- `CZ_PASSPORT` / `CZ_DRIVER_LICENSE` enforce an uppercase prefix in
  `validate_result()`: patterns are matched with `IGNORECASE` (global flags), so
  ordinary two-letter words before a number ("je 123456") would otherwise match
- `CzDateRecognizer` emits the standard `DATE_TIME` entity because no available Czech
  NER model emits DATE labels; `CZ_BANK_ACCOUNT` invalidates date-like `dd/yyyy`
  fragments (e.g. `04/2023`)
- **NLP engine**: neither spaCy nor Stanza publishes a Czech NER model, so the recipe
  uses the official multilingual `xx_ent_wiki_sm` for PERSON/LOCATION/ORG and
  documents a HuggingFace transformers model as the higher-quality upgrade path
- All recognizers use `global_regex_flags: 26` (IGNORECASE | MULTILINE | DOTALL)
  consistent with the rest of Presidio

## Test plan

-  7 test files: one per recognizer (`presidio-analyzer/tests/test_cz_*.py`) plus
      an end-to-end scenario (`test_czech_language_support.py`)
-  130 test cases — positive matches, negative matches, edge cases
- Separate `validate_result()` / `invalidate_result()` tests for the
      checksum-based recognizers (valid, historical-exception, failing, and
      structurally impossible inputs)
- End-to-end test analyzes a full Czech sample (name, birth date, rodné číslo,
      address, phone, e-mail, ID card, passport, driving licence, bank account,
      IBAN), asserts every span/entity, includes a before-state regression test for
      the PHONE_NUMBER collision, and verifies the exact anonymized output — using
      precomputed NLP artifacts so CI needs no model download
- All tests pass locally: `python -m pytest presidio-analyzer/tests/test_cz*.py`
- Full analyzer suite failure profile identical to `main` baseline (remaining
      failures are pre-existing optional-dependency/model gaps, verified via stash
      comparison)
- `ruff check` passes with no issues
- `default_recognizers.yaml` entries follow the existing country-specific
      pattern; default en-only registry verified unaffected via
      `RecognizerRegistryProvider` load test
- `docs/supported_entities.md` updated with Czechia section
- Czech language support recipe added under `docs/recipes/czech-language-support/`
-`CHANGELOG.md` entry added under `[unreleased]`

# Checklist


presidio-analyzer/tests/test_czech_language_support.py
```diff
@@ -0,0 +1,271 @@
...
+    ordered = sorted(
+        results, key=lambda r: (-r.score, -(r.end - r.start), r.start)
+    )
+    kept: List[RecognizerResult] = []
+    for result in ordered:
+        if all(
+            result.end <= other.start or result.start >= other.end for other in kept
+        ):
+            kept.append(result)
+    return sorted(kept, key=lambda r: r.start)
```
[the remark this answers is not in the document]
santiagotri: Good catch — Fixed in: `_resolve_conflicts()` now replicates `MERGE_SIMILAR_OR_CONTAINED`

presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/czechia/cz_passport_recognizer.py
```diff
@@ -0,0 +1,105 @@
...
+            "Cestovní pas (letter prefix)",
+            r"\b[A-Z]{2}\s?\d{6,7}(?![\w/-])",
+            0.3,
+        ),
+        Pattern(
+            "Cestovní pas (digits only)",
+            r"(?<![\w/.-])\d{7,8}(?![\w/-])",
+            0.1,
+        ),
+    ]
```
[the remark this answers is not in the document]
santiagotri: Right — these two letter-prefixed patterns were the only ones still using `\b`,
which does match right after `/` or `-` and contradicted the stated design.

Fixed in [21d6b6a](https://github.com/data-privacy-stack/presidio/pull/2155/changes/21d6b6a887c9774df6ba5e89c4b4c9b04b2a2a12) with the suggested lookbehind on both patterns, plus
regression tests for the newly blocked contexts (`123-AB123456`,
`spis/AB123456`, `19-CZ123456789`, `spis/CZ123456789`). The lookbehind is
strictly more restrictive than `\b` (it only adds the `/`/`-` exclusion), so
no legitimate matches are lost. Note it deliberately does not exclude `.`
like the digit-only patterns do: the dot exclusion there guards against
decimal fractions, which can't produce a letter prefix.

presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/czechia/cz_driver_license_recognizer.py
```diff
@@ -0,0 +1,99 @@
...
+
+    COUNTRY_CODE = "cz"
+
+    PATTERNS = [
+        Pattern(
+            "Řidičský průkaz (2 letters + digits)",
+            r"\b[A-Z]{2}\s?\d{6,9}(?![\w/-])",
+            0.3,
+        ),
+    ]
```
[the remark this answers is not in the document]
santiagotri: Right — these two letter-prefixed patterns were the only ones still using `\b`,
which does match right after `/` or `-` and contradicted the stated design.

Fixed in [21d6b6a](https://github.com/data-privacy-stack/presidio/pull/2155/changes/21d6b6a887c9774df6ba5e89c4b4c9b04b2a2a12) with the suggested lookbehind on both patterns, plus
regression tests for the newly blocked contexts (`123-AB123456`,
`spis/AB123456`, `19-CZ123456789`, `spis/CZ123456789`). The lookbehind is
strictly more restrictive than `\b` (it only adds the `/`/`-` exclusion), so
no legitimate matches are lost. Note it deliberately does not exclude `.`
like the digit-only patterns do: the dot exclusion there guards against
decimal fractions, which can't produce a letter prefix.

docs/recipes/czech-language-support/README.md
```diff
@@ -0,0 +1,179 @@
...
+bank account numbers, official document numbers and Czech-notation dates —
+alongside the English model so that bilingual (EN + CS) documents are
+handled correctly.
+
+This recipe provides:
+
+- `spacy_en_cs.yaml` — a ready-to-use NLP engine configuration that loads both
+  `en_core_web_lg` and the multilingual `xx_ent_wiki_sm` (spaCy publishes no
+  Czech-specific pretrained pipeline; the multilingual WikiNER model is the
+  official artifact that covers Czech PERSON/LOCATION/ORG detection)
```
omri374: Worth mentioning that xx_ent_wiki_sm was not trained on Czech-specific, so one should not assume very high accuracy specifically on Czech entities

docs/recipes/czech-language-support/README.md
```diff
@@ -0,0 +1,179 @@
+# Czech Language Support
```
omri374, line 1: Very well written! Thanks!

presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/czechia/cz_bank_account_recognizer.py
```diff
@@ -0,0 +1,148 @@
...
+        # An account number of all zeros is never assigned.
+        if number.lstrip("0") == "":
+            return False
+
+        if self._mod11(number) and (not prefix or self._mod11(prefix)):
+            return True
+
+        return None
+
+    def invalidate_result(self, pattern_text: str) -> Optional[bool]:
```
omri374, line 161: Bank account pattern might flag  Czech legal citations as accounts (often at score 1.0)

The NNN/YYYY form of Czech legal references collides with the account pattern (?:\d{2,6}-)?\d{2,10}/\d{4}, and this invalidate_result guard only filters them when the numerator is 1–2 digits and <= 31 (the date case). Real law references are 3+ digits and pass straight through:

262/2006 (the actual Zákoník práce / Labour Code) passes the _mod11 checksum, so validate_result returns True and it's promoted to MAX_SCORE (1.0).
115/2010 — same, and the test ("115/2010", None) documents that it isn't invalidated today.
169/2011, 133/2000, 50/2019 — fail the checksum but still surface at the base 0.4.
This matters because the accompanying recipe targets legal / public-administration text, where these citations are extremely common — so it's a high-frequency false positive for the intended audience, not an edge case.

Suggested fix: reject a slash match when there's no - prefix and the 4-digit bank_code is in the 1900–2099 range, regardless of numerator length (drop the len(account_part) <= 2 / <= 31 restriction for that case). Better still, validate bank_code against the finite set of real ČNB bank codes — a year like 2006/2010 is not an issued code. Please also add positive assert no match tests for 262/2006 and 115/2010.

omri374:
Hi @santiagotri, thanks for this PR! great work. There are a few minor comments to address and we can merge it

presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/czechia/cz_bank_account_recognizer.py
santiagotri: You are right thanks. Changes applied on [2d63617](https://github.com/data-privacy-stack/presidio/pull/2155/changes/2d636174ae3626c94bf0be987b7059bea7a389bd) :

1. `invalidate_result` now drops any prefix-less match with a ≤3-digit
   numerator and a year-like 1900–2099 "bank code" (Czech laws never exceed
   3 digits/year), which covers 262/2006, 115/2010, 89/2012, 169/2011,
   50/2019, 133/2000 and subsumes the old date rule — even when the
   numerator passes mod 11, as 262 and 115 do.
2. A checksum pass over a year-like code the ČNB never issued is no longer
   promoted to MAX_SCORE (kept at base 0.4, context decides).

Added the requested no-match tests for 262/2006 and 115/2010, plus 89/2012,
an in-context citation ("zákon č. 262/2006 Sb., zákoník práce") and a
positive test locking in the FioBanka-shaped account (2000145399/2010 → 1.0).

I used the [ČNB bank codes last release July 2026](https://www.cnb.cz/export/sites/cnb/en/payments/.galleries/accounts_bank_codes/download/bank_codes_CR.pdf)

docs/recipes/czech-language-support/README.md
santiagotri: You are right. Sorry for that. Fixed in [fd5036d](https://github.com/data-privacy-stack/presidio/pull/2155/changes/fd5036d3a6120b0a977773c4c4b6253bbd832c72)

docs/recipes/czech-language-support/README.md
santiagotri: Thanks to you. I'm new to this project contribution so I will appreciate all kind of comments or suggestions 👍

Issue #1760, opened 2025-10-22
Can multiple GLiNERRecognizer instances be configured via YAML?


Mehrsary:
Hi, I am trying to configure multiple GLiNERRecognizer instances using Presidio's YAML-based setup (no-code). Is it currently possible to specify more than one GLiNER recognizer with different models or parameters directly in YAML?

something like:

```yaml
recognizers:
  - name: GLiNERRecognizer
    type: predefined
    model_name: "urchade/gliner_multi_pii-v1"
    threshold: 0.4
    ...

  - name: GLiNERRecognizer
    type: predefined
    model_name: "gliner-community/gliner_small-v2.5"
    threshold: 0.25
    ...
```

omri374:
This is currently not supported. One potential way forward is to create a new gliner recognizer class, e.g.:
```py
class GlinerRecognizerSmall(GlinerRecognizer):
     pass
```
Make sure it's in the path, and define this new one in the YAML. I haven't tested this but by the way the YAML parsing works, it should work. Another option is to define the recognizers via code which is more customizable.

Mehrsary:
Would something like this make sense to you? #1761

PratikWayase:
Hi @omri374, I’d like to work on this if possible.

I saw your suggestion about subclassing, but I think a native YAML solution would better fit the “no-code” goal. I’m considering modifying the YAML loader to support an optional instance_id field that creates unique registry keys while keeping the class name the same.

Does this align with Presidio’s design goals? Are there any pitfalls I should watch out for?

omri374:
Hi, thanks! We've made a lot of changes to the yaml based config, and this could be solved already. We'll check, feel free to test this yourself too

omri374:
Ok there seems to be a gap. It would be great if you could suggest a fix

RonShakutai:
Hi @PratikWayase , @omri374 
Did you get a chance to look at this PR?
https://github.com/microsoft/presidio/pull/1819

You should be able to define two recognizers, but it would be better if each one had a different name(same class name).

If you want this to work seamlessly without having to provide a name explicitly, we could probably handle that with a small PR. The idea would be to change the GLiNER recognizer constructor so it initializes the class name by concatenating the recognizer name and model name. Then behind the scenes, the class name could be built automatically instead of being taken directly from the config.

I’d be happy to suport you if you want to write it.

RonShakutai:
For example, this config should work now, @PratikWayase:
```python
recognizers:
[...]

PR #2156, opened 2026-07-12
Polish docs and comments in presidio (#1760)

bglglzd:
This is a focused change for the cited issue with minimal side effects.

Related to #1760.


SharonHart:
Thanks for the contribution. This PR only removes trailing whitespace and does not address #1760, which concerns configuring multiple GLiNER recognizers. Some removed spaces may also be intentional Markdown line breaks. Given the lack of a functional or documentation improvement, I don’t think this change should be merged.

Issue #2136, opened 2026-06-30
Feature Request: Healthcare Recognizer for Common Healthcare Identifiers (Member ID, Claims, Prior Authorization, etc.)


bhargavikalicheti:
**Is your feature request related to a problem? Please describe.**
Many organizations might be using Presidio as part of healthcare AI, conversational AI, document processing, and Retrieval-Augmented Generation (RAG) pipelines to detect and redact sensitive information before processing by LLMs.

While Presidio provides excellent support for common PII entities such as names, phone numbers, SSNs, email addresses, and credit cards, it currently doesn't support for healthcare-specific identifiers that are frequently encountered in regulated healthcare systems.

Examples include:

Insurance Member IDs
Claim Numbers
Prior Authorization Numbers
Prescription Numbers
Provider Tax IDs
Referral Numbers

As a result, healthcare organizations must implement custom recognizers independently, leading to duplicated effort, inconsistent implementations, and additional maintenance.

**Describe the solution you'd like**
I would like to propose adding a Healthcare Recognizer Pack that extends Presidio with commonly used healthcare-specific entity recognizers.

The initial implementation could include recognizers for:

MEMBER_ID
CLAIM_NUMBER
PRIOR_AUTHORIZATION_NUMBER
PRESCRIPTION_NUMBER
PROVIDER_TAX_ID
REFERRAL_NUMBER

Each recognizer would leverage the existing Presidio framework, including:

Pattern-based detection
Context-aware matching
Confidence scoring
Optional validation logic where applicable
Unit tests and documentation

This would enable healthcare developers to use Presidio out of the box for protecting healthcare identifiers in conversational AI, Voice AI, LLM, RAG, and document-processing workflows.

**Describe alternatives you've considered**
Currently, the only option is to develop custom recognizers for each healthcare identifier and maintain them independently within individual projects.

Although this approach works, it results in duplicated implementations across organizations and makes it difficult to establish reusable, community-maintained healthcare recognizers.

Another alternative is using generic regex-based detectors outside Presidio, but these lack Presidio's context-aware scoring, analyzer integration, and extensibility.

**Additional context**
Healthcare AI adoption is rapidly expanding across conversational AI, contact centers, document intelligence, and LLM-based applications. Having reusable healthcare entity recognizers within Presidio would simplify adoption in regulated environments and provide a consistent approach to detecting healthcare-specific identifiers.

If this feature aligns with the project roadmap, I'd be happy to contribute the implementation, including:

[...]

PR #2159, opened 2026-07-12
feat(analyzer): Add Healthcare identifiers recognizer

bhargavikalicheti:
## Change Description

Adds conservative, context-aware US healthcare identifier recognizers to Presidio Analyzer.

New disabled-by-default recognizers:
- `US_HEALTH_INSURANCE_MEMBER_ID`
- `US_PRIOR_AUTHORIZATION_NUMBER`
- `US_CLAIM_NUMBER`
- `US_PRESCRIPTION_NUMBER`
- `US_REFERRAL_NUMBER`
- `US_PROVIDER_TAX_ID`

These recognizers combine plausible identifier patterns with healthcare/insurance context enhancement and entity-specific score thresholds to reduce false positives. Pattern-only matches remain available to callers that explicitly lower the analysis threshold, such as structured-data workflows.

## Issue reference

Fixes **Feature Request: Healthcare Recognizer for Common Healthcare Identifiers (Member ID, Claims, Prior Authorization, etc.)
 #2136**

## Checklist

- [X] I have reviewed the [contribution guidelines](https://github.com/data-privacy-stack/presidio/blob/main/CONTRIBUTING.md)
- [X] I agree to follow this project's [Code of Conduct](https://github.com/data-privacy-stack/presidio/blob/main/CODE_OF_CONDUCT.md)
- [X] I confirm that I have the right to submit this contribution and that it does not knowingly contain proprietary or confidential code.


bhargavikalicheti:
Hi @SharonHart  @omri374 , No rush -  gentle reminder on PR whenever you get a chance. Thank you!

presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/us/us_health_insurance_member_id_recognizer.py
```diff
@@ -0,0 +1,105 @@
...
+    :param supported_language: Language this recognizer supports
+    :param supported_entity: The entity this recognizer can detect
+    :param context_window: Number of characters before/after a match to scan
+    for context.
+    """
+
+    COUNTRY_CODE = "us"
+
+    PATTERNS = [
+        Pattern(
```
omri374, line 37: Please add a reference to the sources of the logic/patterns

presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/us/us_healthcare_admin_recognizers.py
```diff
@@ -0,0 +1,311 @@
+"""Recognizers for US healthcare administrative identifiers."""
+
+from typing import List, Optional
+
+from presidio_analyzer import Pattern, PatternRecognizer, RecognizerResult
+
+
+class _ContextRequiredPatternRecognizer(PatternRecognizer):
```
omri374: Why not handle this through the threshold mechanism? We recently added recognizer and entity specific thresholds, so filtering entities if they didn't meet the bar without context can be handled this way. Moreover, in some cases we do want to identify these entities without context (e.g. in the presidio-structured package, where we count how many out of the records in a column were identified as X)

omri374 (commented):
Thanks! Please add references to be able to trace where the regex pattern is coming from, and see the comment around context management.

presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/us/us_health_insurance_member_id_recognizer.py
bhargavikalicheti: @omri374  - Added the reference details

presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/us/us_healthcare_admin_recognizers.py
bhargavikalicheti: @omri374  Handled through threshold mechanism. please check and let me know.

bhargavikalicheti:
@SharonHart @omri374 Hi! Just a friendly follow up on my PR whenever you have a chance. I'd appreciate a review when your schedule allows. Please let me know if there are any changes you'd like me to make. Thanks!

bhargavikalicheti:
@omri374 @SharonHart  - Hi! just checking if you are okay with this and ready to merge please?

presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/us/us_health_insurance_member_id_recognizer.py
```diff
@@ -0,0 +1,70 @@
...
+    single universal checksum or format. To avoid broad matching of generic
+    alphanumeric IDs, this recognizer requires both:
+    - a plausible alphanumeric member ID pattern, and
+    - nearby healthcare/insurance context.
+
+    CMS consumer guidance illustrates that insurance cards carry payer-defined
+    member numbers. The default regex is therefore a conservative heuristic and
+    can be replaced through the ``patterns`` constructor argument.
+
+    Reference: https://www.cms.gov/files/document/2020-c2c-how-use-health-coverage-slide-deck.pdf
```
omri374: I couldn't find any reference to the insurance number ID in this deck... Maybe I missed it?

presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/us/us_health_insurance_member_id_recognizer.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+
+    COUNTRY_CODE = "us"
+
+    PATTERNS = [
+        Pattern(
+            "Health insurance member ID (alphanumeric)",
+            r"\b(?=[A-Z0-9-]{6,20}\b)(?=[A-Z0-9-]*[A-Z])"
+            r"(?=[A-Z0-9-]*\d)[A-Z]{1,5}-?[A-Z0-9]{5,14}\b",
+            0.3,
```
omri374: 0.3 is high for how broad this is. `global_regex_flags` includes `re.IGNORECASE` by default, so despite the uppercase character classes this matches any 6 to 20 char token starting with letters and containing a digit:

```
covid19, sha256, iphone15pro, rfc2119, gpt4turbo, ICD10CM123, ABC-1234567
```

Package convention for this specificity is 0.05 to 0.1 with an explicit strength label (`UsPassportRecognizer`: `"Passport (very weak)"`, 0.05; `UsBankRecognizer`: `"Bank Account (weak)"`). Suggest 0.1, renaming to `"Health insurance member ID (weak)"`, and noting the case-insensitive runtime behaviour in the docstring.

presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/us/us_healthcare_admin_recognizers.py
```diff
@@ -0,0 +1,251 @@
...
+    universal syntax, the default pattern conservatively requires an ``RX``
+    prefix.
+
+    Reference: https://www.cms.gov/files/document/cms-medicare-part-d-340b-repository-companion-guide-v-1.pdf
+    """
+
+    PATTERNS = [
+        Pattern(
+            "Prescription number",
+            r"\bRX-?\d{6,12}\b",
```
omri374: Requiring a literal `RX` inside the identifier means near-zero recall on real documents. Real prescription numbers are bare digits next to an `Rx#` label; the same applies to `CLM` (Medicare ICNs are 13 to 15 digits) and `PA`.

Anchoring on the label instead of the prefix improves precision and recall together. Matching uses the `regex` module, so variable-width lookbehind works and keeps the label out of the span:

```python
r"(?<=\b(?:rx|prescription)\s*(?:#|no\.?|number)?\s*:?\s*)\d{6,12}\b"
```

Verified: matches `Rx #1234567`, `Prescription number: 7654321`, `prescription 4455667`; does not match `The claim 1234567 was paid`.

presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/us/us_healthcare_admin_recognizers.py
```diff
@@ -0,0 +1,251 @@
...
+    ``XX-XXXXXXX`` to avoid treating SSNs as provider organization IDs.
+
+    CMS reference: https://www.cms.gov/outreach-and-education/mln/wbt/mln4462429-mln-wbt-1500/1500/lesson04/12/index.html
+    IRS format reference: https://www.irs.gov/instructions/iss4
+    """
+
+    PATTERNS = [
+        Pattern(
+            "Provider tax ID",
+            r"\b\d{2}-\d{7}\b",
```
omri374: A bare EIN shape with one generic context word is too loose:

```
'Provider phone extension 12-3456789' -> [('12-3456789', 0.7)]
'provider 00-0000000 listed'          -> [('00-0000000', 0.7)]
```

The IRS publishes the valid campus prefix set for the first two digits; restricting to those is cheap and removes a large share of false positives. Also worth adding `tax id`, `tin`, `ein`, `billing` to `CONTEXT`.

presidio-analyzer/tests/test_us_health_insurance_member_id_recognizer.py
```diff
@@ -0,0 +1,122 @@
...
+)
+def test_when_member_id_has_context_then_detected(
+    text, expected_positions, recognizer, entity
+):
+    """Test context raises plausible member IDs above the threshold."""
+    results = analyze_member_id(text, recognizer, entity)
+    results = sorted(results, key=lambda result: result.start)
+    assert len(results) == len(expected_positions)
+    for result, (start, end) in zip(results, expected_positions):
+        assert_result(result, entity, start, end, 0.6499999999999999)
```
omri374: Please use `pytest.approx(0.65)` rather than asserting the exact float.

presidio-analyzer/tests/test_us_health_insurance_member_id_recognizer.py
```diff
@@ -0,0 +1,122 @@
...
+    [
+        "ABC123456789",
+        "Please store HPN12345A9 in the table",
+        "Order number ABC123456789 shipped yesterday",
+        "Tracking number ZX-987654321 is in transit",
+        "Case number HPN12345A9 is pending review",
+        "Claim number BCBSM1234567 was denied",
+    ],
+)
+def test_when_member_id_lacks_insurance_context_then_below_threshold(
```
omri374, line 142: Other gaps worth covering: lowercase and mixed case inputs; multiple IDs in one text; trailing punctuation; negative pattern cases for the admin recognizers (`PA-12345` too short, `PA-1234567890123` too long)

presidio-analyzer/tests/mocks/nlp_engine_mock.py
```diff
@@ -42,3 +43,14 @@ def get_supported_entities(self) -> List[str]:
 
     def get_supported_languages(self) -> List[str]:
         return ["en"]
+
+
+class ContextAwareNlpEngineMock(NlpEngineMock):
```
omri374: This diverges from the repo convention: `test_context_support.py` uses the `spacy_nlp_engine` fixture for context tests. The mock tokenizes with `\b[\w-]+\b` and uses lowercased tokens as pseudo-lemmas, neither of which matches spaCy, so green tests here do not demonstrate production behaviour.

Prefer the existing fixture. If the mock stays, document the limitation in its docstring and hold off on exporting it from `tests/mocks/__init__.py` until more than one module needs it.

docs/supported_entities.md
```diff
@@ -34,9 +33,15 @@ For more information, refer to the [adding new recognizers documentation](analyz
 |US_BANK_NUMBER|A US bank account number is between 8 to 17 digits.|Pattern match and context|
 |US_DRIVER_LICENSE|A US driver license according to <https://ntsi.com/drivers-license-format/>|Pattern match and context|
 |US_ITIN | US Individual Taxpayer Identification Number (ITIN). Nine digits that start with a "9" and contain a "7" or "8" as the 4 digit.|Pattern match and context|
+|US_CLAIM_NUMBER|A US healthcare claim identifier used in billing and claims processing.|Pattern match and required context|
+|US_HEALTH_INSURANCE_MEMBER_ID|A US health insurance member or subscriber identifier printed on an insurance card. Detection requires healthcare or insurance context.|Pattern match and required context|
```
omri374: The PR description mentions negative-context checks for order, tracking, case and invoice numbers. There is no deny list or negative-context mechanism in the code; those tests pass only because the positive context word is absent. Either implement it or drop the claim.

presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/us/us_healthcare_admin_recognizers.py
```diff
@@ -0,0 +1,251 @@
+"""Recognizers for US healthcare administrative identifiers."""
+
+from typing import Dict, List, Optional
+
+from presidio_analyzer import Pattern, PatternRecognizer
+
+
+class _HealthcareAdminPatternRecognizer(PatternRecognizer):
```
omri374: What's the added value of using this class?

omri374 (commented):
Thanks! Left a few comments, hopefully all are easy to change and clear.

bhargavikalicheti:
Hi @omri374, thank you for the detailed review. I’ve addressed the feedback:
clarified the identifier sources and pattern rationale
reduced broad member ID confidence
anchored administrative IDs to appropriate labels
restricted Provider EIN matching to valid IRS prefixes
replaced exact floating-point assertions with pytest.approx
expanded positive, negative, boundary, casing, punctuation, and multiple-ID tests
replaced the custom NLP mock with the existing spaCy fixture
removed the healthcare administrative base class
corrected the documentation and PR description to describe context enhancement and entity thresholds, without claiming negative-context checks
The PR is ready for another review whenever you have a chance. Thank you!

omri374:
Thank you @bhargavikalicheti. Could you please fix the small conflict on the analyzer yaml?

bhargavikalicheti:
> ###
> * A conversation must be resolved before this pull request can be merged. At least 1 approving review is required by reviewers with write access.
> * All comments must be resolved.

done @omri374. Thanks.

presidio-analyzer/presidio_analyzer/conf/default_recognizers.yaml
```diff
@@ -84,7 +84,42 @@ recognizers:
     enabled: false
     country_code: us
 
-  - name: AbaRoutingRecognizer
```
omri374, line 87: Why is this deleted?

bhargavikalicheti: sorry i didnt mean to..couldve deleted during conflict fix..i put it back.

omri374: No worries

presidio-analyzer/presidio_analyzer/conf/default_recognizers.yaml
```diff
@@ -91,6 +91,48 @@ recognizers:
     enabled: false
     country_code: us
 
+  - name: UsHealthInsuranceMemberIdRecognizer
```
omri374, line 94: Enabling these recognizers via YAML silently wipes their score thresholds (applies to all six new entries)

Enabling any of these six recognizers via this config which is the only supported path, since they ship `enabled: false` — erases the score thresholds the constructors set. `RecognizerListLoader.get()` runs `recognizer.score_thresholds = normalize_score_thresholds(conf.get("score_thresholds"))` (`recognizers_loader_utils.py:435`), which turns the absent key into `{}` and overwrites the `{'US_CLAIM_NUMBER': 0.6}`-style defaults from `__init__`.

Verified with a real `AnalyzerEngine`: after flipping `enabled: true`, `"Tracking number CLM456789123 is active"` → `US_CLAIM_NUMBER` at 0.1, and `"v2patch10build7"` → `US_HEALTH_INSURANCE_MEMBER_ID` at 0.1 — exactly the false positives the PR's tests assert are suppressed (those tests use direct instantiation + `add_recognizer`, so they never hit this).

Two possible fixes: declare `score_thresholds` explicitly on these six yaml entries, or make the loader only assign when the conf actually provides thresholds. The second also fixes this for user-supplied configs that omit the key.

presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/us/us_healthcare_admin_recognizers.py
```diff
@@ -0,0 +1,317 @@
...
+            "Referral number (labelled)",
+            r"(?<=\b(?:referral|infusion\s+referral)"
+            r"(?:\s*(?:#|no\.?|number|id)\s*:?\s*|\s*:\s*|\s+))"
+            r"(?:(?:REF|INF)-?)?\d{6,12}\b",
+            0.35,
+        ),
+        Pattern(
+            "Referral number (weak prefixed)",
+            r"\b(?:REF|INF)-?\d{6,12}\b",
+            0.1,
```
omri374: **Weak patterns can never fire, even with a full context boost** (same math in all four admin recognizers)

The weak patterns (score 0.1) in all four admin recognizers can never clear their 0.6 entity threshold: the maximum context boost gives 0.1 + 0.35 = 0.45. Verified: `"Your specialty referral was approved under REF-1234567 today."` returns `[]` from `AnalyzerEngine` despite the `referral` context word.

In practice the `CONTEXT` lists are dead weight for weak matches — they only ever re-boost labelled matches that already self-boost via their own label token — so REF-/CLM-/RX-/PA-prefixed IDs that aren't immediately preceded by their label are always missed, and that PII passes through anonymization. Either raise the weak scores (e.g. 0.25, so 0.25 + 0.35 ≥ 0.6) or lower the thresholds, the way the member-ID recognizer does (0.1 + 0.35 = 0.45 ≥ 0.4).

---

presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/us/us_healthcare_admin_recognizers.py
```diff
@@ -0,0 +1,317 @@
...
+        r"7[1-7]|8[0-8]|9[0-5]|9[89])"
+    )
+
+    PATTERNS = [
+        Pattern(
+            "Provider tax ID (labelled)",
+            r"(?<=\b(?:(?:(?:billing|rendering|healthcare)\s+provider|"
+            r"provider\s+organization|provider)\s+(?:tax\s*(?:id|number|"
+            r"identification\s+number)|tin|ein)|billing\s+provider)"
+            r"(?:\s*:\s*|\s+))" + VALID_EIN_PREFIX + r"-\d{7}\b",
```
omri374: **TIN labelled pattern misses `TIN#` / `EIN No.` forms**

This lookbehind only allows `:` or whitespace after the label (`(?:\s*:\s*|\s+)`), while every sibling recognizer also accepts a `#|no\.?|number|id` separator. So `"Provider TIN# 12-3456789"` and `"Billing provider EIN No. 12-3456789"` miss the labelled 0.35 pattern.

presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/us/us_healthcare_admin_recognizers.py
```diff
@@ -0,0 +1,317 @@
...
+        patterns = patterns if patterns else self.PATTERNS
+        context = context if context else self.CONTEXT
+        super().__init__(
+            supported_entity=supported_entity,
+            patterns=patterns,
+            context=context,
+            supported_language=supported_language,
+            name=name,
+        )
+        self.score_thresholds = (
```
omri374: **Six copies of the same `__init__`/threshold scaffold**

The `score_thresholds if score_thresholds is not None else {supported_entity: X}` idiom plus the surrounding `__init__` scaffold is copy-pasted six times across the two files (here and at lines 120, 185, 245, 313, plus the member-ID file). A small shared base class (or a `score_thresholds` pass-through parameter) would collapse this — worth doing now, since the fix for the yaml-loader threshold bug otherwise has to be applied in six places, and any missed spot quietly reintroduces it for one entity.

---

presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/us/us_health_insurance_member_id_recognizer.py
```diff
@@ -0,0 +1,75 @@
...
+    :param supported_entity: The entity this recognizer can detect
+    :param score_thresholds: Optional default and entity-specific score thresholds
+    """
+
+    COUNTRY_CODE = "us"
+
+    PATTERNS = [
+        Pattern(
+            "Health insurance member ID (weak)",
+            r"\b(?=[A-Z0-9-]{6,20}\b)(?=[A-Z0-9-]*[A-Z])"
```
omri374: **Redundant lookahead**

`(?=[A-Z0-9-]*[A-Z])` is redundant: the consuming part starts with `[A-Z]{1,5}`, so every match already contains a letter. Removing it yields the identical match set and makes an already four-part regex easier to audit.

---

bhargavikalicheti:
Hi @omri374, thank you for the follow-up review. I’ve addressed the latest feedback:
added explicit YAML score thresholds for all six healthcare recognizers
added regression tests confirming thresholds are retained when enabled through RecognizerRegistryProvider
added support and tests for Provider TIN# and Billing provider EIN No. formats
removed the healthcare entries from CHANGELOG.md per the contribution guidelines
The latest changes are ready for review whenever you have a chance. Thank you!

bhargavikalicheti:
@omri374 just checking if you are good with these changes.

presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/us/us_health_insurance_member_id_recognizer.py
```diff
@@ -0,0 +1,75 @@
...
+        patterns = patterns if patterns else self.PATTERNS
+        context = context if context else self.CONTEXT
+        super().__init__(
+            supported_entity=supported_entity,
+            patterns=patterns,
+            context=context,
+            supported_language=supported_language,
+            name=name,
+        )
+        self.score_thresholds = (
```
omri374: Why is this needed? This would be used even if a user passes a regular confidence threshold without an ability to change it.

presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/us/us_healthcare_admin_recognizers.py
```diff
@@ -0,0 +1,319 @@
...
+            score_thresholds
```
omri374: Same here

presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/us/us_healthcare_admin_recognizers.py
```diff
@@ -0,0 +1,319 @@
...
+        patterns = patterns if patterns else self.PATTERNS
+        context = context if context else self.CONTEXT
+        super().__init__(
+            supported_entity=supported_entity,
+            patterns=patterns,
+            context=context,
+            supported_language=supported_language,
+            name=name,
+        )
+        self.score_thresholds = (
```
omri374: Same here. It's better to avoid defaults here and let the user decide

omri374 (commented):
Apologies for the delay. There's one comment from the previous round they should be addressed. Thanks!

bhargavikalicheti:
@omri374 Thnx. I removed the recognizer-level default score threshold and the corresponding YAML threshold. The recognizer now leaves score_thresholds empty, allowing the analyzer or request-level threshold to control filtering. The tests now pass caller-controlled thresholds explicitly.

omri374 (approved):
Thank you! Approved

PR #2168, opened 2026-07-16
fix(analyzer): skip SpacyRecognizer for SlimSpacyNlpEngine

ultramancode:
## Change Description

`SlimSpacyNlpEngine` disables NER, but the recognizer registry treated it like a regular spaCy engine and added `SpacyRecognizer`.

When `SpacyRecognizer` received an empty supported-entities list from the slim engine, it fell back to its default entities. As a result, the registry reported NER entities that the slim engine could not produce.

When `SlimSpacyNlpEngine` is used, the registry no longer adds `SpacyRecognizer` automatically.
If `SpacyRecognizer` is explicitly configured, analyzer initialization fails because the slim engine does not produce the NER results it requires.

This does not affect tokenization, lemmatization, or context-based score enhancement provided by the slim engine.

## Checklist


presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry.py
```diff
@@ -95,9 +102,13 @@ def add_nlp_recognizer(self, nlp_engine: NlpEngine) -> None:
         :return: None
         """
 
-        if isinstance(nlp_engine, NoOpNlpEngine):
+        if isinstance(nlp_engine, _NLP_ENGINES_WITHOUT_NER_OUTPUT):
             self.validate_nlp_engine_compatibility(nlp_engine)
-            logger.info("Skipping NLP recognizer registration for no-op NLP engine.")
+            logger.info(
```
omri374: Could be a debug message

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry.py
```diff
@@ -27,6 +28,11 @@
 
 logger = logging.getLogger("presidio-analyzer")
 
+_NLP_ENGINES_WITHOUT_NER_OUTPUT = (
```
omri374: This list is a good start, but it's (a) hard to maintain (we might forget to update it) and (b) hard to customize for a user downloading from pypi and adding a custom Nlp Engine. 
Here's my suggestion. In `NlpEngine`, add a `has_ner` property
```py
class NlpEngine(ABC):
    ...
    @property
    def has_ner(self) -> bool:
        """Returns True if this engine performs Named Entity Recognition natively."""
        return False
```

And use this to filter out the `SpacyRecognizer`

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry.py
ultramancode: Thanks, changed this to a debug message

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry.py
ultramancode: Thanks for the suggestion. I added the `has_ner` property as proposed and updated the registry to use it when deciding whether to register an NLP recognizer

ultramancode:
Thanks for the review! I've updated the PR based on your suggestions. The capability-based approach made the implementation cleaner and easier to maintain.

presidio-analyzer/presidio_analyzer/nlp_engine/nlp_engine.py
```diff
@@ -12,6 +12,11 @@ class NlpEngine(ABC):
     on tokens.
     """
 
+    @property
+    def has_ner(self) -> bool:
+        """Returns True if this engine performs Named Entity Recognition natively."""
+        return False
```
omri374: This is a breaking change. If someone has a custom nlp engine, it might stop working if this property is set to False. Better to keep it always True and False for the no-op and slim engines

ultramancode: Thanks! has_ner now defaults to True, while SlimSpacyNlpEngine and NoOpNlpEngine explicitly return False. Tests and documentation have been updated accordingly.

omri374 (dismissed):

omri374 (approved):
LGTM! Thanks!

ultramancode:
Thanks for the review and approval!

PR #1814, opened 2025-12-12
feat: Add Korean passport number recognizer (KR_PASSPORT)

kyoungbinkim:
### Add Korean Passport Number Recognition

Adds Korean passport number recognition capability to Presidio Analyzer, supporting both current and previous passport formats used in South Korea.

#### Overview

This PR implements a new recognizer for South Korean passport numbers, which follow specific patterns defined by the Korean Ministry of Foreign Affairs.

**Reference**: [Microsoft Purview - South Korea Passport Number](https://learn.microsoft.com/en-us/purview/sit-defn-south-korea-passport-number)

##### Pattern Recognition

**Current Format **
- Pattern: `[MmSsRrOoDd]\d{3}[A-Za-z]\d{4}`
- Example: `M123A4567`
- Structure:
  - Type prefix (M/S/R/O/D)
    - M: Regular passport (일반여권)
    - S: Service passport (관용여권)
    - O: Official passport (공무여권)
  - 3 digits
  - 1 alphabetic character 
  - 4 digits
- Confidence score: 0.1

**Previous Format **
- Pattern: `[MmSsRrOoDd]\d{8}`
- Example: `M12345678`
- Structure:
  - Type prefix (M/S/R/O/D)
  - 8 digits
- Confidence score: 0.05

##### Regular Expression Details

```regex
# Current Format
(?<![A-Z0-9a-z])[MmSsRrOoDd]\d{3}[A-Za-z]\d{4}(?![0-9])
```
- `(?<![A-Z0-9a-z])`: Negative lookbehind - ensures the pattern doesn't start mid-alphanumeric sequence
- `[MmSsRrOoDd]`: Matches passport type prefix (case-insensitive)
- `\d{3}`: Exactly 3 digits
- `[A-Za-z]`: Single alphabetic character (checksum)
- `\d{4}`: Exactly 4 digits
- `(?![0-9])`: Negative lookahead - ensures the pattern doesn't end mid-number sequence

```regex
# Previous Format
(?<![A-Z0-9a-z])[MmSsRrOoDd]\d{8}(?![0-9])
```
- Similar structure but with 8 consecutive digits and no checksum letter

##### Context Keywords

Supports context-aware detection with keywords in both English and Korean:

**English**: `passport`, `passport#`, `passport #`, `passport no`, `passport number`

**Korean**: `여권`, `여권번호`, `여권 번호`

#### Files Added

- `kr_passport_recognizer.py`: Implements `KrPassportRecognizer` class extending `PatternRecognizer`
[...]

PR #2170, opened 2026-07-18
fix(analyzer): make every recognizer listed in default_recognizers.yaml loadable

### LINKED ISSUE #2176: Recognizers listed in default_recognizers.yaml raise TypeError when enabled: constructors missing the `name` kwarg
## Describe the bug

`default_recognizers.yaml` ships ~60 recognizers with `enabled: false`. That switch exists to be flipped. Three of them cannot be turned on: `KrBrnRecognizer`, `KrDriverLicenseRecognizer` and `UsMbiRecognizer` raise

```
TypeError: __init__() got an unexpected keyword argument 'name'
```

before the registry finishes loading.

`RecognizerListLoader.get` instantiates every predefined recognizer with a `name` keyword argument, taken from the YAML entry (or from `name` when `class_name` supplies the class — the documented rename path, see `RecognizerListLoader.get_recognizer_name`). A recognizer whose `__init__` does not accept `name` therefore cannot be built from a registry configuration at all.

A fourth recognizer, `KrPassportRecognizer`, has the same constructor defect and is consequently **absent from `default_recognizers.yaml` entirely** — it is the only Korean recognizer missing from it, because adding an entry would have made the registry raise. It also defaults to `supported_language="kr"` while every other Korean recognizer defaults to `ko` (the ISO 639-1 code). #1742 migrated the Korean recognizers from `kr` to `ko`; #1814 added this one afterwards and reintroduced `kr`.

Measured across the whole shipped configuration — every entry forced to `enabled: true`, then loaded:

```
yaml entries:                  86
shipped with enabled: false:   60
load fine when enabled:        81
FAIL when enabled:              5
```

Of the five, two are expected: `HuggingFaceNerRecognizer` and `BasicLangExtractRecognizer` refuse to load with an **actionable** `ImportError` naming the optional dependency to install. The other three are the `TypeError` above, which reads as a user configuration error even though the YAML is correct.

> **Correction to the original report.** This issue previously stated that "`KrRrnRecognizer`, `KrBrnRecognizer`, `KrFrnRecognizer` and `KrDriverLicenseRecognizer` all already accept `name`." That is wrong for two of them. Measured:
>
> ```
> KrRrnRecognizer              accepts name=True
> KrFrnRecognizer              accepts name=True
> KrBrnRecognizer              accepts name=False
> KrDriverLicenseRecognizer    accepts name=False
> KrPassportRecognizer         accepts name=False
> UsMbiRecognizer              accepts name=False
> ```
>
> The scope is therefore wider than "one missing Korean recognizer", and the entries that are listed-but-unenableable matter more to users than the one that is merely unlisted.

## To Reproduce

Flip any of the three to `enabled: true` in a registry configuration and load it:

```python
from presidio_analyzer.recognizer_registry import RecognizerRegistryProvider

conf = {
    "global_regex_flags": 26,
    "supported_languages": ["en", "ko"],
    "recognizers": [
        {
            "name": "KrBrnRecognizer",
            "supported_languages": ["ko", "kr"],
            "type": "predefined",
            "enabled": True,
            "country_code": "kr",
        }
    ],
}
RecognizerRegistryProvider(registry_configuration=conf).create_recognizer_registry()
# TypeError: KrBrnRecognizer.__init__() got an unexpected keyword argument 'name'
```

For `KrPassportRecognizer`, add an entry to `default_recognizers.yaml` copied from its siblings — the same `TypeError` occurs. Instantiating it directly with defaults and analysing with `language="ko"` silently skips it, because it registered under `kr`.

## Expected behavior

- Every recognizer listed in `default_recognizers.yaml` loads when `enabled` is true, or fails with an actionable message naming a missing optional dependency.
- `KrPassportRecognizer` is registered in `default_recognizers.yaml` (`enabled: false`, `country_code: kr`) like its siblings, and defaults to `ko`.

## Why this was not caught

The constructor signature is part of a contract that nothing enforces, and the existing tests miss it from both sides:

- Each recognizer's own unit tests instantiate the class **directly**, where no `name` is passed. They all pass.
- The registry-level contract tests — `test_predefined_pattern_recognizers_have_the_right_regex_flags`, `test_default_recognizers_yaml_country_code_matches_class` — build the **default** configuration, in which those ~60 entries are disabled and therefore never constructed.

Nothing in between ever constructed them. Note the `kr` → `ko` half only affects direct instantiation: when a YAML entry omits `supported_languages` the loader passes the registry's language explicitly, so the class default never applies on the YAML path.

## Additional context

Fixed by #2170, which adds the `name` argument to all four constructors, registers `KrPassportRecognizer`, and adds contract tests that construct every YAML entry so the next violation is caught by CI.

developer0hye:
## Change Description

`RecognizerListLoader.get` instantiates every predefined recognizer with a `name` keyword argument, taken from the YAML entry — or from `name` when `class_name` supplies the class, which is the documented rename path (see `RecognizerListLoader.get_recognizer_name`). A recognizer whose `__init__` does not accept `name` therefore cannot be built from a registry configuration at all. Four recognizers were in that state.

**Three are listed in `default_recognizers.yaml` with `enabled: false` and cannot be turned on.** `KrBrnRecognizer`, `KrDriverLicenseRecognizer` and `UsMbiRecognizer` raise `TypeError: __init__() got an unexpected keyword argument 'name'` before the registry finishes loading. `enabled: false` is an opt-in switch, not a disclaimer — an entry that cannot be enabled should not be listed. The failure also reads as a user configuration error even though the YAML is correct, unlike the optional-dependency entries, which refuse to load with an actionable `ImportError` naming what to install.

**One is absent from the YAML because of the same defect.** `KrPassportRecognizer` was the only Korean recognizer missing from `default_recognizers.yaml`; listing it would have made the registry raise. It also defaulted to `supported_language="kr"` while every other Korean recognizer defaults to `ko` (ISO 639-1). #1742 migrated the Korean recognizers from `kr` to `ko`; #1814 added this one afterwards and reintroduced `kr`, so an `AnalyzerEngine` running `ko` silently skipped it. That half only affects direct instantiation — when an entry omits `supported_languages` the loader passes the registry's language explicitly, so the class default never applies on the YAML path.

Measured across the whole shipped configuration, every entry forced to `enabled: true` and loaded:

```
yaml entries:                  86
shipped with enabled: false:   60
load fine when enabled:        81
FAIL when enabled:              5
```

Two of the five are expected (`HuggingFaceNerRecognizer`, `BasicLangExtractRecognizer` — actionable `ImportError` for an optional dependency). The other three are the `TypeError`.

## Changes

- Added the `name` argument, forwarded to `super().__init__`, on `KrPassportRecognizer`, `KrBrnRecognizer`, `KrDriverLicenseRecognizer` and `UsMbiRecognizer`. Appended last in each signature, so existing positional callers are unaffected.
- Changed `KrPassportRecognizer`'s default `supported_language` to `ko`.
- Registered `KrPassportRecognizer` in `default_recognizers.yaml` with `enabled: false` and `country_code: kr`, matching its siblings.
- Added `tests/test_predefined_recognizer_contract.py`.

## Why a contract test, and why the existing ones missed this

The constructor signature is part of a contract nothing enforced, and the existing tests miss it from both sides:

- Each recognizer's own unit tests instantiate the class **directly**, where no `name` is passed. They all pass.
- The registry-level contract tests — `test_predefined_pattern_recognizers_have_the_right_regex_flags`, `test_default_recognizers_yaml_country_code_matches_class` — build the **default** configuration, in which ~60 entries are disabled and therefore never constructed.

The new tests close the gap from both directions:

- **Signature.** Every predefined `PatternRecognizer` subclass must accept the kwargs the loader passes. This fires when a class is added, before it reaches the YAML at all — the point at which `KrPassportRecognizer` went wrong in #1814.
- **Load.** Every entry in `default_recognizers.yaml` must resolve to a class and must load once enabled, exercised entry by entry so a failure names the recognizer. Entries gated behind an optional dependency are skipped, since refusing to load with an actionable `ImportError` is their intended behavior.
- **Intent.** A `class_name` + `name` entry must produce an instance with the configured name. This is the documented reason the loader passes `name`, so it records why the kwarg contract is load-bearing rather than incidental.

### Scope of the load test

The load test covers non-pattern entries too, deliberately. Narrowing it to `PatternRecognizer` subclasses would silently drop `PhoneRecognizer`, `ZaMobileNumberRecognizer` and `ZaTelephoneNumberRecognizer`, which are not `PatternRecognizer`s and do load.

One entry is excluded by name, with its reason recorded next to the exclusion. A further test asserts every exclusion still matches a shipped entry, so the list cannot rot if an entry is renamed, removed, or fixed:

- **`HuggingFaceNerRecognizer`** cannot load from its shipped entry even with its dependencies installed. `EntityRecognizer.__init__` calls `load()` unconditionally, `load()` requires `model_name`, and the entry does not supply one — so with `transformers` and `torch` present it raises `ValueError: model_name must be set before calling load()`, not the `ImportError` it raises when they are absent. Supplying `model_name` in the YAML would make the test download a model. The entry stays covered by the resolve test. **This is a pre-existing defect in the entry** (`HuggingFaceNerRecognizer` is not usable from the shipped configuration on `main` either); worth a separate issue.

`BasicLangExtractRecognizer` is skipped only on `ImportError`, since refusing to load without the `langextract` extra is that recognizer's intended behavior. The skip is scoped to that name rather than to the exception type, so an `ImportError` from any other entry is a failure instead of a green skip.

That recognizer also carries a `config_path` that it resolves against the working directory, so it raises `FileNotFoundError` when pytest runs from the repository root. The load test sets the working directory to the component root — the same one CI uses — instead of catching that error, because catching `FileNotFoundError` would turn a deleted or renamed shipped config file into a passing skip. A separate test asserts every `config_path` in the shipped configuration resolves to a file that exists, which holds even in an environment that cannot construct the recognizer at all. (The more robust YAML value would be `config_path: langextract_config_basic.yaml`, which `resolve_config_path` finds from any cwd — out of scope here.)

The signature test is scoped to `PatternRecognizer` subclasses. `AzureAILanguageRecognizer` is the one remaining class that does not accept `name`; it is absent from `default_recognizers.yaml` and fixes its own display name, so it is out of scope here — but it *is* reachable from a user config by class name and fails the same way. Pre-existing; worth tracking separately.

## Behavior change

The `kr` → `ko` default is a **breaking change for direct instantiation**. `KrPassportRecognizer()` now registers under `ko`, so a caller analysing with `language="kr"` no longer gets `KR_PASSPORT` results. The failure is **silent**, not an exception: the registry raises only when the whole result set is empty, and the per-entity path just logs `Entity KR_PASSPORT doesn't have the corresponding recognizer in language : kr`. Migration is one line — pass `supported_language="kr"`, or switch the call to `language="ko"`.

Registry/YAML users are unaffected: the loader always passes `supported_language` explicitly, and the new entry declares both `ko` and `kr`. No call site in this repository passes `language="kr"`, and no shipped NLP configuration declares `kr`.

## Verification

- Reverting the three constructors makes the signature test and the load test each fail for exactly those three, with no other failures — the tests catch the defect they were written for.
- Full `presidio-analyzer` suite run on this branch and on an `origin/main` worktree: the failure sets are **identical** (42 pre-existing failures from optional dependencies absent in the local environment — `transformers`, `langextract`, some spaCy models). No regressions.
- The new module was also run with the `transformers`/`torch` module globals patched to non-`None`, which reproduces a full-extras CI environment without installing them: `258 passed, 1 skipped`, no failures. An earlier revision of this branch failed there, which is what the `HuggingFaceNerRecognizer` exclusion above fixes.
- Injecting an `ImportError` into a non-excluded entry's constructor makes the load test fail rather than skip, confirming the narrowed exception scope.
- Pointing a `config_path` at a name that does not exist makes the new config-path test fail, confirming a deleted or renamed shipped file is caught rather than skipped.
- The module passes identically when run from `presidio-analyzer` and from the repository root (`258 passed, 1 skipped` both ways), so it is not working-directory dependent.
- `ruff check` from the repo root (as CI runs it), `ruff format --check` on the modified files, and `git diff --check` all pass.
- The `default_recognizers.yaml` diff is an 8-line pure addition.

Per the changelog policy in #2200, `CHANGELOG.md` is not modified — the earlier revision of this branch edited it, which is what made the PR conflict.

## Issue reference

Fixes #2176. Follows up on #1742 (Korean `kr` → `ko` migration) and #1814 (which added `KrPassportRecognizer`).

## Checklist


presidio-analyzer/tests/test_kr_passport_recognizer.py
```diff
@@ -69,7 +79,7 @@ def test_when_all_passports_then_succeed(
     assert len(results) == expected_len
     for res, (st_pos, fn_pos), (st_score, fn_score) in zip(
         results, expected_positions, expected_score_ranges
-    ):  
+    ):
         print(f"res: {res}, st_pos: {st_pos}, fn_pos: {fn_pos}, st_score: {st_score}, fn_score: {fn_score}")
         if fn_score == "max":
```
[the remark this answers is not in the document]
developer0hye: Removed in 8c22f473, with one correction: this `print` is not added by this PR. It is on `main` at the same line, and the only change here was stripping trailing whitespace from the `):` above it, which pulled it into the diff as context. Worth removing regardless since it sits next to code this PR touches.

presidio-analyzer/tests/test_kr_passport_recognizer.py
```diff
@@ -104,7 +114,53 @@ def test_when_no_passport_then_no_results(recognizer, entities):
...
+    entry = copy.deepcopy(entry)
+    entry["enabled"] = True
+    tmp = Path(tempfile.mkdtemp()) / "conf.yaml"
+    tmp.write_text(
+        yaml.safe_dump({"supported_languages": [language], "recognizers": [entry]})
+    )
+    provider = RecognizerRegistryProvider(conf_file=str(tmp))
+    registry = provider.create_recognizer_registry()
+    entities = {e for rec in registry.recognizers for e in rec.supported_entities}
+    assert "KR_PASSPORT" in entities
```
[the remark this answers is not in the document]
developer0hye: Fixed in 8c22f473 by removing the temporary file rather than wrapping it. `RecognizerRegistryProvider` accepts the parsed mapping directly via `registry_configuration`, which is what `test_predefined_recognizer_contract.py` in this PR already does, so the round trip through the filesystem was never needed and there is nothing left to clean up.

While there, the assertion was also strengthened. It only checked that `KR_PASSPORT` was among the loaded entities; it now asserts the loaded recognizer class. The loader drops a recognizer whose language the registry does not support with a log warning and no exception, so a non-empty registry does not on its own prove the entry under test is what loaded.

presidio-analyzer/tests/test_predefined_recognizer_contract.py
```diff
@@ -0,0 +1,261 @@
+"""Contract tests between the registry loader and predefined recognizers.
```
omri374: The repo's tests are arranged based on the files they test. Could we move the tests here to an existing test file like those which test the recognizer registry provider or recognizer list loader?

omri374 (commented):
Thanks! Looks great, left a minor comment on tests.

presidio-analyzer/tests/test_predefined_recognizer_contract.py
developer0hye: Done in dc9b3ff9 — folded into `tests/test_recognizers_loader_utils.py` and the standalone file is gone.

That file over `test_recognizer_registry_provider.py` because `recognizers_loader_utils.py` is the module whose contract these assert: `RecognizerListLoader` is what passes `name`/`supported_language` to every predefined recognizer, what resolves an entry's class via `get_existing_recognizer_cls`, and what consumes `default_recognizers.yaml`. It also already hosts the sibling shipped-YAML check, `test_default_recognizers_yaml_country_code_matches_class`, so the two now sit together rather than in separate files. The two tests that go through `RecognizerRegistryProvider` do so as the public entry point that runs the loader end to end, which is how `test_supported_countries_via_loader_kwarg` in the same file already works — the load test now says so in its docstring, since driving the provider also covers the configuration validation a real user's entry passes through.

The tests are otherwise unchanged; the module docstring became a section header. One small thing while merging: `test_default_recognizers_yaml_country_code_matches_class` now reads the shipped YAML through the module-level constant the move introduced, instead of re-opening the same file a second time in the same module.

Test count is unchanged at 295 — 294 pass, 1 skip (`BasicLangExtractRecognizer`, optional dependency absent). The two `HuggingFaceNerRecognizer` chunker failures in `test_recognizer_registry_provider.py` are pre-existing locally and reproduce on this branch without the change; they need `transformers`.

juno-junho:
I filed #2217 / #2218 for the same `name`-kwarg crash before finding this PR, sorry for the duplicate. I've rescoped my PR to only the parts yours doesn't cover (three other recognizers missing from the yaml), so the two are complementary now.

One small suggestion while I was in this area: a signature-level regression test keeps future yaml entries honest without instantiating anything (instantiation sweeps don't work here because ML entries like `HuggingFaceNerRecognizer` legitimately require configuration before they can be built). Something like:

```python
def test_when_default_yaml_lists_a_predefined_recognizer_then_its_constructor_accepts_name():
    conf = yaml.safe_load(DEFAULT_RECOGNIZERS_YAML.read_text())
    rejecting = []
    for recognizer_conf in conf["recognizers"]:
        if recognizer_conf.get("type") != "predefined":
            continue
        cls = getattr(predefined, recognizer_conf.get("class_name", recognizer_conf["name"]))
        params = inspect.signature(cls.__init__).parameters
        accepts_kwargs = any(p.kind is p.VAR_KEYWORD for p in params.values())
        if "name" not in params and not accepts_kwargs:
            rejecting.append(recognizer_conf["name"])
    assert not rejecting
```

Before your fix it names exactly the four offenders. Feel free to take it or ignore it, happy either way. Thanks for fixing this.

developer0hye:
No apology needed — and thanks for rescoping #2218 rather than dropping it. `AbaRoutingRecognizer`, `FiPersonalIdentityCodeRecognizer` and `SgUenRecognizer` are outside this PR, they're all `PatternRecognizer`s that already accept `name`, and the two branches only meet on one `default_recognizers.yaml` hunk. Once both land, the load test here parametrizes over your three new entries automatically, so they get the enable-safe check without you wiring anything up.

Took the suggestion, in `7304d492` — but narrowed to the part that was actually uncovered, because measuring it turned up less overlap than either of us assumed in one direction and more in the other.

**Where you were right.** My signature sweep was package-driven *and* scoped to `PatternRecognizer` subclasses, which left five shipped entries unreached: `PhoneRecognizer`, the two `Za*` ones, and the NER/LLM wrappers. Three of those the load test constructs outright, so they were covered more strongly than a signature check anyway. But `HuggingFaceNerRecognizer` is excluded from the load test — its shipped entry supplies no `model_name`, so `EntityRecognizer.__init__` calling `load()` raises `ValueError` once `transformers` is present — which means **nothing in the suite touched its constructor at all**. That's the hole, and it's the one your framing points straight at: a check that builds nothing keeps working where a sweep that builds things has to give up.

So the test now parametrizes over the union of both sources. Neither subsumes the other, which is the part worth recording:

- **package-driven** reaches a class *before* it is listed anywhere;
- **yaml-driven** reaches a listed class the package sweep skips by base class.

**One correction, and it's the reason the sweep starts from the package.** A yaml-driven check names *three* offenders on `main`, not four. `KrPassportRecognizer` isn't in `main`'s `default_recognizers.yaml` — I checked — and that isn't incidental: it's absent *because* its constructor rejected `name`, so listing it would have made the registry raise. The class that most needed catching was the one structurally invisible to a yaml-driven test. That's the #1814 failure mode, and it's why the package sweep stays.

**And one thing that didn't work out the way the snippet implies.** `HuggingFaceNerRecognizer.__init__` takes `**kwargs`, so `accepts_kwargs` exempts it — your version passes it without asserting anything, and so does mine. No signature test can do better here; a `**kwargs` constructor simply doesn't say which kwargs it honors. Mine reports it as a named skip rather than a silent pass, so the gap stays visible in the run:

```
SKIPPED HuggingFaceNerRecognizer.__init__ takes **kwargs, so the signature cannot
        show which kwargs it honors. Reported as a skip rather than a silent pass
        so the gap in coverage stays visible.
```

Net effect: 5 new params — 3 assertions, 2 skips (`HuggingFaceNerRecognizer`, `BasicLangExtractRecognizer`, both `**kwargs`). 297 passed, 3 skipped, up from 294 / 1. Renamed to `test_recognizer_accepts_loader_kwargs` since it's no longer pattern-only.

`HuggingFaceNerRecognizer` being unusable from its shipped entry is a real pre-existing defect, separate from this PR and from yours — it's recorded in `NOT_LOADABLE_FROM_SHIPPED_ENTRY` with a guard test asserting the exclusion still names a shipped entry, so it can't rot into silently narrowed coverage. Worth its own issue; say the word if you'd rather file it, since you found the same edge from the other side.

juno-junho:
The 3-vs-4 mismatch is on me: I counted on my own branch, which already had the KrPassport yaml entry added, so the yaml-driven check saw four there. On `main` it's three, as you said. And your point lands: the class that most needed catching is exactly the one a yaml-driven check is structurally blind to, which argues for the package sweep better than anything in my snippet.

The union parametrization plus the named skip is strictly better than what I posted. Thanks for measuring it instead of taking it as-is; the `**kwargs` limitation in particular is something I hadn't thought through.

I'll file the HuggingFaceNerRecognizer issue. I hit the same edge from the other direction (an enable-everything sweep while benchmarking #2216) and have the repro handy, so it's a natural fit. I'll reference your `NOT_LOADABLE_FROM_SHIPPED_ENTRY` guard so the two ends stay linked.

developer0hye:
Thanks — and the miscount was a useful one to have made. Counting on a branch that already had the entry is exactly the situation the test will be in from here on, so it surfaced the asymmetry faster than counting on `main` would have.

Glad the issue is going to you. A repro from an enable-everything sweep is better evidence than the exclusion constant on its own, since it shows the entry failing the way a user would hit it rather than the way a test avoids it. I'll link the issue from `NOT_LOADABLE_FROM_SHIPPED_ENTRY` once it's up, so the constant points at the reason it exists instead of just naming the class.

Looking forward to #2218 landing.

juno-junho:
It's up: #2222. Thanks again.

presidio-analyzer/tests/test_recognizers_loader_utils.py
```diff
@@ -571,3 +584,295 @@ def test_yaml_country_code_blank_value_raises():
...
+# ---------------------------------------------------------------------------
+
+# Kwargs ``RecognizerListLoader`` passes to every predefined recognizer it
+# builds: ``name`` comes from the YAML entry (or its ``class_name`` alias) and
+# ``supported_language`` from the resolved per-language configuration.
+LOADER_KWARGS = ("name", "supported_language")
+
+# Entries that cannot load from their shipped configuration even with every
+# dependency installed, so the load test below cannot cover them.
+#
```
omri374, line 611: How about creating an issue to fix this? We can keep it as is in this PR but it's better to fix than to have lists of outliers

omri374 (approved):

presidio-analyzer/tests/test_recognizers_loader_utils.py
omri374: Oh I see you did! Great

PR #2179, opened 2026-07-21
fix(anonymizer): include confidence score in OperatorResult

### LINKED ISSUE #2057: Feature Request: Include confidence score in AnonymizerEngine output (OperatorResult) for audit/traceability

When building audit logging or compliance pipelines on top of Presidio, `AnonymizerEngine` currently drops the confidence score from each `RecognizerResult` after anonymization. The resulting `OperatorResult` only carries `start`, `end`, `entity_type`, `text`, and `operator` - the score is silently lost.

This makes it impossible to answer post-anonymization questions like:
- "With what confidence was this entity anonymized?"
- "Were any entities anonymized below our confidence threshold?"
- "Which documents need human review based on low-confidence detections?"

without re-running the analyzer on the original text, which may no longer be available or desirable.

Add an optional `score` field to `OperatorResult`, threaded through from the originating `RecognizerResult` during anonymization:

The score should:
- Default to `None` so existing usage and the deanonymize path are completely unaffected
- Be included in JSON serialization
- Be preserved through conflict resolution (the surviving entity keeps its own score, not that of any dropped entity)
- Be deserializable via `.from_json()`

Zipping `analyzer_results` with `anonymized.items` by position - but this breaks because:
1.  Text positions shift in the anonymized output
2.  Conflict resolution drops some analyzer results, so the list lengths doesn't match

There is no reliable way to recover the score post-anonymization without carrying it through the engine.

**Additional context**
This is a purely additive change - `score: Optional[float] = None` requires no breaking changes to existing code. The score is already present on `RecognizerResult`, it just needs to be passed when constructing `OperatorResult`.

kalra-mohit:
## Summary
Adds an optional `score` field to `OperatorResult`, threaded through from the originating `RecognizerResult` during `AnonymizerEngine.anonymize()` and preserved through `DeanonymizeEngine.deanonymize()`.

- `score` defaults to `None` so existing callers are unaffected
- The surviving entity after conflict resolution keeps its own (max) score, not a dropped entity's — this relies on existing conflict-resolution behavior in `anonymizer_engine.py`, unmodified here
- Included in `to_dict()`/`to_json()` output and round-trips via `from_json()`

Enables audit/traceability use cases where the confidence of an anonymized entity needs to be recoverable after anonymization.

Closes #2057

## Test plan
- [x] Unit tests for `OperatorResult` score default, equality, `to_dict`/`from_json`
- [x] Unit tests for score propagation through `AnonymizerEngine` (single, multiple, conflicting entities)
- [x] Integration tests updated for new `score` field in output
- [x] Round-trip test: encrypt → decrypt preserves score

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ All notable changes to this project will be documented in this file.
 - `BasicLangExtractRecognizer` now honours values under `langextract.model.provider.language_model_params` (including `timeout` and `num_ctx`). Previously these were silently dropped because `langextract.extract()` ignores its `language_model_params` argument when a pre-built `ModelConfig` is passed via `config=`, causing Ollama-backed recognizers to fall back to langextract's 120s default regardless of the configured timeout. The recognizer now merges `language_model_params` into `ModelConfig.provider_kwargs`, which is the path that reaches the provider constructor. Explicit entries under `provider.kwargs:` still take precedence. Also fixed a `TypeError` when `kwargs:` or `language_model_params:` is `null` in the YAML. (#1943, Thanks @lsternlicht)
 
 ### Anonymizer
+#### Changed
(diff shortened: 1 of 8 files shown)


kalra-mohit:
Closing — I found that #2059 already implements this same change (optional \`score\` field threaded through \`OperatorResult\`) and was already approved on 2026-07-13, pending test fixes. Didn't want to duplicate review effort. Thanks @kshvg for getting there first — nice implementation.

PR #2180, opened 2026-07-21
docs(analyzer): document and test configuring multiple GLiNERRecognizer instances via YAML

kalra-mohit:
## Summary
Configuring multiple `GLiNERRecognizer` instances side-by-side via the recognizer registry YAML (e.g. different models/thresholds/entity mappings) already works today by giving each entry a distinct `name` while pointing `class_name` at `GLiNERRecognizer` — the same mechanism used for `HuggingFaceNerRecognizer`. It just wasn't documented, so users had no way to discover it.

This PR:
- Adds a "Configuring multiple GLiNER recognizers via YAML" section with a worked example to the GLiNER sample doc
- Clarifies the `class_name` field in `recognizer_registry_provider.md`
- Adds regression tests (with `GLiNER` mocked, so no heavy optional dependency required) covering both the registry-provider path and the lower-level loader utils
- Notes a known related limitation (entity leakage across instances when `entity_mapping`s differ — see #1760's later comment thread and the pending fix in #2154) so the doc doesn't imply stronger isolation than currently exists

No source code changes — this is docs + tests only.

Closes #1760

## Loading flow

```mermaid
flowchart TD
    A["recognizer_registry.yaml"] --> B1["entry: name=GLiNERRecognizerMultiPII<br/>class_name=GLiNERRecognizer<br/>threshold=0.4"]
    A --> B2["entry: name=GLiNERRecognizerSmall<br/>class_name=GLiNERRecognizer<br/>threshold=0.25"]
    B1 --> C1["GLiNERRecognizer instance A<br/>gliner_multi_pii-v1"]
    B2 --> C2["GLiNERRecognizer instance B<br/>gliner_small-v2.5"]
    C1 --> D["RecognizerRegistry"]
    C2 --> D
    D --> E["AnalyzerEngine.analyze"]
    E -. "known gap #1760 / fix pending #2154:<br/>requested entities appended as labels to ALL instances" .-> C1
    E -. same gap .-> C2
```

## Test plan
- [x] New regression test in `test_recognizer_registry_provider.py`
- [x] New regression test in `test_recognizers_loader_utils.py`
- [x] Verified `class_name` resolution logic predates this change (not new behavior, just newly documented)
- [x] Verified the entity-leakage caveat against the live issue thread and the open fix PR (#2154) before documenting it

PR #2018, opened 2026-05-03
Support multiple GLiNER YAML configurations

ynachiket:
## Change Description

Support configuring multiple `GLiNERRecognizer` instances in YAML by allowing `class_name: GLiNERRecognizer` entries without explicit `name` values and deriving deterministic recognizer names from `model_name` when `name` is omitted.

This preserves explicit `name` values unchanged and keeps the legacy `GLiNERRecognizer` name for the built-in default model.

## Issue reference

Addresses #1760.

## Validation

Ran from `presidio-analyzer` in a local Python 3.12 venv:

```bash
pytest tests/test_gliner_recognizer.py -q
pytest tests/test_recognizers_loader_utils.py -q
pytest tests/test_yaml_recognizer_models.py -q
pytest tests/test_recognizer_registry_provider.py -q
git diff --check
```

Results:
- `tests/test_gliner_recognizer.py`: 3 passed, 12 skipped
- `tests/test_recognizers_loader_utils.py`: 11 passed
- `tests/test_yaml_recognizer_models.py`: 47 passed
- `tests/test_recognizer_registry_provider.py`: 17 passed
- `git diff --check`: passed

Also attempted the full analyzer test suite; unrelated optional dependency tests fail locally without Azure/LangExtract extras installed, while the relevant targeted suites pass.

## Checklist

- [x] I have reviewed the contribution guidelines
- [x] My code includes unit tests
- [x] Targeted unit tests and diff checks pass locally
- [ ] Full test suite passes locally

PR #2181, opened 2026-07-21
fix(analyzer): fix GLiNER multi-instance entity leakage and simplify config

### LINKED ISSUE #1760: Can multiple GLiNERRecognizer instances be configured via YAML?

kalra-mohit:
## Summary

Consolidates two other open approaches to this issue into one working, documented, tested fix, crediting both explicitly:

- **Auto-derived names** (approach credited to #2018, @ynachiket): a `class_name: GLiNERRecognizer` YAML entry no longer requires an explicit unique `name` — when omitted, one is derived automatically from `class_name` + `model_name`.
- **Entity/threshold leakage fix** (approach credited to #2154, @ultramancode, reproducing @hexsm's original report): adds an `include_requested_entities_as_labels` option (default `true`, preserving existing behavior). GLiNER's ad-hoc-label support previously let a globally-requested entity type "leak" into a `GLiNERRecognizer` instance that wasn't configured for it, evaluated against the wrong instance's threshold, when multiple instances are configured with different `entity_mapping`/`threshold` values. Setting this to `false` per-instance closes the leak by restricting GLiNER's label input to only that instance's own configured entities.
- Docs updated (`gliner.md`, `recognizer_registry_provider.md`) to document both, replacing an earlier docs-only PR's (#2180, mine) caveat about the leakage bug with the actual fix.

Closes #1760

## Why consolidated rather than three separate PRs

Posted on the issue thread before starting this work: three open PRs (#2018, #2154, #2180) were addressing complementary but separate slices of the same underlying problem. Rather than have maintainers review and merge three overlapping PRs, this folds the auto-naming behavior and the leakage fix together with the docs so a user configuring multiple `GLiNERRecognizer` instances gets working, documented, isolated behavior in one PR. Happy to split this back up, defer to #2018/#2154 directly, or adjust scope however maintainers prefer — flagged this intent on the issue first specifically to get out ahead of that.

## Root cause (verified against current source, not assumed)

In `GLiNERRecognizer.analyze()`, `__create_input_labels()` appends every globally-requested entity type as an ad-hoc GLiNER label by default. Combined with the result filter (`if entities and presidio_entity not in entities: continue`), which checks against the *global* requested-entity list rather than this instance's own `entity_mapping`, an instance configured only for e.g. `ADDRESS` can still return an `ORGANIZATION` result if `ORGANIZATION` was requested elsewhere in the same `analyze()` call — evaluated against the `ADDRESS` instance's threshold, not the (possibly stricter) threshold configured for `ORGANIZATION` elsewhere.

```mermaid
flowchart TD
    A["YAML: 2x class_name: GLiNERRecognizer\nentries, different entity_mapping/threshold"] --> B["Loader: name auto-derived\nfrom class_name + model_name"]
    B --> C1["Instance A: entity_mapping={Address: ADDRESS}\nthreshold=0.65"]
    B --> C2["Instance B: entity_mapping={Organization: ORGANIZATION}\nthreshold=0.80"]
    D["analyze(entities=[ADDRESS, ORGANIZATION, PRODUCT])"] --> C1
    D --> C2
    C1 -. "include_requested_entities_as_labels=True (old default):\nORGANIZATION leaks in as ad-hoc label,\nreturned at Instance A's 0.65 threshold" .-> BUG["Wrong threshold applied"]
    C1 -- "include_requested_entities_as_labels=False:\nonly ever sends [Address] as labels" --> FIXED["Leak closed"]
```

## Test plan

- [x] Regression test reproducing the exact leakage bug from the issue thread, using a mock that faithfully simulates GLiNER (only returns labels it was actually asked for) — this is the meaningful part: a naive mock returning fixed values regardless of input wouldn't actually exercise the fix
- [x] Test confirming `include_requested_entities_as_labels=False` prevents the leak
- [x] Test confirming the flag doesn't suppress the instance's own legitimate entity type
- [x] Unit tests for name auto-derivation: omitted name + model_name → derived name; two instances with different model_names → distinct names; explicit name not overridden; no name/no model_name → falls back to class_name
- [x] End-to-end test through `RecognizerRegistryProvider` with two full YAML-style configs, verifying both instances load with correct distinct names, thresholds, and the leakage flag applied
- [x] Ran the full affected test suite locally (`test_gliner_recognizer.py`, `test_yaml_recognizer_models.py`, `test_recognizer_registry_provider.py`, `test_recognizers_loader_utils.py`) — 152 passed, 0 failed
- [x] `ruff check` clean on both changed source files


RonShakutai:
Please align this PR with the expected behavior discussed in issue #1760, specifically [my July 22 comment](https://github.com/data-privacy-stack/presidio/issues/1760#issuecomment-5043581936):

- When instances use **different models**, automatically derive unique, human-readable recognizer names from `model_name`.
- When multiple instances use **the same model**, require explicit, unique names. Otherwise, fail with a clear validation error rather than accepting duplicate generated names.

Could you update the implementation to follow this behavior, particularly the same-model validation, and confirm here once it is addressed?

PR #2191, opened 2026-07-23
feat(analyzer): opt-in merge of adjacent same-type entities (#1090)

### LINKED ISSUE #1090: Merge two entities from the same type with whitespace between them

watsos:
## Change Description

Adds `EntityRecognizer.merge_adjacent_text_entities(results, text, entity_types)` — merges adjacent same-type spans separated only by whitespace (e.g. spaCy detecting "Dave" and "Jones" as two PERSON spans instead of one).

Wired into `AnalyzerEngine.analyze()` as a new **opt-in** parameter, `merge_adjacent_entities`, defaulting to `None` (off). Callers pass the specific entity types they want merged (e.g. `["PERSON"]`); everything else is untouched.

**Why opt-in and type-scoped:** An earlier attempt at this (#2046) applied merging unconditionally to every recognizer's output, which would silently fuse genuinely distinct adjacent matches for pattern-based recognizers (e.g. two phone numbers or emails on consecutive lines). Scoping to explicit entity types avoids that failure mode while still solving the NER-fragmentation problem.

Details:
- Merged span keeps `max(score_a, score_b)`, and keeps `analysis_explanation`/`recognition_metadata` from whichever span contributed that winning score.
- Requires `nxt.start >= current.end` before checking the gap, so overlapping spans are never merged.
- Input need not be pre-sorted.
- Added tests: merge basic case, metadata/explanation preservation, 3-token chain, different entity types not merged, non-whitespace gap not merged, entity type not in eligible list, default (no `entity_types`) leaves results untouched, overlapping spans not merged, unsorted input, empty input. 31/31 passing.

## Issue reference

Fixes #1090

## Checklist


watsos:
Fixed — merge now returns results sorted by the same (-score, start, -length) key as remove_duplicates. Pushed in 0d87a8f7

omri374 (dismissed):
Thanks! Approved. 

One thought I have: Should the user pass specific entities that want merged or a bool to merge anything that has a whitespace between and is the same entity? In which case woule I want a name to merge but a location not to?

watsos:
Good question — that's exactly why I went with a list of entity types rather than a bool. 

A blanket "merge anything adjacent with whitespace" would fuse things you don't want fused too (e.g. two adjacent LOCATION spans that happen to be different places). Scoping to explicit types let’s a caller say `merge_adjacent_entities=["PERSON"]` to fix NER fragmentation on names while leaving LOCATION (or anything else) untouched. 

 

Happy to adjust the docs/docstring further if the intent isn't clear enough as written.

 

Please let me know your thoughts.

 

Steve

PR #2198, opened 2026-07-27
Pin Pillow exactly at 12.3.0

SharonHart:
## Summary

- exact-pin Pillow 12.3.0 in `presidio-image-redactor/pyproject.toml`
- regenerate `presidio-image-redactor/uv.lock` with uv 0.11.6
- preserve the single resolved Pillow 12.3.0 package and its SHA-256 artifact hashes

## Context

GitHub's Python Dependabot graph job continues to report a stale transitive Pillow 12.2.0 through matplotlib and pytesseract, even though both dependencies allow Pillow 12.3.0, the project requires Pillow 12.3.0 or newer, and the committed uv lock resolves only Pillow 12.3.0.

This deliberately narrows the requirement to `pillow==12.3.0` to force GitHub to re-resolve the `pyproject.toml` manifest without selecting 12.2.0. The uv regeneration only updates the root requirement metadata and normalizes existing environment markers; no package versions change.

The 13 Pillow alert closures must be verified after this PR merges and GitHub rebuilds the dependency graph.

### DIFF
--- a/presidio-image-redactor/pyproject.toml
+++ b/presidio-image-redactor/pyproject.toml
@@ -23,7 +23,7 @@ readme = "README.md"
 requires-python = ">=3.10,<3.15"
 
 dependencies = [
-    "pillow>=12.3.0,<13.0.0",
+    "pillow==12.3.0",
     "pytesseract (>=0.3.13,<0.4)",
     "presidio-analyzer (>=2.2.0,<3.0.0)",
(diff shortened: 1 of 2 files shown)


SharonHart:
Superseded by #2199, which contains the same exact Pillow pin with a strictly minimal two-line diff and clean uv 0.11.6 lock validation.

PR #2209, opened 2026-08-03
Fix UK_NINO regex matching a leading space

inesaranab:
## Change Description

The `UK_NINO` pattern places an optional space between `\b` and the prefix capture group:

```
\b(?!bg|gb|nk|kn|nt|tn|zz|BG|GB|NK|KN|NT|TN|ZZ) ?([a-ceghj-pr-tw-z...]{1}...)
                                               ^^^
```

When a NINO follows a word, `\b` matches after that word and the ` ?` consumes the separating space, so the match begins one character early. This has two consequences.

**1. The reported span includes the leading space, corrupting anonymized output.**

```
>>> AnonymizerEngine().anonymize(text="my AB123456C", analyzer_results=results).text
'my<UK_NINO>'      # expected: 'my <UK_NINO>'
```

**2. The excluded prefixes stop being excluded**, because the negative lookahead is evaluated at the space rather than at the prefix letters.

```
>>> recognizer.analyze("BG123456A", ["UK_NINO"])        # correctly rejected
[]
>>> recognizer.analyze("my BG123456A", ["UK_NINO"])     # matched
[type: UK_NINO, start: 2, end: 12, score: 0.5]
```

`BG`, `GB`, `NK`, `KN`, `NT`, `TN` and `ZZ` are never issued as NINO prefixes
([HMRC NIM39110](https://www.gov.uk/hmrc-internal-manuals/national-insurance-manual/nim39110)),
which is the reason the lookahead exists.

## Fix

A NINO never begins with a space, and the five remaining ` ?` already handle the optional
spaces *within* the number. Remove the leading one:

```diff
-r"\b(?!bg|gb|nk|kn|nt|tn|zz|BG|GB|NK|KN|NT|TN|ZZ) ?([a-ceghj-pr-tw-z...
+r"\b(?!bg|gb|nk|kn|nt|tn|zz|BG|GB|NK|KN|NT|TN|ZZ)([a-ceghj-pr-tw-z...
```

## Tests

Adds parametrized cases covering both symptoms: spans when the NINO follows a word, and
excluded prefixes in that same position.

One existing expectation is **corrected** rather than added. The case
`"Here is my National Insurance Number YZ 61 48 68 B"` asserted `(36, 50)`, but index 36 is
the space and index 37 is the `Y` — it encoded the buggy span. Every other case in the file
places the NINO at position 0 or after punctuation, where the bug cannot fire, which is why
it went unnoticed.

## Checklist

### DIFF
--- a/presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/uk/uk_nino_recognizer.py
+++ b/presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/uk/uk_nino_recognizer.py
@@ -18,7 +18,7 @@ class UkNinoRecognizer(PatternRecognizer):
     PATTERNS = [
         Pattern(
             "NINO (medium)",
-            r"\b(?!bg|gb|nk|kn|nt|tn|zz|BG|GB|NK|KN|NT|TN|ZZ) ?([a-ceghj-pr-tw-zA-CEGHJ-PR-TW-Z]{1}[a-ceghj-npr-tw-zA-CEGHJ-NPR-TW-Z]{1}) ?([0-9]{2}) ?([0-9]{2}) ?([0-9]{2}) ?([a-dA-D]{1})\b",  # noqa: E501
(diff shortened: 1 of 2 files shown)


omri374 (approved):
Thanks!

inesaranab:
@omri374 your thanks made my day!

omri374:
@inesaranab my pleasure! Thanks for your contribution!

PR #2218, opened 2026-08-04
fix(analyzer): register three predefined recognizers missing from default_recognizers.yaml

### LINKED ISSUE #2217: Three pattern recognizers are missing from default_recognizers.yaml (AbaRouting, FiPersonalIdentityCode, SgUen)
> **Scope correction (2026-08-05):** the `name`-kwarg crash half of this report duplicates #2176, which predates it. I missed it when searching before filing, sorry for the noise. That half is deferred to #2176 / #2170. The remaining scope of this issue is the part not covered there: three pattern recognizers missing from the yaml, plus a regression test.

**Describe the bug**

`AbaRoutingRecognizer`, `FiPersonalIdentityCodeRecognizer` and `SgUenRecognizer` are exported in `predefined_recognizers` but have no entry in `default_recognizers.yaml`, so no-code YAML users can't discover or enable them. (A fourth missing one, `KrPassportRecognizer`, is already being added by #2170 together with its constructor fix.)

**To Reproduce**

Search `default_recognizers.yaml` for `AbaRoutingRecognizer`, `FiPersonalIdentityCodeRecognizer` or `SgUenRecognizer`: no entries, while their siblings (e.g. `UsBankRecognizer`, `SgFinRecognizer`) are all listed.

**Expected behavior**

Every exported pattern recognizer is discoverable in the yaml, registered `enabled: false` so default behavior is unchanged.

**Additional context**

All three constructors already accept the loader's `name` kwarg, so the new entries are enable-safe independently of #2170. PR: #2218.

juno-junho:
## Change Description

Register the three pattern-based predefined recognizers that are exported in code but missing from `default_recognizers.yaml`: `AbaRoutingRecognizer`, `FiPersonalIdentityCodeRecognizer` and `SgUenRecognizer`. Each entry ships `enabled: false` with the matching `country_code`, so default behavior is unchanged while no-code YAML users can discover and enable them. A test flips the three entries on and asserts they instantiate (all three constructors already accept the loader's `name` kwarg, so these entries are enable-safe on their own).

Scope note: this PR originally also fixed the `name`-kwarg TypeError on four other recognizers and registered `KrPassportRecognizer`. That overlaps #2176 / #2170, which predate this PR and which I missed when filing (sorry about that). I rescoped this PR to only the parts #2170 doesn't cover, so the two PRs are complementary and can land in either order.

## Issue reference

Fixes #2217

## Checklist

### DIFF
--- a/presidio-analyzer/presidio_analyzer/conf/default_recognizers.yaml
+++ b/presidio-analyzer/presidio_analyzer/conf/default_recognizers.yaml
@@ -84,6 +84,13 @@ recognizers:
     enabled: false
     country_code: us
 
+  - name: AbaRoutingRecognizer
+    supported_languages:
+    - en
+    type: predefined
+    enabled: false
+    country_code: us
+
   - name: NhsRecognizer
     supported_languages: 
     - en
@@ -134,6 +141,13 @@ recognizers:
 
   - name: SgFinRecognizer
     supported_languages: 
+    - en
+    type: predefined
+    enabled: false
+    country_code: sg
+
+  - name: SgUenRecognizer
+    supported_languages:
     - en
     type: predefined
     enabled: false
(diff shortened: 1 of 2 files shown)


omri374 (approved):
Thanks!

Issue #1457, opened 2024-09-30
AzureAILanguageRecognizer not working through the yaml configuration


omri374:
I have an AI application and middleware proxy which makes a call to the Presidio Analyzer and Anonymizer containers with the users' prompt, and expects to receive a callback with the PII (e.g. names replaced by <PERSON>) deleted.

The default SpaCY NLP model misses some less common names, so I want to integrate the commercial AAI service as [described here](https://microsoft.github.io/presidio/samples/python/text_analytics/). In my Docker deploy-compose.override.yml file I have the following services defined:

```
  presidio-anonymizer:
    image: http://mcr.microsoft.com/presidio-anonymizer
    restart: always
    ports:
      - "127.0.0.1:5001:3000"
  presidio-analyzer:
    image: http://mcr.microsoft.com/presidio-analyzer
    restart: always
    ports:
      - "127.0.0.1:5002:3000"
    environment:
      - AZURE_AI_KEY=REDACTED
      - AZURE_AI_ENDPOINT=https://REDACTED.azure.com/
    volumes:
      - ./presidio/default_recognizers.yaml:/usr/bin/presidio-analyzer/presidio_analyzer/conf/default_recognizers.yaml
```

I am attempting to use the predefined recognizer at https://github.com/microsoft/presidio/blob/main/presidio-analyzer/presidio_analyzer/predefined_recognizers/azure_ai_language.py by adding the following lines to the default_recognizers.yaml file which I am bind mounting into the Analyzer container with the following lines added:

```
  - name: AzureAILanguageRecognizer
    type: predefined

```
Unfortunately when I attempt to start up the Analyzer container, it crash loops with this error in the container logs:

```
Traceback (most recent call last):
  File "/usr/bin/presidio-analyzer/app.py", line 138, in <module>
    server = Server()
  File "/usr/bin/presidio-analyzer/app.py", line 44, in __init__
    self.engine: AnalyzerEngine = AnalyzerEngineProvider(
  File "/usr/bin/presidio-analyzer/presidio_analyzer/analyzer_engine_provider.py", line 75, in create_engine
    registry = self._load_recognizer_registry(
  File "/usr/bin/presidio-analyzer/presidio_analyzer/analyzer_engine_provider.py", line 121, in _load_recognizer_registry
    registry = provider.create_recognizer_registry()
  File "/usr/bin/presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry_provider.py", line 59, in create_recognizer_registry
[...]

PR #1819, opened 2025-12-15
Feat: add class_name to allow multiple recognizers from same class

RonShakutai:
## Added `class_name` field to recognizer YAML configuration for separating display names from Python class names

This change introduces an optional `class_name` field that allows custom recognizer instance names while explicitly specifying the Python class to instantiate. This is particularly useful when loading multiple instances of the same recognizer class with different configurations.

### How it works:
- `name`: The display name used in analysis results (e.g., "MyCustomEmailDetector")
- `class_name` (new, optional): The Python class name for instantiation (e.g., "EmailRecognizer")
- When `class_name` is present, it's used for class lookup; otherwise the system falls back to using `name`

### Backward compatibility:
Fully backward compatible - existing configurations without `class_name` continue to work unchanged.

## Change Description

Describe your changes

## Issue reference

Fixes #XX

## Checklist

- [ ] I have reviewed the [contribution guidelines](https://github.com/microsoft/presidio/blob/main/CONTRIBUTING.md)
- [ ] I have signed the CLA (if required)
- [ ] My code includes unit tests
- [ ] All unit tests and lint checks pass locally
- [ ] My PR contains documentation updates / additions if required

PR #1458, opened 2024-09-30
Bug/azure ai language context

omri374:
## Change Description

Azure AI language recognizer is missing a parameter which prevents using it as part of the no-code yaml configuration.

## Issue reference

This PR fixes issue #1457 

## Checklist

- [X] I have reviewed the [contribution guidelines](https://github.com/microsoft/presidio/blob/main/CONTRIBUTING.md)
- [X] I have signed the CLA (if required)
- [X] My code includes unit tests
- [X] All unit tests and lint checks pass locally
- [X] My PR contains documentation updates / additions if required

PR #2224, opened 2026-08-05
fix(analyzer): let AzureAILanguageRecognizer load from a registry config, and un-mute the test that guards it

### LINKED ISSUE #2223: Two tests in test_analyzer_engine_provider.py can never run, hiding a regression in AzureAILanguageRecognizer
## Summary

Two tests in `presidio-analyzer/tests/test_analyzer_engine_provider.py` are gated with

```python
@pytest.mark.skipif(pytest.importorskip("azure"), reason="Optional dependency not installed")
```

which skips them **whether or not the dependency is installed**. One of the two is the regression test added in #1458 for #1457. It has been muted since March 2025, and while muted the bug it guards came back — `AzureAILanguageRecognizer` cannot be constructed from a registry configuration on `main` today.

Three separable defects, in the order they need fixing.

## 1. The skip markers are a double negative

`pytest.importorskip(name)` skips when the module is missing and otherwise **returns the module object**, which is truthy. Wrapping it in `skipif` therefore loses both ways:

| `azure` present? | What happens |
| --- | --- |
| No | `importorskip` raises `Skipped` while the decorator is evaluated, i.e. at module import — **the whole file fails to collect** |
| Yes | returns a module → truthy → `skipif(True)` → **skipped** |

Measured on `main`, `presidio-analyzer`:

```
$ python -m pytest tests --collect-only -q | tail -1
3248 tests collected           # without the azure-ai-language extra
3292 tests collected           # with it

$ python -m pytest tests/test_analyzer_engine_provider.py -rs -q     # extra NOT installed
SKIPPED [1] tests/test_analyzer_engine_provider.py:244: could not import 'azure'
1 skipped                      # exit code 5 - no tests collected

$ python -m pytest tests/test_analyzer_engine_provider.py -rs -q     # extra installed
SKIPPED [1] tests/test_analyzer_engine_provider.py:243: Optional dependency not installed
SKIPPED [1] tests/test_analyzer_engine_provider.py:274: Optional dependency not installed
41 passed, 3 skipped
```

So the impact is not two tests:

- **CI** installs `--all-extras` for the Analyzer component (`.github/workflows/ci.yml:74`), so the two tests skip on every run, reporting `Optional dependency not installed` while it *is* installed. The reason string states the opposite of the truth, which is why this reads as normal in the log.
- **A contributor's environment** without the extra loses **44 tests** — the entire module — at collection time, with a single skip line to show for it.

This is the only place in the repository with this shape. The other 15 `importorskip` call sites invoke it inside a function or fixture, which is the correct call-time usage, and `tests/test_azure_openai_langextract_recognizer.py:9` uses the correct module-level idiom (`pytest.skip(..., allow_module_level=True)`).

## 2. What the muted test was hiding

Calling `test_analyzer_engine_provider_with_azure_ai_language` directly, bypassing the marker:

```
TypeError: AzureAILanguageRecognizer.__init__() got an unexpected keyword argument 'name'
```

Reproduced independently through the loader, with the entry shape from #1457:

```python
RecognizerListLoader.get(
    recognizers=[{"name": "AzureAILanguageRecognizer", "type": "predefined",
                  "supported_languages": ["en"]}],
    supported_languages=["en"], global_regex_flags=26,
)
# TypeError: AzureAILanguageRecognizer.__init__() got an unexpected keyword argument 'name'
```

`RemoteRecognizer.__init__` accepts `supported_entities`, `name`, `supported_language`, `version` and `context`. `AzureAILanguageRecognizer.__init__` accepts none of `name` / `context`, so any registry-configuration path into it raises before the registry finishes loading — the crash-loop symptom originally reported in #1457.

How it came back:

| Date | | |
| --- | --- | --- |
| 2024-10-02 | #1458 | Fixes #1457 by adding `**kwargs`, and adds the regression test |
| 2025-03-10 | #1521 | An unrelated PR adds the `skipif(importorskip(...))` marker — test muted |
| 2025-12-01 | #1800 | Removes `**kwargs` again. CI green, because the test cannot run |
| 2026-01-05 | #1819 | `class_name` support makes the loader pass `name`, changing the failing kwarg from `context` to `name` |

#1800's reasoning was correct as far as it went: `**kwargs` was forwarded to `RemoteRecognizer.__init__`, which does not accept `**kwargs`, so it never absorbed anything — it relayed the same `TypeError` one frame deeper, with a traceback pointing at `remote_recognizer.py` rather than at the recognizer the user configured. What the PR did not do was add the explicit parameters the loader actually passes. The other two files it touched have since been repaired — `ahds_recognizer.py` and `lm_recognizer.py` both take `name: Optional[str] = None` today. `azure_ai_language.py` is the one left.

One wrinkle for whoever fixes it: the class hardcodes its display name in the `super()` call.

```python
super().__init__(..., name="Azure AI Language PII", version="5.2.0")
```

With `**kwargs` still in place, a caller-supplied `name` would have raised `got multiple values for keyword argument 'name'` instead. So the parameter has to be added with a precedence rule, not merely forwarded. Keeping `"Azure AI Language PII"` as the default when the caller passes nothing preserves the current behaviour and the existing assertion in the muted test (`rec.name == "Azure AI Language PII"`).

Note that #2170's contract tests do not cover this class: it is absent from `default_recognizers.yaml`, so the load test does not reach it, and it is not a `PatternRecognizer`, so the signature test does not either. It is still reachable from a user config by name, as #1457 shows.

## 3. The AHDS twin needs an environment variable, not a package

`test_analyzer_engine_provider_with_ahds`, un-muted:

```
ValueError: AHDS de-identification endpoint is required.
            Please provide an endpoint or set the AHDS_ENDPOINT environment variable.
```

Its gate is wrong in both mechanism and condition. `tests/test_ahds_recognizer.py:26` already has the right shape for this (`requires_env_vars()` → `skipif(endpoint == "")`); this test needs the same condition in addition to the dependency check.

## 4. The gate names the wrong module

`azure.*` are namespace packages, and two extras contribute to them:

```toml
azure-ai-language = ["azure-ai-textanalytics", "azure-core"]
ahds              = ["azure-identity", "azure-health-deidentification"]
```

Installing only `ahds` makes `import azure` succeed without `azure.ai.textanalytics` present, so even a correctly-written gate on the bare `azure` name would let the Azure AI Language test through into an `ImportError`. The condition should name the module each test actually needs.

## Proposed fix

Gate on `importlib.util.find_spec` rather than `importorskip`, so that a missing extra skips the individual tests instead of dropping the module from collection:

```python
import importlib.util

_HAS_TEXT_ANALYTICS = importlib.util.find_spec("azure.ai.textanalytics") is not None

@pytest.mark.skipif(not _HAS_TEXT_ANALYTICS,
                    reason="azure-ai-language extra not installed")
def test_analyzer_engine_provider_with_azure_ai_language():
    ...
```

plus the `AHDS_ENDPOINT` condition on the AHDS test, and `name` / `context` parameters on `AzureAILanguageRecognizer.__init__`.

These want to land together: fixing the marker alone turns the Azure AI Language test red on `main`, since the defect it detects is still there.

## Environment

`main` at 94168ca7, `presidio-analyzer`, Python 3.12.10, Windows. The "extra installed" rows were produced with the real `azure-ai-textanalytics` / `azure-core`, not a stub.

One practical note for anyone reproducing the "not installed" rows: `pip uninstall azure-ai-textanalytics azure-core` is not enough. `azure-common` also claims the `azure` namespace, and pip leaves an empty `site-packages/azure/ai/` directory behind, either of which keeps `import azure` succeeding and puts you back on the `skipif` branch. The directory has to be removed for `importorskip("azure")` to see an absent module.

developer0hye:
## Change Description

`AzureAILanguageRecognizer` cannot be built from a registry configuration on `main`. `RecognizerListLoader` passes `name` for every entry it constructs, and passed `context` before that; the constructor accepts neither, so any YAML entry naming the class raises before the registry finishes loading:

```python
RecognizerListLoader.get(
    recognizers=[{"name": "AzureAILanguageRecognizer", "type": "predefined",
                  "supported_languages": ["en"]}],
    supported_languages=["en"], global_regex_flags=26,
)
# TypeError: AzureAILanguageRecognizer.__init__() got an unexpected keyword argument 'name'
```

This is #1457, reported in 2024 against the `context` kwarg and fixed in #1458 by adding `**kwargs`. #1800 removed the `**kwargs` — correctly, since `RemoteRecognizer.__init__` does not accept them either, so it never absorbed anything and only moved the same `TypeError` one frame deeper — but without adding the explicit parameters that were holding the loader path up. The other two recognizers that PR touched have since been repaired; `ahds_recognizer.py` and `lm_recognizer.py` both take `name: Optional[str] = None` today.

The regression test #1458 added was already unable to run by then, which is the second half of this PR.

## Why the regression test could not fire

```python
@pytest.mark.skipif(pytest.importorskip("azure"), reason="Optional dependency not installed")
```

`pytest.importorskip` skips when a module is missing and otherwise returns the module, which is truthy. Both branches lose:

| `azure` present? | Result |
| --- | --- |
| No | `Skipped` is raised while the decorator is evaluated, i.e. at module import — **the whole file drops out of collection** |
| Yes | truthy module → `skipif(True)` → **skipped**, reporting `Optional dependency not installed` while it is installed |

Measured on `main`:

```
$ python -m pytest tests --collect-only -q | tail -1
3248 tests collected      # without the azure-ai-language extra
3292 tests collected      # with it
```

So 44 tests — the entire module — are silently uncollected in any environment without the extra, and the two tests the marker was written for skip on every CI run, where `--all-extras` is used (`ci.yml:74`).

`azure` is also the wrong module to name. `azure.*` are namespace packages and the `ahds` extra populates them too, so `import azure` can succeed with `azure.ai.textanalytics` absent.

## Changes

- **`azure_ai_language.py`** — added `name` and `context`, appended last so existing positional callers are unaffected. `name` falls back to `"Azure AI Language PII"` when omitted, matching how `AzureHealthDeidRecognizer` resolves the same conflict between a configured name and a hardcoded display name.
- **`test_analyzer_engine_provider.py`** — both markers now gate on `importlib.util.find_spec` for the module each test actually imports. A local helper swallows the `ModuleNotFoundError` that `find_spec` raises when an intermediate parent package is missing, which is what `azure.health.deidentification` does without the `ahds` extra.
- The AHDS test additionally requires `AHDS_ENDPOINT`. `AzureHealthDeidRecognizer.__init__` builds a client when the config supplies none, and that path raises `ValueError` without the variable — so the endpoint is as much a precondition as the package. `tests/test_ahds_recognizer.py:26` already gates on it this way.
- **`tests/conf/test_azure_ai_language_reco.yaml`** — dropped `ta_client: "test"`, which no longer reaches the constructor (see follow-up 1 below), and added a second entry so the fixture covers both entry shapes: the plain form where `name` doubles as the class selector, and the `class_name` rename form. The test supplies credentials through `monkeypatch.setenv` instead; nothing reaches the network, as the SDK client is constructed locally and `analyze` is overridden.
- **`tests/test_azure_ai_language_recognizer.py`** — three direct-construction tests for the name default, the name override, and `context`.

## Verification

- Reverting only `azure_ai_language.py` fails exactly three tests — `test_name_can_be_overridden`, `test_context_is_accepted`, `test_analyzer_engine_provider_with_azure_ai_language` — and nothing else.
- Full `presidio-analyzer` suite, this branch against an `origin/main` baseline in the same environment: **identical failure sets**, 42 pre-existing failures/errors from optional dependencies absent locally (`transformers`, `langextract`, some spaCy models). Compared line by line, not by count.

  ```
  main:   26 failed, 3168 passed, 83 skipped, 16 errors
  branch: 26 failed, 3172 passed, 82 skipped, 16 errors
  ```

  The delta is exactly this PR: three new tests, plus the Azure AI Language test moving from skipped to passing.
- With the extra uninstalled and `site-packages/azure` removed, `pytest tests --collect-only` reports **3295** on this branch against **3248** on `main`. The module is collected, its 41 other tests run, and only the two optional-dependency tests skip, with accurate reasons.
- `ruff check` from the repo root as CI runs it, `ruff format --check` on the modified module, and `git diff --check` all pass.

Reproducing the "extra absent" rows takes more than `pip uninstall azure-ai-textanalytics azure-core`: `azure-common` also claims the `azure` namespace, and pip leaves an empty `site-packages/azure/ai/` behind. Either keeps `import azure` succeeding.

## Follow-ups, not in this PR

1. **`PredefinedRecognizerConfig` silently drops recognizer-specific kwargs.** It inherits pydantic's default `extra="ignore"`, so `ta_client` — and `azure_ai_key` / `azure_ai_endpoint`, which users would reasonably put in YAML — never reach the constructor. `HuggingFaceRecognizerConfig`, `GLiNERRecognizerConfig` and `LangExtractRecognizerConfig` each set `extra="allow"` to work around this one class at a time. Worth a decision on whether that should be the default for predefined entries.
2. **`AzureHealthDeidRecognizer` does not accept `context` either.** Same shape as the `name` gap fixed here, and the same YAML path reaches it. Left alone to keep this diff scoped.
3. `AzureAILanguageRecognizer` is still absent from `default_recognizers.yaml`, so #2170's contract tests do not reach it — the load test only walks shipped entries, and the signature test only covers `PatternRecognizer` subclasses. Adding a shipped entry is a separate question, since the recognizer needs credentials to construct.

Per the changelog policy in #2200, `CHANGELOG.md` is not modified.

## Issue reference

Closes #2223. Follows up on #1457 / #1458 (the original report and fix), #1521 (which introduced the marker), #1800 (which removed the `**kwargs`) and #1819 (which made the loader pass `name`).

## Checklist

### DIFF
--- a/presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/azure_ai_language.py
+++ b/presidio-analyzer/presidio_analyzer/predefined_recognizers/third_party/azure_ai_language.py
@@ -25,6 +25,8 @@ class AzureAILanguageRecognizer(RemoteRecognizer):
         ta_client: Optional["TextAnalyticsClient"] = None,
         azure_ai_key: Optional[str] = None,
         azure_ai_endpoint: Optional[str] = None,
+        name: Optional[str] = None,
+        context: Optional[List[str]] = None,
     ):
         """
         Wrap the PII detection in Azure AI Language.
@@ -36,15 +38,24 @@ class AzureAILanguageRecognizer(RemoteRecognizer):
         the client will be created using the key and endpoint.
         :param azure_ai_key: Azure AI for language key
         :param azure_ai_endpoint: Azure AI for language endpoint
+        :param name: Name of the recognizer. Defaults to "Azure AI Language PII".
+        :param context: List of context words to boost confidence score.
 
         For more info, see https://learn.microsoft.com/en-us/azure/ai-services/language-service/personally-identifiable-information/overview
         """
 
+        # `name` and `context` are appended last so existing positional callers
+        # are unaffected. They exist because RecognizerListLoader passes both
+        # when building a recognizer from a registry configuration, and
+        # RemoteRecognizer.__init__ accepts neither **kwargs nor unknown keys.
+        # Without them, any YAML entry naming this class raises TypeError before
+        # the registry finishes loading (#1457, and again after #1800).
         super().__init__(
             supported_entities=supported_entities,
             supported_language=supported_language,
-            name="Azure AI Language PII",
+            name=name if name else "Azure AI Language PII",
             version="5.2.0",
+            context=context,
         )
 
         is_available = bool(TextAnalyticsClient)
--- a/presidio-analyzer/tests/conf/test_azure_ai_language_reco.yaml
+++ b/presidio-analyzer/tests/conf/test_azure_ai_language_reco.yaml
@@ -1,9 +1,15 @@
 recognizer_registry:
   global_regex_flags: 26
   recognizers:
+    # Plain form: `name` doubles as the class selector and is also passed to the
+    # constructor. This is the entry shape from #1457.
     - name: MockAzureAiLanguageRecognizer
       type: predefined
-      ta_client: "test" # This is a placeholder for testing purposes
+    # Rename form: `class_name` selects the class, `name` becomes the instance
+    # name. This is the documented reason the loader passes `name` at all.
+    - name: Azure AI Language PII
+      class_name: MockAzureAiLanguageRecognizer
+      type: predefined
 
 
 supported_languages:
--- a/presidio-analyzer/tests/test_analyzer_engine_provider.py
+++ b/presidio-analyzer/tests/test_analyzer_engine_provider.py
@@ -1,5 +1,7 @@
 # ruff: noqa: D103,D205,E501,F541,F841,W293
 
+import importlib.util
+import os
 import re
 from pathlib import Path
 from typing import List
@@ -16,6 +18,27 @@ from presidio_analyzer.predefined_recognizers import (
     SpacyRecognizer,
     StanzaRecognizer,
 )
+
+def _has_module(name: str) -> bool:
+    # find_spec imports the parent packages of a dotted name, so a missing
+    # intermediate raises ModuleNotFoundError instead of returning None --
+    # "azure.health.deidentification" does exactly that without the ahds extra.
+    try:
+        return importlib.util.find_spec(name) is not None
+    except (ImportError, ValueError):
+        return False
+
+
+# Gate the optional-dependency tests below on the module each one actually
+# imports. `pytest.importorskip` cannot be used for this: evaluated inside a
+# decorator it runs at module import, so a missing extra takes the whole file
+# out of collection instead of skipping the two tests that need it.
+#
+# The bare `azure` name is also too coarse. `azure.*` are namespace packages and
+# the `ahds` extra populates them too, so `import azure` can succeed while
+# `azure.ai.textanalytics` is absent.
+_HAS_TEXT_ANALYTICS = _has_module("azure.ai.textanalytics")
+_HAS_HEALTH_DEID = _has_module("azure.health.deidentification")
 
 
 def get_full_paths(analyzer_yaml, nlp_engine_yaml=None, recognizer_registry_yaml=None):
@@ -241,9 +264,15 @@ def test_analyzer_engine_provider_with_files_per_provider():
 
 
 @pytest.mark.skipif(
-    pytest.importorskip("azure"), reason="Optional dependency not installed"
-)  # noqa: E501
-def test_analyzer_engine_provider_with_azure_ai_language():
+    not _HAS_TEXT_ANALYTICS, reason="azure-ai-language extra not installed"
+)
+def test_analyzer_engine_provider_with_azure_ai_language(monkeypatch):
+    # The constructor builds a client when none is injected, falling back to
+    # these variables. Nothing here reaches the network: the SDK client is
+    # constructed locally and `analyze` is overridden below.
+    monkeypatch.setenv("AZURE_AI_KEY", "test-key")
+    monkeypatch.setenv("AZURE_AI_ENDPOINT", "https://example.invalid/")
+
     analyzer_yaml, _, _ = get_full_paths(
         "conf/test_azure_ai_language_reco.yaml",
     )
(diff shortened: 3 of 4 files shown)


omri374 (approved):
Thanks!

PR #2226, opened 2026-08-08
fix: snap chunk start to a word boundary in CharacterBasedTextChunker

inesaranab:
## Change Description

`CharacterBasedTextChunker` snaps `end` forward to the next boundary char so a chunk never *ends* mid-word. But `start`, computed as `end - chunk_overlap`, gets no equivalent treatment — the subtraction discards the alignment established two lines earlier.

With `chunk_size=10, chunk_overlap=3`:

text = "This is a test string for chunking purposes"

before: ['This is a test', 'st string for', 'or chunking', 'ing purposes']
after:  ['This is a test', ' string for', ' chunking purposes']

Three of the four chunks begin mid-word. An NER recognizer receives `'st'`, `'or'`, `'ing'` as standalone tokens and can confidently mislabel them as entities, producing a false positive that redacts the wrong span.

**Fix:** advance `start` to the next boundary char, mirroring the loop `end` already has. Three lines.

Present since #1805, which introduced this chunker.

## Issue reference

No existing issue. Found while using the chunker via `GLiNERRecognizer` on long transcripts.

## Checklist

### DIFF
--- a/presidio-analyzer/presidio_analyzer/chunkers/character_based_text_chunker.py
+++ b/presidio-analyzer/presidio_analyzer/chunkers/character_based_text_chunker.py
@@ -119,5 +119,8 @@ class CharacterBasedTextChunker(BaseTextChunker):
                 break
             start = end - self._chunk_overlap
 
+            while start < end and text[start] not in self._boundary_chars:
+                start += 1
+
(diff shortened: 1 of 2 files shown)


omri374 (approved):
Approved! Thanks

PR #1857, opened 2026-02-14
feat: Add UK driving licence number recognizer

tee-jagz:
Add UkDrivingLicenceRecognizer for detecting DVLA driving licence numbers. The 16-character alphanumeric format encodes surname, date of birth, and initials. Uses regex with embedded month/day validation and surname padding checks. Disabled by default per country-specific convention.

## Change Description

Add a new country-specific recognizer for UK driving licence numbers (entity: `UK_DRIVING_LICENCE`). The recognizer detects 16-character DVLA licence numbers using regex pattern matching with context words and structural validation of the surname padding. The DVLA check digit algorithm is not public, so validation rejects clearly invalid patterns but does not confirm validity.

**Files changed:**

- New: `uk_driving_licence_recognizer.py` — recognizer class
- New: `test_uk_driving_licence_recognizer.py` — 21 tests
- Updated: `default_recognizers.yaml`, `__init__.py` exports, `supported_entities.md`
- Updated: `default_recognizers.yaml` — registered with `enabled: false`
- Updated: `supported_entities.md` — added `UK_DRIVING_LICENCE` row

## Issue reference

N/A — new feature contribution

## Checklist

- [x] I have reviewed the [contribution guidelines](https://github.com/microsoft/presidio/blob/main/CONTRIBUTING.md)
- [x] I have signed the CLA (if required)
- [x] My code includes unit tests
- [x] All unit tests and lint checks pass locally
- [x] My PR contains documentation updates / additions if required

PR #2236, opened 2026-08-28
fix(analyzer): remove a non-language code and a duplicate entry from default_recognizers.yaml

heeoneie:
## Change Description

Two data problems in `presidio-analyzer/presidio_analyzer/conf/default_recognizers.yaml`, with a regression test for each.

### 1. `kr` is not a language code

The five Korean entries listed `kr` alongside `ko` under `supported_languages`. `ko` is the ISO 639-1 language code; `kr` is the ISO 3166-1 country code, which those same entries already carry in `country_code`. Korea is the only country in the file where the country code leaked into the language list — Canada, the closest analogue, correctly pairs `country_code: ca` with `supported_languages: [en, fr]`. `AGENTS.md` states the rule directly: "Use ISO 639-1 language codes (`ko` for Korean, never `kr`)", so the shipped default contradicts the repository's own instruction to contributors.

Per review this is now a **deprecation, not a removal**. `kr` goes from `KrDriverLicenseRecognizer`, `KrBrnRecognizer` and `KrFrnRecognizer`, which have only ever defaulted to `ko`, so no configuration can depend on it. It stays on `KrRrnRecognizer` and `KrPassportRecognizer`, whose class defaults were `kr` until #1742 (2025-10-08) and #2170 (2026-08-05), and is scheduled for removal in a later release.

**Correction to my earlier description.** @omri374 caught two errors, and they are not the same kind of error.

The first was a misplaced condition. I put "**10** Korean recognizers" in the `["en", "ko"]` row. The count itself is real — `main` does build 10 — but only when the registry lists both `ko` and `kr`. With `["en", "ko"]` it builds 5, because `_is_language_supported_globally` drops every instance whose language the registry does not name, and `kr` is not one of them. The number was right; the row it sat in was wrong. The table below puts it on the configuration it actually describes.

The second was wrong outright, not merely mis-scoped. I claimed the second copy could return duplicate results. It cannot, under any registry configuration: `RecognizerRegistry.get_recognizers` filters on `language == rec.supported_language`, so a single `analyze()` call only ever runs one language's instances. That claim is withdrawn rather than narrowed.

Measured by building a registry from the shipped YAML with the five Korean entries enabled and counting `registry.recognizers` plus "language is not supported" warnings:

| registry `supported_languages` | before | after |
| --- | --- | --- |
| `["ko"]` | 5 recognizers, **5 warnings** per registry build | 5 recognizers, 2 warnings |
| `["en", "ko"]` | 5 recognizers, 5 warnings | 5 recognizers, 2 warnings |
| `["ko", "kr"]` | 10 recognizers (5 `ko` + 5 `kr`), 0 warnings | 7 (5 `ko` + 2 `kr`), 0 warnings |
| `["kr"]` | 5 recognizers, all `kr`, 5 warnings | 2 recognizers, both `kr`, 5 warnings — the retained alias |

A warning is emitted per instance whose declared language the registry does not name, and it runs in both directions: `["ko", "kr"]` names both codes and so emits none, while `["kr"]` emits five for the `ko` instances it discards. The warnings on a `["ko"]` setup are the same mechanism pointed the other way.

So the actual cost of `kr` is five warnings on every registry build of a normal `["ko"]` setup, plus a shipped default that contradicts the ISO 639-1 rule in `AGENTS.md`. The doubling is real but only on a registry that lists both codes, and even there the two sets never meet in one `analyze()` call. The two remaining warnings after this PR are the deprecated alias announcing itself, which is the intended cost of keeping it.

`ConfigurationValidator.validate_language_codes` does not catch any of this — `kr` is a well-formed two-letter code, just not the right one.

This also touches one existing test. #2170 parametrised `test_loads_from_default_recognizers_yaml` over `["ko", "kr"]`. An earlier revision of this PR dropped the `kr` case; with the alias retained it is back, now asserting a documented alias rather than the defect.

### 2. `UkPostcodeRecognizer` was listed twice

#1858 added it; #1857 added it again and merged later, so the file carried two byte-identical blocks and the registry built two identical instances.

To be precise about severity: analyzer output is unaffected, because identical spans are merged downstream. The cost is a duplicate regex pass and a config file that misstates itself.

That understates it, as @hfyeomans points out below. Construction is gated per config entry, but removal is not: `__remove_disabled_nlp_recognizers` builds its disabled set from the *class names* of entries that are not enabled, then drops every instance of those classes for the language. So if someone later enables one of the two blocks and leaves the other as it is, the recognizer is built from the enabled entry and then removed again by the class-name match on the disabled one, leaving only a `logger.info` behind. Removing the duplicate makes that path unreachable for this entry. The class-name-wide matching itself is a separate question about the loader and belongs in its own issue, not here.

## Issue reference

No existing issue — I searched and found none for either finding. Happy to open one first if you would rather have that order.

## Verification

Environment: `uv sync --locked --all-extras --group dev`, plus `en_core_web_lg`, per `AGENTS.md`.

- **Full `presidio-analyzer` suite: 3459 passed, 13 skipped, 0 failed.** This supersedes what I wrote earlier. That claim — 36 failing/erroring items, byte-identical before and after — described an environment missing the optional extras (langextract, azure, slim_spacy). With the extras installed as `AGENTS.md` specifies, nothing fails, so there is no before/after failure set to compare: the suite is simply green.
- **Each of the four commits is green on its own**, checked by building out each one and running the two affected test files (87 / 88 / 89 / 89 passed).
- **The regression tests do catch the regression.** Against `main`'s YAML all three fail:
  - `duplicate recognizer entries: ['UkPostcodeRecognizer']`
  - `non-language codes in supported_languages: {'KrBrnRecognizer': ['kr'], 'KrDriverLicenseRecognizer': ['kr'], 'KrFrnRecognizer': ['kr']}` — exactly the three entries that lose the alias, with the two carved-out entries correctly absent
  - the alias test, because all five recognizers load under `["kr"]` on `main`
- **The carve-out is enforced, not just documented.** Removing `kr` from the two retained entries fails two tests: `test_deprecated_kr_alias_still_loads_its_recognizers` and `test_loads_from_default_recognizers_yaml[kr]`.
- **The before/after table above was measured, not inferred**: I built a registry from the shipped YAML with the Korean entries enabled and counted `registry.recognizers` and the "language is not supported" warnings for each language configuration.
- `ruff check` passes. Test files are excluded from it by `*/tests/*`, so I also ran both touched files by explicit path and compared against this branch's base: identical findings on both (one pre-existing `D209` in `test_recognizer_registry.py`, seven in `test_kr_passport_recognizer.py`). My lines add none.

## A note on the language guard

`test_default_recognizers_yaml_declares_languages_not_countries` keeps an explicit set of ISO 639-1 codes, so that adding a language becomes a deliberate edit rather than a side effect of a new entry. It now also carries a carve-out for the two entries that keep `kr`, with a comment naming the deprecation and the release the alias goes away in.

**One open question there.** That comment, and the changelog entry, are both supposed to name the removal release, and I left a `<release TBD>` placeholder rather than guess at a version. Tell me which release it should be and I will fill both in.

**And one note on where the deprecation is documented.** You asked for a line in the migration docs; I could not find such a document in this repository. The closest are `docs/project_transition.md`, which covers the move to Data Privacy Stack, and `docs/build_release.md`, which covers release machinery — neither is a migration or upgrade guide. I put the note in `CHANGELOG.md` under a new Analyzer → `#### Deprecated` heading, since that is where this repository already records user-visible changes. Happy to move it if you had another file in mind.

## AI use

Claude Opus 5 assisted with the investigation, the edits, and running the checks above; each commit carries a `Generated-by:` trailer. Every claim here was verified against a local checkout before it was written.

## Checklist

- [x] All unit tests and lint checks pass locally — scoped exactly as described under Verification


heeoneie:
Thanks — all of this is addressed, pushed as four commits (`116afb0d`..`5bc2c7b6`).

**UK postcode duplicate.** Unchanged, as you said. The duplicate guard now also records @hfyeomans's point in its docstring, so the next reader learns why it is more than a wasted regex pass.

**`kr` alias — option 2.** Dropped from `KrDriverLicenseRecognizer`, `KrBrnRecognizer` and `KrFrnRecognizer`. Kept on `KrRrnRecognizer` and `KrPassportRecognizer`. That also restores the `kr` case of `test_loads_from_default_recognizers_yaml`, which an earlier revision of this PR had removed; it now asserts a documented alias instead of the defect it was reverted for.

**Carve-out.** `test_default_recognizers_yaml_declares_languages_not_countries` carries a `deprecated_language_aliases` allowance for exactly those two entries, with a comment naming both original class defaults, why the alias is kept, and the release it goes away in. The three siblings get no allowance, so the guard still fails if `kr` reappears on them — verified against `main`'s YAML, where it reports exactly `{'KrBrnRecognizer': ['kr'], 'KrDriverLicenseRecognizer': ['kr'], 'KrFrnRecognizer': ['kr']}` and leaves the two carved-out entries out of it.

**Configuration-path test.** `test_deprecated_kr_alias_still_loads_its_recognizers` builds a registry through `RecognizerRegistryProvider` with `supported_languages: ["kr"]` and the Korean entries enabled, and asserts that exactly `KrRrnRecognizer` and `KrPassportRecognizer` load, both under `kr`. Removing the alias from the YAML fails it, so the reason for keeping it is enforced rather than only stated in a comment.

**Entry parsing.** Both guards normalize before reading anything, so a bare-string entry produces an assertion rather than a `TypeError`.

One deviation worth flagging, since it goes against the obvious reading of your note. I did *not* route the duplicate check through `RecognizerListLoader.get_recognizer_name()`. That helper prefers `class_name`, because it answers "which class do I instantiate" — so keying the duplicate check on it would have reported `{name: A, class_name: X}` and `{name: B, class_name: X}` as duplicates, which is precisely the shape you said is the correct way to have two entries of the same class. Instead a small `_entry_name()` prefers the instance `name` and falls back to `get_recognizer_name()` only for a bare-string entry or one that names only a class, where the class name *is* the instance name. The check stays keyed on name alone, as you asked.

I also widened the language side beyond string entries. `_get_recognizer_languages` accepts three shapes — absent or `None`, a list of codes, and a list of `{"language": ..., "context": ...}` mappings — so normalizing only string *entries* would still have left the mapping form unhandled. `_declared_languages()` covers all three.

**Description.** Corrected. You caught two errors and they are different in kind, so the description now separates them: "10 recognizers" is a real count that I attached to the wrong row (it happens on `["ko", "kr"]`, not `["en", "ko"]`, because `_is_language_supported_globally` drops the `kr` instances a registry never names), whereas the duplicate-results claim is withdrawn outright — `get_recognizers` filters on `language == rec.supported_language`, so no registry configuration can make one `analyze()` call run both sets.

The table is replaced with four configurations I measured rather than reasoned about. One number I want to state plainly rather than leave for someone to find later: option 2 does **not** take the warnings to zero. A normal `["ko"]` setup goes from 5 warnings per registry build to 2, because the two retained aliases announce themselves on every build. That is the standing cost of keeping them until the removal release.

**Deprecation note — please redirect me if this is the wrong home.** You asked for a line in the migration docs, and I could not find such a document in this repository. The closest are `docs/project_transition.md`, which covers the move to Data Privacy Stack, and `docs/build_release.md`, which covers release machinery; neither is a migration or upgrade guide. Rather than guess at a path, I put the note in `CHANGELOG.md` under a new Analyzer → `#### Deprecated` heading, which is where this repository already records user-visible changes and where the only prior `Deprecated` section lives. Happy to move it if you had another file in mind.

**One question before this is finished.** Both the carve-out comment and the changelog entry are meant to name the release the alias goes away in, and you did not name one — so I left a `<release TBD>` placeholder rather than invent a version. Which release should they say? I will fill in both in a follow-up commit.

Verification: full `presidio-analyzer` suite is 3459 passed, 13 skipped, 0 failed on `uv sync --locked --all-extras --group dev` plus `en_core_web_lg`; each of the four commits is green on its own; `ruff check` is clean and both touched test files report the same findings as this branch's base.

omri374:
on the migration docs- I meant the changelog. Please add a comment under "changed"

heeoneie:
Moved. The note now sits under `### Analyzer` -> `#### Changed` in the `[unreleased]` block, following the `Added` -> `Changed` -> `Fixed` order used in 2.2.363. The `#### Deprecated` heading is gone and the bullet text is unchanged.

Two `<release TBD>` placeholders are still open - one in `CHANGELOG.md` and one in `presidio-analyzer/tests/test_recognizer_registry.py`, where a test pins the release in which the `kr` alias is removed. I'm happy to fill both in as soon as you know which release that will be. If you'd rather not commit to a version yet, I can reword them to describe the deprecation without naming a release, so this doesn't sit waiting on a number.

omri374 (approved):
Thanks!

heeoneie:
Thanks for merging this.

One loose end: the two `<release TBD>` placeholders went in as-is — one in
`CHANGELOG.md` and one in `presidio-analyzer/tests/test_recognizer_registry.py`,
where a test pins the release in which the `kr` alias is removed. Happy to
send a one-line follow-up PR filling both in once the version is decided, or
to reword them so they describe the deprecation without naming a release if
you'd rather not pin one. Just let me know which you prefer.

PR #2245, opened 2026-09-09
fix(analyzer): match full TLDs in UrlRecognizer instead of a prefix

### LINKED ISSUE #2244: UrlRecognizer.BASE_URL_REGEX does not match full URL for some TLDs
**Describe the bug**
UrlRecognizer.BASE_URL_REGEX does not match full URL for some TLDs

I ran into issues with a substring match on  strings like `www.some.bank`.  Both `ba` and `bank` are TLDs.
The regex accounts for `ba`, but not `bank`. 

There are numerous other TLDs that have a common substring between a two letter TLD and a longer TLD.

https://data.iana.org/TLD/tlds-alpha-by-domain.txt

I'm going to fix this in a fork, let me know if you'd like a PR.

**To Reproduce**
1. detect URLs for the string `www.some.bank`
2. see match is `www.some.ba`

**Expected behavior**
Expect match to be `www.some.bank`

stephencattaneo:
The URL patterns never asserted that the matched TLD ends the host, so the engine stopped at the first alternative that fit. `chase.bank` matched only `chase.ba`, `home.barclays` only `home.bar`, and `www.americanexpress` only `www.am` - silently truncating the host and dropping the path, so anonymizing `chase.bank/login` left `nk/login` in the output.

Add a `(?![a-z0-9\-])` boundary after the TLD. Ordering alone cannot fix this: `bar` is itself a gTLD, so `home.barclays` truncates regardless of how the alternation is sorted.

Replace the hand-maintained TLD list with the full IANA root zone (1438 entries, vendored as tlds.txt with update_tlds.py to refresh it), so financial TLDs such as .bank, .barclays, .americanexpress, .insurance and .creditunion are covered.

Compress the alternation into a prefix tree rather than a flat `(?:aaa)|(?:aarp)|...`. Despite covering twice as many TLDs, matching is ~30% faster (32.9ms -> 22.8ms per 26KB document) since the engine can discard whole subtrees on a single character; compiled patterns cost +4.4MB, once per process, as PATTERNS is class-level.

Keeping the TLD set inside the regex, rather than validating the last label in Python, avoids false negatives: the host part is greedy, so `Visit www.example.com.Then call` matches `www.example.com.Then`, and post-hoc validation can only reject the whole match. Backtracking finds the correct boundary and reports `www.example.com`.

Note `www.microsoft` is now detected, as MICROSOFT is a delegated brand TLD - the same reason BARCLAYS and AMERICANEXPRESS are matched.

## Change Description

Describe your changes

## Issue reference

Fixes #2244 

## Checklist

- [X] I have reviewed the [contribution guidelines](https://github.com/data-privacy-stack/presidio/blob/main/CONTRIBUTING.md)
- [X] I agree to follow this project's [Code of Conduct](https://github.com/data-privacy-stack/presidio/blob/main/CODE_OF_CONDUCT.md)
- [X] I confirm that I have the right to submit this contribution and that it does not knowingly contain proprietary or confidential code.


presidio-analyzer/presidio_analyzer/predefined_recognizers/generic/url_recognizer.py
```diff
@@ -18,7 +109,16 @@ class UrlRecognizer(PatternRecognizer):
...
+    BASE_URL_REGEX = (
+        r"((www\d{0,3}[.])?[a-z0-9.\-]{1,253}[.]"
+        + tlds_to_regex(TLDS)
+        # Require the TLD to end the host. Without this the engine happily
+        # stops on a shorter TLD that merely prefixes the real one.
+        + r"(?![a-z0-9\-])"
+        + r"(?:/[^\s()<>\"']*)?)"
+    )
 
     PATTERNS = [
```
omri374, line 123: Please split the no-scheme pattern by TLD case so confidence reflects the false-positive risk:

- `Standard Url` with scheme: unchanged, 0.6, case-insensitive. `https://Example.Com` is a URL regardless of case.
- `Non schema URL`: 0.5, TLD must be all-lowercase or all-uppercase: `(?-i:` + lowercase trie + `|` + uppercase trie + `)` inside the existing `(?i)` pattern. The `regex` module supports scoped flags.
- `Non schema URL, mixed-case TLD`: 0.3, the current case-insensitive pattern.

`remove_duplicates` already keeps the higher-scoring result for an identical span, so no `analyze` override is needed, and the pattern name lands in `AnalysisExplanation` for free. I prototyped this and got `to.microsoft` / `TO.MICROSOFT` / `www.chase.bank` at 0.5, `Done.Next` / `to.Microsoft` / `Example.Com` at 0.3, `https://Example.Com` at 0.6.

Please add those as test cases. Note `Example.Com` already matched at 0.5 on main, so this is a behavior change to declare in the description.

presidio-analyzer/presidio_analyzer/predefined_recognizers/generic/url_recognizer.py
```diff
@@ -1,7 +1,98 @@
...
+    the pattern smaller and lets the engine discard whole subtrees on a single
+    character.
+
+    :param tlds: TLDs to match, case-insensitively
+    :return: A non-capturing regex group matching exactly those TLDs
+    """
+    root: Dict = {}
+    for tld in tlds:
+        node = root
+        for char in tld.lower():
```
omri374, line 47: Gotcha for the uppercase trie above: this lowercases the input, so `tlds_to_regex(t.upper() for t in tlds)` returns the same lowercase fragment. Either drop the lowercasing here (the file is already lowercase) or `.upper()` the generated fragment, which is safe since it only contains letters, digits and regex punctuation.

presidio-analyzer/pyproject.toml
```diff
@@ -20,7 +20,7 @@ classifiers = [
 keywords = ["presidio_analyzer"]
 urls = {Homepage = "https://github.com/data-privacy-stack/presidio"}
 readme = "README.md"
-include = ["conf/*",]
+include = ["conf/*", "predefined_recognizers/generic/tlds.txt"]
```
omri374, line 23: This is a no-op, please revert. `include` under `[project]` is not a PEP 621 field, so poetry-core ignores it, and the path would be wrong relative to the project root anyway. I built the wheel with and without this line: `tlds.txt` ships either way, because poetry-core packages every file under the package directory.

presidio-analyzer/presidio_analyzer/predefined_recognizers/generic/update_tlds.py
```diff
@@ -0,0 +1,41 @@
+"""Refresh the vendored IANA root zone TLD list used by ``UrlRecognizer``."""
```
omri374, line 1: This is a maintenance script that does network I/O, and it currently ships in the wheel (I checked the built artifact). Please move it out of the package and we'll periodically update it manually

PR #2246, opened 2026-09-09
fix(analyzer): use the PatternRecognizer default when deny_list_score is omitted in a registry config

### LINKED ISSUE #2242: Omitted deny_list_score yields 0.0 through RecognizerRegistryProvider but 1.0 through PatternRecognizer and add_pattern_recognizer_from_dict
## Describe the bug

A custom deny-list recognizer that omits `deny_list_score` gets a different score depending on how it is loaded:

| Path | `deny_list_score` | Deny-list pattern score |
|---|---|---|
| `PatternRecognizer(supported_entity=..., deny_list=[...])` | 1.0 | 1.0 |
| `RecognizerRegistry.add_pattern_recognizer_from_dict({...})` | 1.0 | 1.0 |
| `RecognizerRegistryProvider(registry_configuration={...})` (YAML / dict registry config) | **0.0** | **0.0** |

Cause: `CustomRecognizerConfig.deny_list_score` in `presidio_analyzer/input_validation/yaml_recognizer_models.py` declares `default=0.0`, and the validated model is dumped with that default present, so the loader passes `deny_list_score=0.0` to `PatternRecognizer` even when the YAML said nothing. `PatternRecognizer.__init__` defaults to `1.0`.

A deny-list match with score 0.0 is filtered out by any positive `score_threshold`, so a YAML deny list without an explicit score effectively never fires through the provider path.

## To Reproduce

```python
from presidio_analyzer import PatternRecognizer
from presidio_analyzer.recognizer_registry import RecognizerRegistry, RecognizerRegistryProvider

conf = {
    "supported_languages": ["en"],
    "global_regex_flags": 26,
    "recognizers": [
        {"name": "Titles", "supported_entity": "TITLE", "deny_list": ["Mr.", "Mrs."]}
    ],
}
reg = RecognizerRegistryProvider(registry_configuration=conf).create_recognizer_registry()
r = next(x for x in reg.recognizers if x.name == "Titles")
print(r.deny_list_score, [p.score for p in r.patterns])   # 0.0 [0.0]

print(PatternRecognizer(supported_entity="TITLE", deny_list=["Mr."]).deny_list_score)  # 1.0

reg2 = RecognizerRegistry()
reg2.add_pattern_recognizer_from_dict(
    {"name": "T2", "supported_language": "en", "supported_entity": "TITLE", "deny_list": ["Mr."]}
)
print(reg2.recognizers[-1].deny_list_score)  # 1.0
```

## Expected behavior

Omitting `deny_list_score` means the `PatternRecognizer` default (`1.0`) on every path. An explicit value is honored as today.

## Proposed fix

Change `CustomRecognizerConfig.deny_list_score` to `Optional[float] = Field(default=None, ge=0.0, le=1.0)` and dump recognizer entries with `exclude_unset=True`, so an omitted key never reaches the constructor.

## Behavior change

Deny-list matches from YAML registry configs without an explicit `deny_list_score` change from `0.0` to `1.0`. Anyone relying on the `0.0` default should set `deny_list_score` explicitly. The change must be called out in the PR description and the docs (`docs/analyzer/recognizer_registry_provider.md`).

## Environment

`main` at 5e2fcea9, presidio-analyzer.

L4XB:
## Change Description

A custom deny-list recognizer that omits `deny_list_score` currently gets a different score depending on how it is loaded: `PatternRecognizer(...)` and `RecognizerRegistry.add_pattern_recognizer_from_dict(...)` give the constructor default `1.0`, while `RecognizerRegistryProvider` (YAML / dict registry config) gives `0.0`, because `CustomRecognizerConfig.deny_list_score` declared `default=0.0` and that default was dumped and passed to the recognizer. A deny-list match with score `0.0` is filtered by any positive `score_threshold`, so a YAML deny list without an explicit score never fired through the provider path.

This PR:

- changes `CustomRecognizerConfig.deny_list_score` to `Optional[float] = Field(default=None, ge=0.0, le=1.0)`;
- gives `CustomRecognizerConfig` the same `model_dump(exclude_none=True)` override the other pass-through config models already have, so an omitted field is left out of the kwargs that reach `PatternRecognizer.from_dict` and the constructor default applies. Other `Optional` fields on the custom config (`patterns`, `context`, `supported_language`, `country_code`, ...) already default to `None` in the recognizer constructor, and the loader reads them with `.get(...)`, so leaving them out of the dump does not change their handling;
- documents the default in `docs/analyzer/recognizer_registry_provider.md`.

**Behavior change (existing users):** a registry config (YAML or dict) whose custom recognizer has a `deny_list` but no `deny_list_score` now yields deny-list matches with score `1.0` instead of `0.0`, i.e. those terms are now detected instead of being dropped below the threshold. Configs that set `deny_list_score` explicitly are unchanged. Detection patterns and scores of the predefined recognizers are not touched.

Tests:

- model-level: an omitted `deny_list_score` is `None` and absent from the dump; an explicit value is dumped unchanged;
- through `RecognizerRegistryProvider` (new-style entry and legacy `supported_language` entry): the constructed recognizer has `deny_list_score == 1.0`, its deny-list pattern has score `1.0`, the same as an in-code `PatternRecognizer`, and `analyze("Dear Mr. Smith")` returns the match with score `1.0`; an explicit `deny_list_score: 0.4` is honored.

Ran `uv run pytest tests/test_yaml_recognizer_models.py tests/test_recognizer_registry_provider.py` (104 passed; the two `*_chunker_config` tests fail on `main` too in an environment without the `transformers` extra) and `ruff check` / `ruff format --check` on the changed source file.

## Issue reference

Fixes #2242

## Checklist

### DIFF
--- a/docs/analyzer/recognizer_registry_provider.md
+++ b/docs/analyzer/recognizer_registry_provider.md
@@ -109,7 +109,7 @@ The recognizer list comprises of both the predefined and custom recognizers, for
   - `enabled`: enables or disables the recognizer.
   - `supported_entity`: the detected entity associated by the recognizer.
   - `deny_list`: A list of words to detect, in case the recognizer uses a predefined list of words.
-  - `deny_list_score`: confidence score for a term identified using a deny-list.
+  - `deny_list_score`: confidence score for a term identified using a deny-list. Optional; when omitted, the `PatternRecognizer` default of `1.0` is used, the same as for recognizers created in code.
   - `score_thresholds`: optional score thresholds for this recognizer. Use `default` as the recognizer-wide threshold and entity names for overrides. Note that supplying `analyzer_engine.analyze(score_threshold=...)` bypasses recognizer-level thresholds for that request. The precedence is: Presidio Analyzer analyzer.analyze(score_threshold=...) > an entity specific threshold > a recognizer default threshold (`default`) > the Presidio Analyzer `default_score_threshold`.
   - `text_chunker`: configures how long texts are split for NER recognizers (`GLiNERRecognizer`, `HuggingFaceNerRecognizer`). Accepts a dict with `chunker_type` and params. Available types: `character` (default) and `tokenizer` (uses the model's tokenizer for accurate token-based splitting). Example:
 
--- a/presidio-analyzer/presidio_analyzer/input_validation/yaml_recognizer_models.py
+++ b/presidio-analyzer/presidio_analyzer/input_validation/yaml_recognizer_models.py
@@ -365,7 +365,14 @@ class CustomRecognizerConfig(BaseRecognizerConfig):
         default=None, description="Words to deny/exclude"
     )
     deny_list_score: Optional[float] = Field(
-        default=0.0, ge=0.0, le=1.0, description="Deny list score"
+        default=None,
+        ge=0.0,
+        le=1.0,
+        description=(
+            "Score for deny-list matches. When omitted, the PatternRecognizer "
+            "default (1.0) applies, as it does for recognizers created in code "
+            "or through add_pattern_recognizer_from_dict"
+        ),
     )
 
     # Language validation (legacy and new formats)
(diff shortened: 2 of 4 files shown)


L4XB:
#2254 fixed #2242 on `main` with the same change in the same four files: an omitted `deny_list_score` now takes `PatternRecognizer`'s default instead of `0.0`. So this one is superseded, and I'm closing it. Thanks!

PR #2248, opened 2026-09-10
Config hardening PRD: Recognizer configuration conformance suite (#1)

omri374:
## Change Description

Adds `tests/test_recognizer_config_conformance.py`, the conformance suite for the recognizer registry / YAML configuration layer (ADR module M0): a parametrized check that every concrete recognizer's constructor accepts the keys `RecognizerListLoader` injects from a registry YAML entry (`name`, `supported_language`, `context`, and one of `supported_entity`/`supported_entities`), a field-reach test over every entry of the shipped `conf/default_recognizers.yaml`, and a per-class round-trip test that builds a synthetic registry entry for every concrete recognizer class and asserts `name`/`context`/`score_thresholds` actually reach the constructed instance. It also fixes, in the loader, the registry-build crash the first of these tests found (a `context` entry on a recognizer whose constructor does not accept it), and adds `logger.warning` calls for two previously-silent cases the tests surfaced.

### Behavior changes

None to detection results or scores. Two new `logger.warning` lines, and one `TypeError` that becomes a warning:

- `RecognizerListLoader._prepare_recognizer_kwargs` now drops `context` when the recognizer class does not accept it, and logs a WARNING naming the class and the key. Before this PR the same entry raised `TypeError: __init__() got an unexpected keyword argument 'context'` and the whole registry failed to build. Nothing changes for classes that accept `context`.
- The loader already silently dropped `supported_entity`/`supported_entities` for a class whose constructor accepts neither (e.g. `BasicLangExtractRecognizer`, which derives its entities from its own config file). That case now logs a WARNING naming the class and the dropped key instead of staying silent.

No entity or PII values are logged, only class names and key names. No constructor signature changes.

### Bare-string entries now build a recognizer

`recognizers: [CreditCardRecognizer]` is the shorthand the configuration schema accepts for a predefined recognizer with no further settings. `_split_recognizers` required a mapping for the predefined list and excluded strings from the custom list, so the entry matched neither and **no recognizer was built**: the registry came back empty with no error, and a mixed list silently lost only its string entries. A bare string is now expanded to `{name: <str>, type: predefined}` at that single chokepoint.

Behavior changes:

- A bare-string entry now constructs the named recognizer instead of being silently discarded. A registry using this form gains the recognizers it already asked for.
- A bare-string entry naming an unknown class now raises `PredefinedRecognizerNotFoundError`, matching what a `name:` in a mapping entry already did, instead of yielding an empty registry with no diagnostic.
- Mapping entries are unaffected, and the shipped `default_recognizers.yaml` has no bare-string entries, so the default registry is unchanged.

Five tests cover the path (split normalization, construction, one instance per registry language, mixed with mapping entries, unknown name raising); all five were verified to fail without the fix. Four more pin that YAML-defined `type: custom` recognizers are untouched by the predefined-path rules this PR adds: their fields survive and none of the new warnings fire, an entry without a `type` key is still custom, and a multi-language custom entry keeps per-language context.

### Why `context` is handled in the loader, not the constructors

`EntityRecognizer.context` is one flat list of context words applied to every result the recognizer emits. That is meaningful for a recognizer that detects a single entity type (the pattern recognizers), and meaningless for one that detects several: `MedicalNERRecognizer`, `AzureHealthDeidRecognizer` and the LangExtract recognizers would boost a DATE as much as a NAME. Those classes deliberately do not accept `context`, and this PR does not add it to them.

What was broken is that a registry entry setting `context` for such a class crashed the whole registry build. The fix is one rule in the loader: a `context` key the class does not accept is dropped with a warning naming the class. The conformance contract is therefore `name`, `supported_language` and an entity key; `context` is not part of it.

### What's in each commit / story

1. **Story 1** — `test_recognizer_accepts_registry_injected_keys`, parametrized over every concrete `EntityRecognizer` subclass, walking each constructor's `__init__` MRO to compute which registry-injected keys are reachable. `KNOWN_CONTRACT_GAPS` regression-locks the three gaps found (all `context`); `ENTITIES_FROM_OWN_CONFIG` documents the LangExtract family's legitimate exemption from the entity-key requirement.
2. **Story 2** — the loader-side `context` rule described above plus its two tests (`test_context_dropped_with_warning_for_class_not_accepting_it`, `test_context_kept_without_warning_for_class_accepting_it`), and the dropped-entity-key warning plus its two tests (`test_dropped_entity_key_warns_for_class_defining_its_own_entities`, `test_no_warning_when_class_accepts_the_entity_key`), all in `test_recognizers_loader_utils.py`. `KNOWN_CONTRACT_GAPS` is empty: every concrete recognizer already meets the `name` / `supported_language` / entity-key contract.
3. **Story 3** — `test_shipped_entry_fields_reach_constructed_recognizer`, parametrized over every entry of `conf/default_recognizers.yaml`: loads each through `RecognizerRegistryProvider` and asserts language count, `supported_language`, per-language `context`, `name`, and (when set) `supported_entities`/`score_thresholds` all reach the constructed instance(s). Reuses `NOT_LOADABLE_FROM_SHIPPED_ENTRY` from `tests/test_recognizers_loader_utils.py` (imported, not duplicated).
4. **Story 4** — `test_synthetic_entry_round_trips_to_every_concrete_class`, parametrized over every concrete recognizer class with a synthetic `conf_<Class>` entry; `REQUIRED_KWARGS`/`REQUIRED_ENV` supply what a handful of classes need to construct (an Azure endpoint/credentials, `HuggingFaceNerRecognizer`'s `model_name`); `NOT_LOADABLE_AS_PREDEFINED_ENTRY` documents three classes that structurally cannot be built through a `type: predefined` entry at all (`LocalRecognizer`, `PatternRecognizer`, `ZaPhoneNumberRecognizer` — all subclassing-only base classes); for a class that does not accept `context` the test asserts the entry still loads, the instance keeps the base default `[]`, and the loader WARNING names the class; `CONTEXT_NOT_APPLIED` documents `BasicLangExtractRecognizer`'s pre-existing (unchanged) "accepts but doesn't apply `context`" behavior. `test_unknown_key_is_not_silent` documents today's silent-drop gap for an unrecognized YAML key and is marked `xfail(strict=True, reason="flipped in turn 06 (derived schema)")` — it is expected to fail today; that is the point, and it will force the marker's removal the moment turn 06 closes the gap.

### Notes / known limitations

- Commit `a2106de` added `context` to three constructors; commit `45c8a3b` reverts that and moves the fix into the loader, per review. The net diff contains no constructor changes.
- Commit `ecd8a19` addresses the second Copilot review round:
  - **Optional extras no longer hard-fail.** Six classes import `transformers`, the Azure SDKs or `langextract` in their constructor, so on a core/dev install without `--all-extras` the suite raised `ImportError` instead of skipping. `OPTIONAL_DEPENDENCY_MODULES` now skips those parametrized cases when the module is absent. The skip probes for the module rather than catching the exception, so an unexpected `ImportError` from any class still fails. Verified with those modules hidden: the 7 previously failing cases skip and 302 pass.
  - **A `context` carve-out for `**kwargs` leaves was added here and then reverted as unsound.** The intent was to keep `context` for a leaf constructor accepting `**kwargs`, on the theory that such a class could read the key itself. That is wrong: `_reachable_init_param_names` already follows `**kwargs` up the MRO, so "unreachable" means the forwarding chain breaks at a class that does not declare `context`, and leaving the key in raises `TypeError` at that parent -- the exact crash this rule exists to prevent. Reproduced directly against the `ChildForwardsKwargs` -> `StrictParent` test double. The drop is unconditional when `context` is unreachable, and the regression test now constructs the class rather than only inspecting the prepared kwargs.
  - **`_entry_language_configs` now mirrors the loader** on the no-`supported_languages` path, carrying the entry-level `context` into each fallback language instead of `None`. Eleven shipped entries take that path, so the test would have stopped checking context if one ever set the key. An empty language list is now rejected explicitly rather than treated as omitted.
  - Corrected the comment claiming the branch leaves the entity key in `kwargs`: `supported_entity` is always filtered out later when not declared.

### Known gaps, deliberately left open

- **Entry-level `context` is silently discarded when `supported_languages` is a bare list of language codes.** `BaseRecognizerConfig` accepts that shape, but only the per-language `{language, context}` form reaches the recognizer. `test_entry_context_is_dropped_for_bare_language_list` pins today's behavior rather than changing the loader, which would alter detection for anyone whose YAML relies on the current result. No shipped entry combines the two keys.
- **The synthetic round-trip entry does not set an entity key**, so the loader's `supported_entity`/`supported_entities` path is covered at the signature level but not end to end. Closing it needs a valid entity name per class, which is a larger change than this turn should carry.

- Commit `6b5233b2` addresses the last Copilot review comment: the shipped-entry test gated its `supported_entities` assertion on truthiness, so an entry explicitly setting `supported_entities: []` was treated as "not set" and skipped. It now gates on key presence. Verified separately that `supported_entities: []` does **not** round-trip today (the recognizer falls back to its own default entity list), so the previous truthiness guard would have hidden exactly the kind of silent-drop this suite exists to catch. No shipped entry sets the key, so nothing changes for the current configuration. The same commit renames `dropped_keys` to `ineffective_keys` and renames `test_dropped_entity_key_warns_*` to `test_ineffective_entity_key_warns_*`, because that branch warns without removing the key.
- Commits `7cd44e3` and `31fa9b0` address the Copilot review comment: both loader warnings and the `context` drop now key off `RecognizerListLoader._reachable_init_param_names` (constructor parameters reachable through `**kwargs` forwarding along the MRO), so `StanzaRecognizer` and `TransformersRecognizer`, which forward to `SpacyRecognizer`, are not reported as ignoring keys they apply.

- `uv run ruff format --check .` flags 26 pre-existing files unrelated to this PR (confirmed identical on `origin/main` before this branch's changes); the actual CI lint gate (`.github/workflows/ci.yml`) only runs `ruff check`, which is fully green. No file touched by this PR needs reformatting.
- Full local run with `uv sync --locked --all-extras --group dev`: `3772 passed, 13 skipped, 1 xfailed`. `ruff check` and `ruff format --check` are clean on all three touched files. Without the optional extras installed, 7 of the new conformance cases fail with `ImportError` rather than skipping (transformers, langextract, azure SDKs); CI installs all extras, so this only affects contributors running a partial environment.

## Issue reference

Part of the ADR at https://github.com/data-privacy-stack/presidio-product-core/issues/139.

Turn 02 of 6. Turn 01 is PR #2210.

## Checklist

- [x] My PR contains documentation updates / additions if required (none needed: no new entities, no schema changes)

https://claude.ai/code/session_014cYn8gUDL2txRa5V7r4vLZ

---
_Generated by [Claude Code](https://claude.ai/code/session_014cYn8gUDL2txRa5V7r4vLZ)_

### DIFF
--- a/presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
+++ b/presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
@@ -321,6 +321,23 @@ class RecognizerListLoader:
         accepts_supported_entity = RecognizerListLoader.SUPPORTED_ENTITY in params
         accepts_supported_entities = RecognizerListLoader.SUPPORTED_ENTITIES in params
 
+        # ``context`` is one flat word list applied to every result a recognizer
+        # emits, so it only makes sense for recognizers that detect a single
+        # entity type. Recognizers that detect several (NER models, remote PHI
+        # services, LLM extractors) deliberately do not accept it. When a
+        # registry entry sets context for such a class, drop it with a warning
+        # instead of letting the constructor raise ``TypeError`` and take the
+        # whole registry down.
+        if "context" in kwargs and "context" not in params and not has_var_kw:
+            kwargs.pop("context")
+            logger.warning(
+                "%s does not accept 'context'; ignoring the context words "
+                "configured for it. Context words boost every result a "
+                "recognizer emits, so recognizers that detect several entity "
+                "types do not support them.",
+                recognizer_cls.__name__,
+            )
+
         # A class that accepts neither key defines its entities itself (e.g. from
         # a config file, as LangExtract-based recognizers do) rather than from the
         # registry entry. Warn -- rather than silently dropping the value -- when
--- a/presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
+++ b/presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
@@ -269,30 +269,43 @@ class RecognizerListLoader:
         )
 
     @staticmethod
-    def _reachable_init_param_names(cls: Type) -> Set[str]:
-        """
-        Union of ``__init__`` parameter names reachable via **kwargs forwarding.
+    def _reachable_init_param_names(recognizer_cls: Type[EntityRecognizer]) -> Set[str]:
+        """Return the constructor parameter names reachable through the MRO.
 
-        Walks ``cls``'s constructor MRO, accumulating each ``__init__``'s
-        parameter names, and stops after the first ``__init__`` that has no
-        ``**kwargs`` -- a keyword argument can't be forwarded past that point,
-        so any name declared further up is unreachable from ``cls``.
+        Walks ``recognizer_cls.__mro__``. Each class that defines its own
+        ``__init__`` contributes its parameter names (excluding ``self`` and
+        the ``*args`` / ``**kwargs`` slots). Traversal stops after the first
+        ``__init__`` that does not accept ``**kwargs``: once a constructor
+        stops forwarding keyword arguments, a parameter declared only further
+        up the chain can no longer be reached from a registry entry.
+
+        :param recognizer_cls: The recognizer class to inspect.
         """
-        reachable: Set[str] = set()
-        for klass in cls.__mro__:
+        names: Set[str] = set()
+        for klass in recognizer_cls.__mro__:
             init = klass.__dict__.get("__init__")
             if init is None:
                 continue
             try:
-                params = inspect.signature(init).parameters
+                parameters = inspect.signature(init).parameters
             except (TypeError, ValueError):
                 break
-            reachable.update(params)
+            names.update(
+                name
+                for name, param in parameters.items()
+                if name != "self"
+                and param.kind
+                not in (
+                    inspect.Parameter.VAR_KEYWORD,
+                    inspect.Parameter.VAR_POSITIONAL,
+                )
+            )
             if not any(
-                p.kind == inspect.Parameter.VAR_KEYWORD for p in params.values()
+                param.kind == inspect.Parameter.VAR_KEYWORD
+                for param in parameters.values()
             ):
                 break
-        return reachable
+        return names
 
     @staticmethod
     def _prepare_recognizer_kwargs(
@@ -347,6 +360,14 @@ class RecognizerListLoader:
         accepts_supported_entity = RecognizerListLoader.SUPPORTED_ENTITY in params
         accepts_supported_entities = RecognizerListLoader.SUPPORTED_ENTITIES in params
 
+        # Parameters reachable through the constructor chain: the leaf signature
+        # plus, while each ``__init__`` forwards ``**kwargs``, its parents'. A
+        # subclass such as StanzaRecognizer accepts ``supported_entities`` and
+        # ``context`` through SpacyRecognizer even though its own signature
+        # names neither, so the two warnings below key off this set and not the
+        # leaf signature alone.
+        reachable = RecognizerListLoader._reachable_init_param_names(recognizer_cls)
+
         # ``context`` is one flat word list applied to every result a recognizer
         # emits, so it only makes sense for recognizers that detect a single
         # entity type. Recognizers that detect several (NER models, remote PHI
@@ -354,7 +375,7 @@ class RecognizerListLoader:
         # registry entry sets context for such a class, drop it with a warning
         # instead of letting the constructor raise ``TypeError`` and take the
         # whole registry down.
-        if "context" in kwargs and "context" not in params and not has_var_kw:
+        if "context" in kwargs and "context" not in reachable:
             kwargs.pop("context")
             logger.warning(
                 "%s does not accept 'context'; ignoring the context words "
@@ -364,26 +385,24 @@ class RecognizerListLoader:
                 recognizer_cls.__name__,
             )
 
-        # A class that accepts neither key on its own __init__ *usually* defines
-        # its entities itself (e.g. from a config file, as LangExtract-based
-        # recognizers do) rather than from the registry entry -- but a class
-        # that forwards **kwargs to a base class which does declare the key
-        # (e.g. TransformersRecognizer/StanzaRecognizer forwarding to
-        # SpacyRecognizer) genuinely applies it further up the chain. Only warn
-        # when the key is unreachable through the whole **kwargs-forwarding
-        # MRO chain, so the message is accurate: it never claims a value is
-        # ignored when some base class will actually consume it.
-        if not accepts_supported_entity and not accepts_supported_entities:
-            reachable_params = RecognizerListLoader._reachable_init_param_names(
-                recognizer_cls
-            )
+        # A class that accepts neither key anywhere in its constructor chain
+        # defines its entities itself (e.g. from a config file, as
+        # LangExtract-based recognizers do) rather than from the registry entry.
+        # Warn -- rather than silently dropping the value -- when the entry
+        # actually tried to set one, so a user relying on it finds out why it
+        # had no effect instead of debugging a mismatch later.
+        entity_key_reachable = (
+            RecognizerListLoader.SUPPORTED_ENTITY in reachable
+            or RecognizerListLoader.SUPPORTED_ENTITIES in reachable
+        )
+        if not entity_key_reachable:
             dropped_keys = [
                 key
                 for key in (
                     RecognizerListLoader.SUPPORTED_ENTITY,
                     RecognizerListLoader.SUPPORTED_ENTITIES,
                 )
-                if key in kwargs and key not in reachable_params
+                if key in kwargs
             ]
             if dropped_keys:
                 logger.warning(
--- a/presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
+++ b/presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
@@ -334,10 +334,11 @@ class RecognizerListLoader:
             - supported_entities: kept if explicitly accepted or if the recognizer
               accepts **kwargs.
 
-        If the class accepts neither key (it defines its supported entities from
-        its own configuration, e.g. a LangExtract config file) and the entry set
-        one anyway, a ``logger.warning`` names the class and the dropped key
-        instead of silently discarding it.
+        If the key is unreachable anywhere in the class's constructor chain (it
+        defines its supported entities from its own configuration, e.g. a
+        LangExtract config file) and the entry set one anyway, a
+        ``logger.warning`` names the class and the key that has no effect,
+        instead of staying silent about it.
         """
         kwargs = {**recognizer_conf, **language_conf}
 
--- a/presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
+++ b/presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
@@ -289,6 +289,11 @@ class RecognizerListLoader:
             - supported_entity: kept only if explicitly accepted.
             - supported_entities: kept if explicitly accepted or if the recognizer
               accepts **kwargs.
+
+        If the class accepts neither key (it defines its supported entities from
+        its own configuration, e.g. a LangExtract config file) and the entry set
+        one anyway, a ``logger.warning`` names the class and the dropped key
+        instead of silently discarding it.
         """
         kwargs = {**recognizer_conf, **language_conf}
 
@@ -315,6 +320,30 @@ class RecognizerListLoader:
 
         accepts_supported_entity = RecognizerListLoader.SUPPORTED_ENTITY in params
         accepts_supported_entities = RecognizerListLoader.SUPPORTED_ENTITIES in params
+
+        # A class that accepts neither key defines its entities itself (e.g. from
+        # a config file, as LangExtract-based recognizers do) rather than from the
+        # registry entry. Warn -- rather than silently dropping the value -- when
+        # the entry actually tried to set one, so a user relying on it finds out
+        # why it had no effect instead of debugging a mismatch later.
+        if not accepts_supported_entity and not accepts_supported_entities:
+            dropped_keys = [
+                key
+                for key in (
+                    RecognizerListLoader.SUPPORTED_ENTITY,
+                    RecognizerListLoader.SUPPORTED_ENTITIES,
+                )
+                if key in kwargs
+            ]
+            if dropped_keys:
+                logger.warning(
+                    "%s does not accept 'supported_entity' or 'supported_entities'; "
+                    "ignoring %s from its configuration because %s defines its "
+                    "supported entities from its own configuration.",
+                    recognizer_cls.__name__,
+                    " and ".join(dropped_keys),
+                    recognizer_cls.__name__,
+                )
 
         # 1. Normalize: Convert plural -> singular if needed
         # (Only when singular is accepted and plural is NOT accepted)
--- a/presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
+++ b/presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
@@ -318,6 +318,11 @@ class RecognizerListLoader:
 
         This function adapts supported_entity/supported_entities based on the
         recognizer class __init__ signature to avoid passing unexpected kwargs.
+        Note: a key can remain present in the returned kwargs while still being
+        effectively ignored by the constructed recognizer -- e.g. a class that
+        accepts **kwargs but never reads ``supported_entities`` from it. This
+        function only controls what reaches the constructor call, not whether
+        the constructor uses it.
 
         - If recognizer accepts only supported_entity (singular), convert
           supported_entities -> supported_entity (first element).
@@ -406,7 +411,7 @@ class RecognizerListLoader:
             ]
             if dropped_keys:
                 logger.warning(
-                    "%s does not accept 'supported_entity' or 'supported_entities'; "
+                    "%s does not apply 'supported_entity' or 'supported_entities'; "
                     "ignoring %s from its configuration because %s defines its "
                     "supported entities from its own configuration.",
                     recognizer_cls.__name__,
--- a/presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
+++ b/presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
@@ -269,6 +269,32 @@ class RecognizerListLoader:
         )
 
     @staticmethod
+    def _reachable_init_param_names(cls: Type) -> Set[str]:
+        """
+        Union of ``__init__`` parameter names reachable via **kwargs forwarding.
+
+        Walks ``cls``'s constructor MRO, accumulating each ``__init__``'s
+        parameter names, and stops after the first ``__init__`` that has no
+        ``**kwargs`` -- a keyword argument can't be forwarded past that point,
+        so any name declared further up is unreachable from ``cls``.
+        """
+        reachable: Set[str] = set()
+        for klass in cls.__mro__:
+            init = klass.__dict__.get("__init__")
+            if init is None:
+                continue
+            try:
+                params = inspect.signature(init).parameters
+            except (TypeError, ValueError):
+                break
+            reachable.update(params)
+            if not any(
+                p.kind == inspect.Parameter.VAR_KEYWORD for p in params.values()
+            ):
+                break
+        return reachable
+
+    @staticmethod
     def _prepare_recognizer_kwargs(
         recognizer_conf: Dict[str, Any],
         language_conf: Dict[str, Any],
@@ -338,19 +364,26 @@ class RecognizerListLoader:
                 recognizer_cls.__name__,
             )
 
-        # A class that accepts neither key defines its entities itself (e.g. from
-        # a config file, as LangExtract-based recognizers do) rather than from the
-        # registry entry. Warn -- rather than silently dropping the value -- when
-        # the entry actually tried to set one, so a user relying on it finds out
-        # why it had no effect instead of debugging a mismatch later.
+        # A class that accepts neither key on its own __init__ *usually* defines
+        # its entities itself (e.g. from a config file, as LangExtract-based
+        # recognizers do) rather than from the registry entry -- but a class
+        # that forwards **kwargs to a base class which does declare the key
+        # (e.g. TransformersRecognizer/StanzaRecognizer forwarding to
+        # SpacyRecognizer) genuinely applies it further up the chain. Only warn
+        # when the key is unreachable through the whole **kwargs-forwarding
+        # MRO chain, so the message is accurate: it never claims a value is
+        # ignored when some base class will actually consume it.
         if not accepts_supported_entity and not accepts_supported_entities:
+            reachable_params = RecognizerListLoader._reachable_init_param_names(
+                recognizer_cls
+            )
             dropped_keys = [
                 key
                 for key in (
                     RecognizerListLoader.SUPPORTED_ENTITY,
                     RecognizerListLoader.SUPPORTED_ENTITIES,
                 )
-                if key in kwargs
+                if key in kwargs and key not in reachable_params
             ]
             if dropped_keys:
                 logger.warning(
--- a/presidio-analyzer/tests/test_recognizer_config_conformance.py
+++ b/presidio-analyzer/tests/test_recognizer_config_conformance.py
@@ -320,3 +320,197 @@ def test_shipped_entry_fields_reach_constructed_recognizer(
         expected_thresholds = normalize_score_thresholds(entry["score_thresholds"])
         for instance in instances:
             assert instance.score_thresholds == expected_thresholds
+
+
+# ---------------------------------------------------------------------------
+# Story 4: per-class round-trip and no-silent-drop tests
+# ---------------------------------------------------------------------------
+
+# Constructor kwargs a class needs beyond what the synthetic entry below
+# already supplies, keyed by class name. Most concrete recognizers construct
+# fine from the synthetic entry's fields alone (constructor defaults handle
+# the rest); these are the exceptions that need a value with no usable
+# default.
+REQUIRED_KWARGS: Dict[str, Dict[str, Any]] = {
+    # No default endpoint; the constructor raises ValueError without one.
+    "AzureOpenAILangExtractRecognizer": {
+        "azure_endpoint": "https://example-resource.openai.azure.com/"
+    },
+    # HuggingFaceNerRecognizer.load() requires model_name. load() is patched
+    # to a no-op for this test (see `patched_loads`), so this is not
+    # strictly required for construction to succeed -- supplied anyway for a
+    # realistic entry, per this dict's own purpose.
+    "HuggingFaceNerRecognizer": {"model_name": "dslim/bert-base-NER"},
+}
+
+# Environment variables a class's constructor reads directly (not via a
+# registry YAML kwarg) and needs set to construct without real credentials.
+REQUIRED_ENV: Dict[str, Dict[str, str]] = {
+    # AzureHealthDeidRecognizer needs either a `client` instance or these two
+    # env vars to build its default Azure client. `client` cannot be
+    # supplied through a registry entry -- the schema
+    # (`PredefinedRecognizerConfig`) has no `client` field and silently
+    # drops unknown keys -- so the env-var path is the only one reachable
+    # from configuration, and the only one this test can exercise.
+    "AzureHealthDeidRecognizer": {
+        "AHDS_ENDPOINT": "https://fake.endpoint.example",
+        "ENV": "development",
+    },
+    # Same shape as AzureHealthDeidRecognizer above: azure_ai_key /
+    # azure_ai_endpoint are real constructor parameters, but
+    # AzureAILanguageRecognizer also has no dedicated entry in
+    # CONFIG_MODEL_MAP, so PredefinedRecognizerConfig's default
+    # extra="ignore" drops them if passed as registry kwargs. The
+    # constructor's own env-var fallback is the only reachable path.
+    "AzureAILanguageRecognizer": {
+        "AZURE_AI_KEY": "fake-key",
+        "AZURE_AI_ENDPOINT": "https://fake.endpoint.example",
+    },
+}
+
+# Classes whose constructor accepts `context` (so they pass the Story 1/2
+# contract check) but do not apply it to the constructed instance. Pre-
+# existing behavior this turn does not change -- see
+# BasicLangExtractRecognizer's own docstring ("context ... optional,
+# currently not used by LLM recognizers"). The round-trip test below skips
+# only the context assertion for these classes.
+CONTEXT_NOT_APPLIED = {"BasicLangExtractRecognizer"}
+
+# Classes that cannot be constructed through a `type: predefined` registry
+# entry at all, even with REQUIRED_KWARGS -- not a config-layer gap, but a
+# structural mismatch this suite cannot paper over:
+NOT_LOADABLE_AS_PREDEFINED_ENTRY = {
+    # Requires `supported_entities` positionally and implements neither
+    # `analyze` nor a real `load` -- a base class for subclassing
+    # (SpacyRecognizer, PatternRecognizer, ...), never meant to be named
+    # directly in configuration.
+    "LocalRecognizer",
+    # Requires `supported_entity` and (`patterns` or `deny_list`)
+    # positionally. The last two can only be set through a registry entry
+    # with `type: custom` -- `RecognizerRegistryConfig.parse_recognizers`
+    # rejects `patterns`/`deny_list` on a `type: predefined` entry outright
+    # ("... is marked as 'predefined' but contains 'patterns' or
+    # 'deny_list' ..."), so no predefined entry can ever supply them.
+    "PatternRecognizer",
+    # Requires `target_classification` positionally, with no default and no
+    # schema field to set it from (base `PredefinedRecognizerConfig` has
+    # none, and no CONFIG_MODEL_MAP entry adds one) -- a base class for
+    # subclassing (ZaMobileNumberRecognizer, ZaTelephoneNumberRecognizer),
+    # never meant to be named directly in configuration.
+    "ZaPhoneNumberRecognizer",
+}
+
+
+def _synthetic_entry(cls: Type[EntityRecognizer]) -> Dict[str, Any]:
+    return {
+        "name": f"conf_{cls.__name__}",
+        "class_name": cls.__name__,
+        "type": "predefined",
+        "supported_languages": [{"language": "en", "context": ["zeta"]}],
+        "score_thresholds": {"default": 0.42},
+        **REQUIRED_KWARGS.get(cls.__name__, {}),
+    }
+
+
+ROUND_TRIP_CLASSES = [
+    cls
+    for cls in CONCRETE_RECOGNIZER_CLASSES
+    if cls.__name__ not in NOT_LOADABLE_AS_PREDEFINED_ENTRY
+]
+
+
+def test_not_loadable_as_predefined_entry_names_only_real_classes():
+    """Guard ``NOT_LOADABLE_AS_PREDEFINED_ENTRY`` against stale entries."""
+    names = {cls.__name__ for cls in CONCRETE_RECOGNIZER_CLASSES}
+    assert NOT_LOADABLE_AS_PREDEFINED_ENTRY <= names
+
+
+@pytest.mark.parametrize("cls", ROUND_TRIP_CLASSES, ids=lambda cls: cls.__name__)
+def test_synthetic_entry_round_trips_to_every_concrete_class(
+    cls, patched_loads, monkeypatch
+):
+    """A synthetic registry entry round-trips to every concrete class.
+
+    Proves, for every concrete recognizer class -- not only the ones shipped
+    enabled in ``default_recognizers.yaml`` -- that ``name``, per-language
+    ``context``, and ``score_thresholds`` set in a registry entry actually
+    reach the constructed instance, not just that the constructor accepts
+    the keys (Story 1/2 already covers acceptance at the signature level).
+    """
+    for name, value in REQUIRED_ENV.get(cls.__name__, {}).items():
+        monkeypatch.setenv(name, value)
+
+    entry = _synthetic_entry(cls)
+    configuration = {
+        "global_regex_flags": 26,
+        "supported_languages": ["en"],
+        "recognizers": [entry],
+    }
+
+    registry = RecognizerRegistryProvider(
+        registry_configuration=configuration
+    ).create_recognizer_registry()
+
+    instances = [r for r in registry.recognizers if type(r) is cls]
+    assert len(instances) == 1, (
+        f"{cls.__name__}: expected exactly one constructed instance, got "
+        f"{len(instances)}"
+    )
+    instance = instances[0]
+
+    assert instance.name == f"conf_{cls.__name__}"
+    assert instance.supported_language == "en"
+    if cls.__name__ not in CONTEXT_NOT_APPLIED:
+        assert instance.context == ["zeta"], (
+            f"{cls.__name__}: context did not reach the constructed instance"
+        )
+    assert instance.score_thresholds == {"default": 0.42}
+
+
+def test_context_not_applied_names_only_real_classes():
+    """Guard ``CONTEXT_NOT_APPLIED`` against stale entries."""
+    names = {cls.__name__ for cls in CONCRETE_RECOGNIZER_CLASSES}
+    assert CONTEXT_NOT_APPLIED <= names
+
+
+@pytest.mark.xfail(strict=True, reason="flipped in turn 06 (derived schema)")
+def test_unknown_key_is_not_silent(caplog):
+    """An unknown registry key must never be silently dropped.
+
+    Today it is: ``PredefinedRecognizerConfig`` (the schema
+    ``CreditCardRecognizer`` and most predefined recognizers validate
+    against) leaves pydantic's default ``extra="ignore"``, so an
+    unrecognized key such as ``no_such_key`` disappears at parse time with
+    no error and no log line -- exactly the silent-drop failure mode this
+    whole suite exists to catch. Turn 06 (derived schema) is expected to
+    make this fail loudly instead; ``xfail(strict=True)`` means this test
+    starts *failing* the moment that happens, forcing the xfail marker to be
+    removed rather than silently masking the fix forever.
+    """
+    configuration = {
+        "global_regex_flags": 26,
+        "supported_languages": ["en"],
+        "recognizers": [
+            {
+                "name": "CreditCardRecognizer",
+                "type": "predefined",
+                "supported_language": "en",
+                "no_such_key": 1,
+            }
+        ],
+    }
+
+    raised = False
+    with caplog.at_level("WARNING", logger="presidio-analyzer"):
+        try:
+            RecognizerRegistryProvider(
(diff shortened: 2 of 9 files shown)


presidio-analyzer/tests/test_recognizer_config_conformance.py
```diff
@@ -0,0 +1,516 @@
...
+  round-trips its ``name``, ``context``, and ``score_thresholds`` -- and an
+  unknown key is never silently dropped.
+"""
+
+from __future__ import annotations
+
+import inspect
+from typing import Any, Dict, List, Set, Tuple, Type
+
+import presidio_analyzer.predefined_recognizers  # noqa: F401 -- see below
```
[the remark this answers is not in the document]
omri374: This is an intentional side-effect import, not dead code: it registers every concrete predefined-recognizer subclass on `EntityRecognizer` before `get_all_existing_recognizers()` is called later in the module, which is what this conformance suite parametrizes over. It's already marked `# noqa: F401` with an explanatory comment on the next lines, and `ruff check` (the CI lint gate) passes. Removing it would break test collection for Stories 1 and 4. Not making a change here.

---

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
```diff
@@ -316,6 +321,30 @@ def _prepare_recognizer_kwargs(
...
+            ]
+            if dropped_keys:
+                logger.warning(
+                    "%s does not accept 'supported_entity' or 'supported_entities'; "
+                    "ignoring %s from its configuration because %s defines its "
+                    "supported entities from its own configuration.",
+                    recognizer_cls.__name__,
+                    " and ".join(dropped_keys),
+                    recognizer_cls.__name__,
+                )
```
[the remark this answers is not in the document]
omri374: Confirmed and fixed in 7cd44e3: added `RecognizerListLoader._reachable_init_param_names`, which walks the `**kwargs`-forwarding MRO chain (stopping at the first `__init__` without `**kwargs`, same rule the conformance suite's own reachability check uses) and gates the warning on that reachable set instead of the leaf signature alone. Verified directly against `TransformersRecognizer` (forwards to `SpacyRecognizer`, which declares `supported_entities`) — no warning fires and the value is applied — and added a regression test (`test_no_warning_when_kwargs_forwarding_reaches_a_declaring_parent`) using the existing `ChildForwardsKwargs`/`StrictParent` fixtures.

---

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
```diff
@@ -316,6 +347,54 @@ def _prepare_recognizer_kwargs(
...
```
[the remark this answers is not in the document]
omri374: Fixed in fd5836c: reworded to "does not apply 'supported_entity' or 'supported_entities'" — accurate whether the key is technically accepted via `**kwargs` and simply never consumed, or not accepted at all. Also clarified `_prepare_recognizer_kwargs`'s docstring to make the same distinction explicit (a key can remain in the returned kwargs while still being effectively ignored downstream), and swapped two hardcoded `global_regex_flags: 26` in the conformance tests for the already-imported `GLOBAL_REGEX_FLAGS` constant per the other two suggestions in this review.

---

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
```diff
@@ -289,6 +328,11 @@ def _prepare_recognizer_kwargs(
             - supported_entity: kept only if explicitly accepted.
             - supported_entities: kept if explicitly accepted or if the recognizer
               accepts **kwargs.
+
+        If the class accepts neither key (it defines its supported entities from
+        its own configuration, e.g. a LangExtract config file) and the entry set
+        one anyway, a ``logger.warning`` names the class and the dropped key
+        instead of silently discarding it.
```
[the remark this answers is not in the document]
omri374: Fixed in e1df81e: reworded to "if the key is unreachable anywhere in the class's constructor chain ... a logger.warning names the class and the key that has no effect" instead of "dropped".

---

presidio-analyzer/tests/test_recognizers_loader_utils.py
```diff
@@ -283,6 +286,155 @@ def test_uninspectable_signature_drops_entity_keys():
     assert "supported_entity" not in kwargs
 
 
+def test_dropped_entity_key_warns_for_class_defining_its_own_entities(caplog):
+    """A class that accepts neither supported_entity nor supported_entities
+    (it defines its entities from its own configuration, e.g. a LangExtract
+    config file) still loads when the entry sets supported_entities -- but a
+    WARNING naming the class and the dropped key is logged instead of
+    silently discarding the value.
+    """
```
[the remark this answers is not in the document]
omri374: Fixed in e1df81e: reworded to "has neither supported_entity nor supported_entities reachable anywhere in its constructor chain" and "the ineffective key is logged instead of staying silent about it" -- matches the test's own assertion, which confirms `kwargs["supported_entities"]` is still `["X"]` after the call.

---

presidio-analyzer/tests/test_recognizers_loader_utils.py
```diff
@@ -283,6 +286,155 @@ def test_uninspectable_signature_drops_entity_keys():
...
+            recognizer_cls=BasicLangExtractRecognizer,
+        )
+
+    warning_messages = [
+        r.getMessage() for r in caplog.records if r.levelname == "WARNING"
+    ]
+    assert any(
+        "BasicLangExtractRecognizer" in m and "supported_entities" in m
+        for m in warning_messages
+    ), f"expected a dropped-entity-key WARNING, got {warning_messages!r}"
```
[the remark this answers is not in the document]
omri374: Fixed in e1df81e: the failure message now reads "expected an ineffective-entity-key WARNING".

---

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
```diff
@@ -268,6 +268,45 @@ def get_existing_recognizer_cls(recognizer_name: str) -> Type[EntityRecognizer]:
...
+        """
+        names: Set[str] = set()
+        for klass in recognizer_cls.__mro__:
+            init = klass.__dict__.get("__init__")
+            if init is None:
+                continue
+            try:
+                parameters = inspect.signature(init).parameters
+            except (TypeError, ValueError):
+                break
```
omri374, line 305: why break? when is this expected to happen?

omri374: That's the `except (TypeError, ValueError): break` guard around `inspect.signature(init)` — it fires when a class's `__init__` can't be introspected at all (e.g. one assigned to a non-callable, as the `Uninspectable` test double in `test_recognizers_loader_utils.py` does with `__init__ = 123`, or an `__init__` implemented in C without a usable signature). It's not expected to happen for any real predefined recognizer today — every one of them has a plain Python `__init__` — this is a defensive fallback so the function degrades gracefully (returns whatever parameter names it already collected from the more-derived classes) instead of raising, if such a class is ever added or an unusual third-party base class is subclassed. Same pattern as the `except (TypeError, ValueError):` branch already used in `_prepare_recognizer_kwargs` just above it for the same reason.

---

omri374:
Heads up: the base branch changed from `main` to `config_hardening` (currently identical to `main`), and `.github/workflows/ci.yml` triggers only on `pull_request: branches: [main]`. Since the retarget, the full CI matrix (lint, CodeQL, Test Analyzer/Anonymizer/CLI/Structured/Image Redactor across all Python versions) hasn't run at all on this PR's latest commit (`6b5233b`) — only the unrelated `label` workflow fired. `mergeable_state` shows clean with no failing checks, but that's an absence of checks, not a pass.

If retargeting to `config_hardening` as a shared integration branch for the 6-turn ADR plan is intentional, this PR (and the others that land on it) won't get real CI coverage until either the workflow's branch trigger includes `config_hardening` or it's retargeted back to `main`. Flagging before merge rather than assuming — let me know which way you want this handled and I'll make the change.

---

presidio-analyzer/tests/test_recognizer_config_conformance.py
```diff
@@ -0,0 +1,557 @@
...
+
+SHIPPED_ENTRIES = _shipped_entries()
+# Entries that cannot construct from their shipped configuration alone (see
+# the set's own docstring/comment in tests/test_recognizers_loader_utils.py
+# for why) -- reused here rather than duplicated, so exactly one such set
+# exists in the test suite.
+LOADABLE_SHIPPED_ENTRIES = [
+    entry
+    for entry in SHIPPED_ENTRIES
+    if _entry_id(entry) not in NOT_LOADABLE_FROM_SHIPPED_ENTRY
```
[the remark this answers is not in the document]
omri374: Fixed in ecd8a19: reuses `OPTIONAL_DEPENDENCY_ENTRIES` from `test_recognizers_loader_utils.py` for this parametrization, matching the pattern in `test_yaml_entry_loads_when_enabled`. Verified with the optional modules made unimportable — `BasicLangExtractRecognizer` now skips here instead of failing.

---

presidio-analyzer/tests/test_recognizer_config_conformance.py
```diff
@@ -0,0 +1,557 @@
...
+    configuration = {
+        "global_regex_flags": GLOBAL_REGEX_FLAGS,
+        "supported_languages": ["en"],
+        "recognizers": [entry],
+    }
+
+    with caplog.at_level("WARNING", logger="presidio-analyzer"):
+        registry = RecognizerRegistryProvider(
+            registry_configuration=configuration
+        ).create_recognizer_registry()
```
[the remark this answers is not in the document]
omri374: Fixed in ecd8a19: adds `OPTIONAL_DEPENDENCY_MODULES`, a class-to-module skip policy, and probes for module presence directly rather than catching the exception, so an unexpected `ImportError` from any other class still fails. Verified with the relevant third-party packages made unimportable: exactly the 6 affected classes here (plus the shipped-entry case) skip, and every other case still passes or fails on its own merits.

---

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
```diff
@@ -316,6 +366,73 @@ def _prepare_recognizer_kwargs(
...
+        # it existed, a leaf constructor accepting ``**kwargs`` received
+        # ``context`` and could consume it without declaring the parameter.
+        # Dropping the key for those classes would be a silent loss of
+        # configuration for out-of-tree recognizers, so the drop is scoped to
+        # strict leaf signatures -- the only ones that raise TypeError on an
+        # unexpected keyword. Every in-repo recognizer that fails the
+        # reachability check has a strict leaf signature, so the shipped set
+        # behaves identically either way.
+        if "context" in kwargs and "context" not in reachable and not has_var_kw:
+            kwargs.pop("context")
```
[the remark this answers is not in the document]
omri374: Confirmed and fixed in 84d4159. Reproduced the crash directly first: `ChildForwardsKwargs(context=["zeta"])` raises `TypeError: StrictParent.__init__() got an unexpected keyword argument 'context'`, exactly as described. `has_var_kw` at the leaf is not a sound safety check — `_reachable_init_param_names` already assumes `**kwargs` is forwarded up the MRO while computing reachability, so "unreachable" means some class in that forwarding chain has no `**kwargs` of its own and doesn't declare `context`; keeping the value can't be proven safe in general.

Reverted to unconditionally dropping `context` when unreachable, regardless of the leaf's own `**kwargs` — registry-build safety over a hypothetical out-of-tree recognizer reading it from `**kwargs` without declaring it. Rewrote the regression test to actually construct `ChildForwardsKwargs(**kwargs)`, not just inspect the prepared dict, so this class of bug fails loudly next time.

---

presidio-analyzer/tests/test_recognizer_config_conformance.py
```diff
@@ -0,0 +1,670 @@
...
+# Classes that cannot be constructed through a `type: predefined` registry
+# entry at all, even with REQUIRED_KWARGS -- not a config-layer gap, but a
+# structural mismatch this suite cannot paper over:
+NOT_LOADABLE_AS_PREDEFINED_ENTRY = {
+    # Requires `supported_entities` positionally and implements neither
+    # `analyze` nor a real `load` -- a base class for subclassing
+    # (SpacyRecognizer, PatternRecognizer, ...), never meant to be named
+    # directly in configuration.
+    "LocalRecognizer",
+    # Requires `supported_entity` and (`patterns` or `deny_list`)
```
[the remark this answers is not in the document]
omri374: Checked directly rather than taking this on faith: `LocalRecognizer` is not actually abstract. `inspect.isabstract(LocalRecognizer)` returns `False` and `LocalRecognizer.__abstractmethods__` is an empty frozenset — it does implement the `load`/`analyze` methods it inherits from `EntityRecognizer` (even if trivially), so it *is* included in `CONCRETE_RECOGNIZER_CLASSES` and the guard test passes (`test_not_loadable_as_predefined_entry_names_only_real_classes` is green, confirmed by running it directly). No change made here — the described collection failure doesn't reproduce.

---

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
```diff
@@ -316,6 +366,76 @@ def _prepare_recognizer_kwargs(
         accepts_supported_entity = RecognizerListLoader.SUPPORTED_ENTITY in params
         accepts_supported_entities = RecognizerListLoader.SUPPORTED_ENTITIES in params
 
+        # Parameters reachable through the constructor chain: the leaf signature
+        # plus, while each ``__init__`` forwards ``**kwargs``, its parents'. A
+        # subclass such as StanzaRecognizer accepts ``supported_entities`` and
+        # ``context`` through SpacyRecognizer even though its own signature
+        # names neither, so the two warnings below key off this set and not the
+        # leaf signature alone.
+        reachable = RecognizerListLoader._reachable_init_param_names(recognizer_cls)
```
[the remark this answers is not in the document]
omri374: Confirmed and fixed in f6cd16bd. Reproduced first: a `**kwargs` leaf forwarding to a parent declaring only `supported_entity` (singular) really does raise `TypeError: unexpected keyword argument 'supported_entities'` when constructed with the prepared kwargs, since neither the plural->singular conversion nor the singular-drop safety net looked past the leaf signature.

Both now key off `_reachable_init_param_names`, matching the entity-key warning logic just above them. Left the plural-kept-on-`**kwargs` filter as is — that's the established, tested compat path for a class that reads the value from `**kwargs` without declaring it (e.g. `BasicLangExtractRecognizer`), and by the time that filter runs, the plural key has already been converted away in the fixed scenario. Verified no current concrete recognizer hits this combination either way. Added `test_plural_converted_to_singular_when_only_singular_is_reachable`, which constructs the class (not just inspects the prepared dict) with a new `StrictSingularParent`/`ChildForwardsKwargsToSingularParent` fixture pair.

---

presidio-analyzer/tests/test_recognizer_config_conformance.py
```diff
@@ -0,0 +1,678 @@
...
+    # Mirrors the loader: context is dropped only for a strict leaf signature
+    # that cannot reach a ``context`` parameter. A leaf accepting **kwargs keeps
+    # the key, so no warning fires for it.
+    leaf_params = inspect.signature(cls.__init__).parameters
+    leaf_accepts_var_kw = any(
+        p.kind == inspect.Parameter.VAR_KEYWORD for p in leaf_params.values()
+    )
+    accepts_context = (
+        "context" in _reachable_init_param_names(cls) or leaf_accepts_var_kw
+    )
```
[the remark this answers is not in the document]
omri374: Fixed in f6cd16bd: removed the `or leaf_accepts_var_kw` — it was a leftover from before the `**kwargs` carve-out for `context` was reverted in the previous commit and no longer matched what the loader does. `accepts_context` is now just `"context" in _reachable_init_param_names(cls)`.

---

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
```diff
@@ -313,12 +381,83 @@ def _prepare_recognizer_kwargs(
...
+
+        # ``context`` is one flat word list applied to every result a recognizer
```
omri374: do we need all this prose? code is supposed to be self explanatory

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
```diff
@@ -313,12 +381,83 @@ def _prepare_recognizer_kwargs(
...
+        # Keeping the key in kwargs for a leaf with **kwargs is therefore not
+        # provably safe in general: it still raises TypeError once forwarding
+        # reaches that stricter ancestor. Registry-build safety takes priority
+        # over preserving a value a hypothetical out-of-tree recognizer might
+        # read from **kwargs without declaring it.
+        if "context" in kwargs and "context" not in reachable:
+            kwargs.pop("context")
+            logger.warning(
+                "%s does not accept 'context'; ignoring the context words "
+                "configured for it. Context words boost every result a "
```
omri374: not every result. Only if there's a context match (and this is just with the existing context mechanism)

presidio-analyzer/tests/test_recognizer_config_conformance.py
```diff
@@ -0,0 +1,695 @@
...
+# ``--all-extras``) these constructors refuse to build with an actionable
+# ImportError/ValueError -- intended behavior, not a conformance failure -- so
+# the parametrized cases below report a skip instead of a hard failure.
+#
+# The skip is decided by probing for the module, not by catching the exception:
+# an unexpected ImportError from any class, including these, still fails the
+# test rather than turning green. Classes whose optional import happens only in
+# ``load()`` (GLiNER, Stanza, Transformers NER) are absent on purpose -- this
+# suite patches ``load`` to a no-op, so they construct without the extra.
+OPTIONAL_DEPENDENCY_MODULES: Dict[str, Tuple[str, ...]] = {
```
omri374: This list would go out of date quickly. Any way of doing this dynamically?

presidio-analyzer/tests/test_recognizer_config_conformance.py
```diff
@@ -0,0 +1,695 @@
...
+            )
+
+
+def test_optional_dependency_modules_names_only_real_classes():
+    """Guard ``OPTIONAL_DEPENDENCY_MODULES`` against stale entries."""
+    names = {cls.__name__ for cls in CONCRETE_RECOGNIZER_CLASSES}
+    assert set(OPTIONAL_DEPENDENCY_MODULES) <= names
+
+
+def _reachable_init_param_names(cls: Type[EntityRecognizer]) -> Set[str]:
```
omri374: why is this in the test? We have it in the recognizer_loader_utils

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
omri374: Fixed in d17a1c74: reworded to "Context words boost a result's score only when they match text near the recognized entity" — this describes the existing `ContextAwareEnhancer`/`LemmaContextAwareEnhancer` matching mechanism (score boost only fires on an actual context-word match nearby, not unconditionally on every emitted result), not new behavior from this PR. Also fixed the same imprecise phrasing in the conformance suite's own module-level comment, which had the identical wording.

---

presidio-analyzer/tests/test_recognizer_config_conformance.py
omri374: Good catch — fixed in d17a1c74: removed the test's local copy and switched both call sites to `RecognizerListLoader._reachable_init_param_names` from production code. Confirmed the two implementations were byte-for-byte identical logic before removing the duplicate, and the full suite (648 tests) still passes with the shared implementation.

---

presidio-analyzer/tests/test_recognizer_config_conformance.py
omri374: I tried an AST-based dynamic approach: walk each class's MRO, parse each module's source, and pull the module names out of any top-level `try: import X ... except ImportError:` block — the pattern all six of these classes already use. It found two real problems that make it unsafe as a full replacement:

1. **False negatives (would let a hard failure through instead of skipping):** `AzureHealthDeidRecognizer` calls `get_azure_credential()`, whose `azure.identity` import guard lives in a different module (`presidio_analyzer.llm_utils.azure_auth_helper`) that isn't in the class's MRO — an AST walk of the class's own ancestor chain never sees it. Same problem for `BasicLangExtractRecognizer`, whose `langextract` guard lives in `presidio_analyzer.llm_utils.langextract_helper`, reached only via a runtime call to `check_langextract_available()`, not an import in the MRO chain.
2. **False positives (would silently skip a case that should run):** `GLiNERRecognizer` has the same module-level `try/except ImportError` pattern for `gliner`, but that import is only used in `load()`, not `__init__` — construction succeeds fine without it (this suite patches `load` to a no-op). A naive AST walk can't distinguish "guarded name used in `__init__`" from "guarded name used only in `load()`" without much deeper analysis, and getting that wrong means silently losing conformance coverage for a class that should have been exercised.

Chasing both of those correctly would mean either executing construction and parsing free-form exception text/types that aren't consistent across these six classes (some raise `ImportError`, some `ValueError`, with hand-written messages, no shared structured exception), or a much deeper transitive-call static analysis — more fragile than the list it would replace, for a suite whose whole point is catching real regressions rather than introducing new ways to silently skip them.

I didn't push the AST version for that reason. Let me know if you'd rather trade that safety for less maintenance burden — happy to reconsider, but the small guard test (`test_optional_dependency_modules_names_only_real_classes`) already catches stale/renamed class names, so the realistic failure mode left is "forgot to add a new optional-dependency recognizer," which fails as a hard test failure (not a silent skip) the moment CI runs without the extra installed.

---

omri374: what about a more structural change? add fields to specific recognizer classes or have them implement mixin/abstract classes?

omri374: Yes — that's better. Pushed in 1ef481b1: added `EntityRecognizer.OPTIONAL_DEPENDENCY_MODULES: ClassVar[Tuple[str, ...]] = ()`, and each of the six recognizers that needs an extra declares it directly on itself, right next to its `try/except ImportError`:

- `AzureAILanguageRecognizer` → `("azure.ai.textanalytics", "azure.core.credentials")`
- `AzureHealthDeidRecognizer` → `("azure.health.deidentification", "azure.identity")`
- `LangExtractRecognizer` (base of `AzureOpenAILangExtractRecognizer` and `BasicLangExtractRecognizer`) → `("langextract",)`
- `HuggingFaceNerRecognizer` (base of `MedicalNERRecognizer`) → `("transformers", "torch")`

Putting it on `LangExtractRecognizer`/`HuggingFaceNerRecognizer` rather than each leaf class means the two subclasses inherit the declaration for free via normal attribute lookup — no separate entry needed when a new LangExtract or HuggingFace-NER-based recognizer is added, only when a genuinely new optional dependency shows up. The test suite now just reads `cls.OPTIONAL_DEPENDENCY_MODULES` (walking the MRO isn't even needed — Python's own attribute lookup already does it). Removed the standalone dict, the file-level guard test, and the AST-parsing approach I'd tried and rejected in the previous reply — this is simpler and doesn't have those correctness gaps. Verified: all six classes still resolve to the exact same modules as before, full suite (647 tests) and lint pass.

---
_Generated by [Claude Code](https://claude.ai/code)

---

presidio-analyzer/tests/test_recognizer_config_conformance.py
```diff
@@ -0,0 +1,647 @@
...
+    expected = KNOWN_CONTRACT_GAPS.get(cls.__name__, set())
+    assert missing == expected, (
+        f"{cls.__name__} is missing registry-injected keys {sorted(missing)}; "
+        f"KNOWN_CONTRACT_GAPS expects exactly {sorted(expected)}. Either make "
+        f"the constructor accept the missing key(s), or update "
+        f"KNOWN_CONTRACT_GAPS to match reality."
+    )
+
+
+def test_known_contract_gaps_names_only_real_classes():
```
omri374: is this required? KNOWN_CONTRACT_GAPS is empty. Let's simplify

presidio-analyzer/tests/test_recognizer_config_conformance.py
```diff
@@ -0,0 +1,647 @@
...
+        _skip_if_optional_dependency_missing(entry_cls)
+    language_configs = _entry_language_configs(entry)
+    languages = [language for language, _ in language_configs]
+    conf_entry = dict(entry, enabled=True)
+    configuration = {
+        "global_regex_flags": GLOBAL_REGEX_FLAGS,
+        "supported_languages": languages,
+        "recognizers": [conf_entry],
+    }
+    monkeypatch.chdir(PACKAGE_ROOT)
```
omri374, line 313: why is this needed here?

presidio-analyzer/tests/test_recognizer_config_conformance.py
```diff
@@ -0,0 +1,647 @@
...
+    # rejects `patterns`/`deny_list` on a `type: predefined` entry outright
+    # ("... is marked as 'predefined' but contains 'patterns' or
+    # 'deny_list' ..."), so no predefined entry can ever supply them.
+    "PatternRecognizer",
+    # Requires `target_classification` positionally, with no default and no
+    # schema field to set it from (base `PredefinedRecognizerConfig` has
+    # none, and no CONFIG_MODEL_MAP entry adds one) -- a base class for
+    # subclassing (ZaMobileNumberRecognizer, ZaTelephoneNumberRecognizer),
+    # never meant to be named directly in configuration.
+    "ZaPhoneNumberRecognizer",
```
omri374, line 437: can we create a dedicated test for this one?

presidio-analyzer/tests/test_recognizer_config_conformance.py
```diff
@@ -0,0 +1,647 @@
...
+# currently not used by LLM recognizers"). The round-trip test below skips
+# only the context assertion for these classes. Classes that do not accept
+# `context` at all are handled dynamically: the loader drops the key with a
+# warning and the instance keeps the base-class default, ``[]``.
+CONTEXT_NOT_APPLIED = {"BasicLangExtractRecognizer"}
+
+# Classes that cannot be constructed through a `type: predefined` registry
+# entry at all, even with REQUIRED_KWARGS -- not a config-layer gap, but a
+# structural mismatch this suite cannot paper over:
+NOT_LOADABLE_AS_PREDEFINED_ENTRY = {
```
omri374, line 419: could we skip abstract classes instead of naming each one? why can't we instantiate PatternRecognizer?

presidio-analyzer/tests/test_recognizer_config_conformance.py
```diff
@@ -0,0 +1,647 @@
...
+    assert instance.score_thresholds == {"default": 0.42}
+
+
+def test_context_not_applied_names_only_real_classes():
+    """Guard ``CONTEXT_NOT_APPLIED`` against stale entries."""
+    names = {cls.__name__ for cls in CONCRETE_RECOGNIZER_CLASSES}
+    assert CONTEXT_NOT_APPLIED <= names
+
+
+def test_entry_context_is_dropped_for_bare_language_list():
```
omri374, line 596: please check if there aren't duplicate tests with existing tests of this functionality

presidio-analyzer/tests/test_recognizer_config_conformance.py
omri374: Right — removed in eb740bbd. With `KNOWN_CONTRACT_GAPS` empty, `set(KNOWN_CONTRACT_GAPS) <= names` was always trivially true, so this guard test never actually caught anything. And it turns out it's redundant anyway: `test_recognizer_accepts_registry_injected_keys` already surfaces a stale/renamed entry on its own — if a key in `KNOWN_CONTRACT_GAPS` doesn't match any real class name, `KNOWN_CONTRACT_GAPS.get(cls.__name__, set())` just returns an empty `expected` for the real (renamed) class, so `missing != expected` fails loudly for that class with a message telling you to update the dict. No separate guard needed.

---

presidio-analyzer/tests/test_recognizer_config_conformance.py
omri374: Checked directly — none of these three are actually abstract, so "skip abstract classes" wouldn't touch them (that filter already runs in `_concrete_recognizer_classes` via `inspect.isabstract`, which is exactly why they're still in `CONCRETE_RECOGNIZER_CLASSES` and need this separate exclusion):

```
LocalRecognizer        isabstract=False  abstractmethods=frozenset()
PatternRecognizer       isabstract=False  abstractmethods=frozenset()
ZaPhoneNumberRecognizer isabstract=False  abstractmethods=frozenset()
```

For `PatternRecognizer` specifically: it's fully instantiable in Python (`PatternRecognizer(supported_entity="X", patterns=[...])` works fine directly), the problem is the YAML *schema*, not the class. A `type: predefined` entry can't carry `patterns`/`deny_list` at all — `RecognizerRegistryConfig.parse_recognizers` rejects it before construction even starts:

```
ValueError: Recognizer 'conf_PatternRecognizer' is marked as 'predefined' but
contains 'patterns' or 'deny_list' which are only valid for custom recognizers.
Either use type: 'custom' or remove these fields.
```

And without patterns/deny_list, the constructor itself refuses: `ValueError: Pattern recognizer should be initialized with patterns or with deny list`. So it's a genuine structural gap between the `predefined` entry schema and what this class needs, not something an abstractness check could catch. Added a dedicated test in eb740bbd that actually reproduces this (and the other two) rather than asking you to trust the comment — see the reply on the `can we create a dedicated test for this one?` thread.

---

presidio-analyzer/tests/test_recognizer_config_conformance.py
omri374: This is needed for one specific shipped entry: `BasicLangExtractRecognizer`'s entry in `default_recognizers.yaml` sets `config_path: presidio_analyzer/conf/langextract_config_basic.yaml`, a relative path the recognizer resolves against the current working directory. Without pinning the cwd to `PACKAGE_ROOT`, this test would only pass when pytest happens to be invoked from the package root and fail (or resolve the wrong file) from anywhere else — e.g. running `pytest` from the repo root instead of `presidio-analyzer/`.

This mirrors the exact same pattern already established in `test_recognizers_loader_utils.py::test_yaml_entry_loads_when_enabled` (same `PACKAGE_ROOT` import, same `monkeypatch.chdir(PACKAGE_ROOT)`, same reasoning) — not something new invented for this test, just reusing the existing fix for the same underlying issue.

---

presidio-analyzer/tests/test_recognizer_config_conformance.py
omri374: Added in eb740bbd: `test_not_loadable_as_predefined_entry_actually_fails`, parametrized over all three classes in `NOT_LOADABLE_AS_PREDEFINED_ENTRY` (`LocalRecognizer`, `PatternRecognizer`, `ZaPhoneNumberRecognizer`). Each case builds the smallest possible `type: predefined` entry naming the class and asserts construction fails with the *specific* error the exclusion's comment claims (verified by actually reproducing each one first — see my reply on the "abstract classes" thread for `PatternRecognizer`'s). A companion guard test (`test_not_loadable_as_predefined_entry_errors_matches_the_set`) keeps the new error-mapping dict in sync with the exclusion set itself. This replaces "trust the comment" with an executable proof, and will fail loudly if a future change ever closes one of these gaps without updating the exclusion set.

---

presidio-analyzer/tests/test_recognizer_config_conformance.py
omri374: Checked — no duplicate. This test covers a genuinely different code path from the existing context tests:

- `test_context_dropped_with_warning_for_class_not_accepting_it` / `test_context_dropped_for_leaf_forwarding_kwargs_to_a_strict_parent` (in `test_recognizers_loader_utils.py`) cover a class that doesn't *accept* `context` at all — `_prepare_recognizer_kwargs` dropping the key.
- `test_recognizer_registry_provider.py`'s context assertions (`spanish_recognizer.context`, `zipcode_recognizer.context`) only use the per-language `{language, context}` dict shape, and the one entry there using a bare list (`CreditCardRecognizer: supported_languages: [en]`) never sets `context` at all.

`test_entry_context_is_dropped_for_bare_language_list` is the only test that combines a class that *does* accept `context` (`CreditCardRecognizer`) with the bare-list-of-language-codes shape *and* an entry-level `context` key set — the specific combination where `_get_recognizer_languages` silently returns `None` instead of the entry's context. No other test in the suite exercises that shape/key combination.

---

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizers_loader_utils.py
```diff
@@ -70,28 +70,41 @@ def _get_recognizer_context(
...
+        string it matches neither list and the recognizer is never built.
+
         :param recognizers_conf: The recognizers' configuration
         """
 
+        normalized = [
+            {"name": recognizer_conf, "type": "predefined"}
+            if isinstance(recognizer_conf, str)
+            else recognizer_conf
+            for recognizer_conf in recognizers_conf
```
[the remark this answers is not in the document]
omri374: Confirmed and fixed in da67ac4c — but the root cause is one level up from here, not in this normalization step. Reproduced first: `{"global_regex_flags": 26, "recognizers": [{"name": "CreditCardRecognizer", "type": "predefined"}]}` (the pre-existing mapping form, no bare string involved) raises the identical `TypeError: 'NoneType' object is not iterable`. So this isn't specific to the shorthand — it's `RecognizerRegistryProvider.create_recognizer_registry` passing `self.configuration.get("supported_languages")` (`None`, since the defaults merge is skipped whenever a raw dict already supplies `recognizers` + `global_regex_flags`) straight into `RecognizerListLoader.get`, which iterates it directly. `RecognizerRegistry.__init__` already falls back to `["en"]` a few lines later, just too late to help.

Fixed by applying that same `["en"]` default where `supported_languages` is first read in `create_recognizer_registry`, before it reaches `RecognizerListLoader.get`. This fixes both the bare-string and mapping-form cases at once. Added `test_bare_string_entry_builds_when_supported_languages_is_omitted`, which reproduces the original crash and asserts the entry now loads with the `["en"]` default.

---

presidio-analyzer/tests/test_recognizer_config_conformance.py
```diff
@@ -0,0 +1,689 @@
...
+    raised = False
+    with caplog.at_level("WARNING", logger="presidio-analyzer"):
+        try:
+            RecognizerRegistryProvider(
+                registry_configuration=configuration
+            ).create_recognizer_registry()
+        except ValueError as exc:
+            # Only counts if it actually names the unknown key -- an unrelated
+            # ValueError must not be mistaken for the gap having been closed.
+            if "no_such_key" in str(exc):
```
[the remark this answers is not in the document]
omri374: Confirmed and fixed in da67ac4c. Verified the exact failure mode by simulating turn 06's `extra="forbid"` fix locally (monkeypatching `PredefinedRecognizerConfig.model_config`): the wrapped exception's `str(exc)` is literally just `"Invalid recognizer registry configuration"` — the pydantic `ValidationError` naming `no_such_key` is preserved only on `exc.__cause__` (`raise ValueError(...) from e`). The old check (`"no_such_key" in str(exc)`) stayed `False` even after simulating the fix, so it would have re-raised and kept the test XFAIL forever instead of flipping to XPASS.

Fixed by checking both `str(exc)` and `str(exc.__cause__)`. Re-ran the same simulation after the fix: the check now correctly comes back `True`.

---

omri374:
Addressed the two suppressed findings from the latest Copilot review (review 5223397738) that don't have their own inline thread:

**`recognizers_loader_utils.py:411`** — confirmed real. An explicitly empty `supported_entities: []` failed the truthiness check (`isinstance(x, list) and x`), so the plural key was never removed and, kept by the `**kwargs`-compat filter, forwarded to a strict singular-only parent — reproduced directly: `ChildForwardsKwargsToSingularParent(supported_entities=[])` raised `TypeError: unexpected keyword argument 'supported_entities'`. Fixed in da67ac4c: the plural key is now always popped once singular-only is reachable, only setting the singular default when the list is non-empty. Added `test_empty_plural_dropped_when_only_singular_is_reachable`, which constructs the class with the prepared kwargs.

**`test_recognizers_loader_utils.py:578`** — confirmed real. The test stripped `supported_language` out of the prepared kwargs before constructing `ChildForwardsKwargs`, because `StrictParent` didn't declare it — so a regression in how `supported_language` (or any other injected key) is handled wouldn't have been caught here, only in the context-specific assertion. Fixed in da67ac4c: `StrictParent` now declares every registry-injected key (`supported_entities`, `name`, `supported_language`) — still not `context`, which is the key under test — so the test constructs with the real, unmodified prepared kwargs instead of a hand-stripped subset.

Both verified by direct reproduction before and after the fix, full suite (652 tests) and lint clean.

---

presidio-analyzer/presidio_analyzer/recognizer_registry/recognizer_registry_provider.py
```diff
@@ -66,7 +66,14 @@ def __init__(
...
         """Create a recognizer registry according to configuration loaded previously."""
-        supported_languages = self.configuration.get("supported_languages")
+        # A raw dict registry_configuration that supplies "recognizers" and
+        # "global_regex_flags" but omits "supported_languages" skips the
+        # defaults merge in RecognizerConfigurationLoader.get, leaving this
+        # None. RecognizerRegistry.__init__ falls back to ["en"] a few lines
+        # below, but RecognizerListLoader.get -- called first, right here --
+        # iterates this value directly and raises TypeError on None. Apply
+        # the same ["en"] default before that call, not just after it.
+        supported_languages = self.configuration.get("supported_languages") or ["en"]
```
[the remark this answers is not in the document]
omri374: Confirmed and fixed in b337cf5d. Reproduced first: `{"global_regex_flags": 26, "supported_languages": [], "recognizers": ["CreditCardRecognizer"]}` silently built an English `CreditCardRecognizer` the caller never asked for. Switched the check from `or ["en"]` to `is None` specifically, since `RecognizerRegistryConfig` deliberately preserves an explicit `[]` as distinct from omitted (`test_recognizer_registry_config_empty_languages` in `test_yaml_recognizer_models.py`). Added `test_explicit_empty_supported_languages_is_not_coerced_to_en`, which now asserts `registry.recognizers == []` for that configuration.

---

omri374:
Re: the suppressed finding in review 5224240201 (`recognizer_registry_provider.py:76`, the `EsNifRecognizer`/per-entry-language case) — the explicit-empty-list part is fixed in b337cf5d (see my reply on the other thread). For the per-entry-language part specifically, I don't think this is a new bug to fix here.

Reproduced the "before" state directly: with `supported_languages` omitted entirely and an entry setting its own `supported_languages: ["es"]`, the code **already crashed** before my previous fix (`TypeError: argument of type 'NoneType' is not iterable`) — same failure, just a hard crash instead of a silent drop. My fix didn't turn a working case into a broken one; it replaced one failure mode with the codebase's existing, documented behavior for language mismatches.

That behavior is intentional and documented in this repo's own `AGENTS.md`: *"Non-English recognizers must set the top-level `supported_languages` in the test config — it defaults to `[\"en\"]` and silently filters everything else."* An entry declaring `es` while the registry-level list is `["en"]` (whether that `["en"]` came from an explicit value or this default) has always been silently filtered by `_is_language_supported_globally` — that's not something introduced by my change, and I verified it reproduces identically with an explicit `supported_languages: ["en"]` set alongside the `es` entry, no omission involved.

Deriving the effective registry-level languages from per-entry declarations would be a real, separate feature (and a behavior change for any existing config relying on today's filtering) — out of scope for the crash fix this turn was addressing. Happy to open a follow-up if you'd like that revisited, but didn't want to silently expand this PR's scope with an invented fallback policy.

---

PR #1995, opened 2026-04-20
feat: add Turkish National ID (TR_NATIONAL_ID) recognizer

mrcuren:
Adds Turkish National ID (TCKN / TC Kimlik Numarası) recognizer to Presidio Analyzer.

The TCKN is an 11-digit identification number issued to Turkish citizens and foreign residents. It is the primary PII under Turkish KVKK (Personal Data Protection Law).

**Features:**
- Pattern recognition for 11-digit numbers starting with 1-9
- Algorithmic validation of 10th and 11th digits per official NVI specification
- Context words for higher confidence detection (TC Kimlik, TCKN, etc.)
- Disabled by default as per country-specific recognizer guidelines

## Issue reference
Part of #1973

## Testing
- Added `test_tr_national_id_recognizer.py` with >=90% coverage on changed lines
- Tests include valid checksums, invalid checksums, and false positive checks (short/long inputs, non-digits)
- All existing tests continue to pass

## Checklist
- [x] I have reviewed the [contribution guidelines](https://github.com/microsoft/presidio/blob/main/CONTRIBUTING.md)
- [x] My code follows the project style guidelines (ruff, pytest)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have updated the CHANGELOG.md under the Unreleased section
- [x] I have updated the supported_entities.md documentation
- [x] I have added my recognizer to `default_recognizers.yaml` with `enabled: false`
- [x] I have added my recognizer to `__init__.py` and `__all__`

Issue #1973, opened 2026-04-17
feat: add Turkey (TR) country-specific predefined recognizers


mrcuren:
## Summary

This issue proposes adding Turkey-specific PII recognizers to Presidio Analyzer, following the same pattern established by existing country-specific recognizers (e.g., German recognizers in #1828, UK recognizers in #1862).

### 1. `TrNationalIdRecognizer` — `TR_NATIONAL_ID`
Turkish National Identification Number (TC Kimlik No / TCKN).
- 11-digit number, first digit ≠ 0
- Checksum validation based on the official Nüfus ve Vatandaşlık İşleri Genel Müdürlüğü (NVI) algorithm, widely documented in Turkish software engineering resources and referenced in KVKK compliance guidelines

### 2. `TrLicensePlateRecognizer` — `TR_LICENSE_PLATE`
Turkish vehicle license plates (standard civilian format).
- Format: `[province_code 01-81] [1-3 letters] [2-4 digits]`
- Province code validation (01–81)
- Scope: standard civilian plates only (diplomatic CD/CC and temporary S-plates are out of scope for this PR)
- Examples: `34 ABC 1234`, `06 A 123`

### 3. `TrPhoneNumberRecognizer` — `TR_PHONE_NUMBER` *(lower priority)*
Turkish mobile phone numbers.
- Supports international (`+90 5XX ...`), national (`0 5XX ...`), and local (`5XX ...`) formats
- Format-only validation (no operator prefix validation, as mobile number portability / MNP makes operator-based checks unreliable)
- Reference: ITU-T E.164, Turkey country code +90

## Implementation Plan

I plan to implement these in **separate PRs** to keep reviews focused:
- **PR 1:** `TrNationalIdRecognizer` (TCKN) — highest value, well-defined algorithm
- **PR 2:** `TrLicensePlateRecognizer`
- **PR 3:** `TrPhoneNumberRecognizer` *(if the approach is agreed upon)*

All recognizers will be disabled by default in `default_recognizers.yaml`, placed under `country_specific/turkey/`, with full unit tests and documentation updates.

## Related

- #1828 (German recognizers)
- PR #1862 (UK recognizers)

SharonHart:
@mrcuren See if phone number recognizer is covered by the generic phone number one, it should have country specific checks

mrcuren:
[...]

PR #2250, opened 2026-09-12
feat: add Turkish Tax ID (TR_TAX_ID / VKN) recognizer

gkhnelbstn:
## Summary

Adds `TrTaxIdRecognizer` (`TR_TAX_ID`) for the Turkish Tax Identification
Number (Vergi Kimlik Numarası / VKN) — a 10-digit identifier issued by the
Revenue Administration (GİB) to legal entities and to individuals not
eligible for the 11-digit National ID. It follows the same shape as
`TrNationalIdRecognizer` (#1995): pattern match, context words, and its own
checksum.

Related to #1973, which scoped the original three Turkey recognizers
(national ID, license plate, phone number) — VKN wasn't in that list, but
it's the natural next one: a `tax_id` column in a Turkish system genuinely
mixes VKN (companies) and TCKN (sole traders), both digit strings of
adjacent length, so a checksum rather than a length check is what tells
them apart.

## Changes

- `presidio-analyzer/presidio_analyzer/predefined_recognizers/country_specific/turkey/tr_tax_id_recognizer.py` — the recognizer, with the GİB checksum algorithm and a docstring explaining it (the algorithm itself is undocumented by GİB but is the one their own e-Fatura/e-Devlet integrations implement, and it's widely reproduced in Turkish accounting/ERP software)
- `presidio-analyzer/tests/test_tr_tax_id_recognizer.py` — 29 cases: valid/invalid checksums, wrong length, non-digits, the leading-zero case (VKN allows it, unlike TCKN), context words, entity/language/country_code
- `country_specific/turkey/__init__.py`, `predefined_recognizers/__init__.py` — export
- `conf/default_recognizers.yaml` — registered, `enabled: false` (country-specific default), `country_code: tr` set from the start (no follow-up migration needed the way #2000 added it retroactively to `TrNationalIdRecognizer`)
- `docs/supported_entities.md`, `CHANGELOG.md`

## Testing

```
pytest tests/test_tr_tax_id_recognizer.py -v
# 29 passed

pytest tests/test_tr_national_id_recognizer.py tests/test_tr_license_plate_recognizer.py tests/test_recognizer_registry.py -q
# 103 passed -- no regression on the sibling Turkey recognizers or the country-filter registry

ruff check <touched files>
# All checks passed!
```

`ruff format --check` on the new test file reports it would reformat the
parametrize tuples onto multiple lines each — I left it matching
`test_tr_national_id_recognizer.py`'s existing (also unformatted-by-this-rule)
style rather than diverge from the sibling file; happy to reformat either
file if that's preferred.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Analyzer
 #### Added
+- Added `TrTaxIdRecognizer` for the Turkish Tax Identification Number (`TR_TAX_ID` / VKN), a 10-digit identifier issued by the Revenue Administration (GİB), using pattern match, context, and GİB checksum validation. Disabled by default.
 - Added `UuidRecognizer` (generic, entity type `UUID`) to detect UUIDs in the standard 8-4-4-4-12 hyphenated hexadecimal format, covering RFC 4122 versions 1-5 and RFC 9562 versions 6-8. Validates version and variant nibbles and filters the nil UUID to reduce false positives.
 - South African ID number (`ZA_ID_NUMBER`) recognizer for the 13-digit national identity number, using pattern matching, context words, birth-date validation, and Luhn checksum validation. Disabled by default.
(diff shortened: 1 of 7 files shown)


zelihaguven (approved):
Reviewed as a Turkish speaker. VKN vs TCKN length/checksum distinction and the leading-zero case look right for GİB practice. Tests cover the sibling Turkey recognizers; `enabled: false` + `country_code: tr` match the country-specific default. LGTM from the domain/docs side — happy to defer to maintainers on registry wiring.

PR #2253, opened 2026-09-13
fix(analyzer): keep Stanza multi-word tokens as surface tokens to preserve text and entity offsets

### LINKED ISSUE #2249: StanzaNlpEngine: German multi-word tokens (im/am/zum) replace the doc text and drop every NER entity; indented text ends in 500 "Did not find word"
**Describe the bug**

With `StanzaNlpEngine` and the German model, any text containing a German multi-word token (a preposition-article contraction such as `im`, `am`, `zum`, `zur`, `beim`, `vom`, `ins`, `ans`) loses **every** NER entity: `/analyze` returns HTTP 200 with an empty list where the same sentence with the contraction written out returns the PERSON. On indented text (JSON snippets, tables, lists) the same defect turns into HTTP 500 `Did not find word '...' in the list of tokens although it is expected to be found` from `LemmaContextAwareEnhancer`.

Cause (read from `presidio_analyzer/nlp_engine/stanza_nlp_engine.py`, same on `main` today):

1. `StanzaNlpEngine.load` hardcodes `processors="tokenize,pos,lemma,ner"`; Stanza force-adds `mwt` for `de` ("Language de package default expects mwt, which has been added") and expands `im` → `in` + `dem`.
2. The inlined spacy-stanza tokenizer flattens `token.words` in `StanzaTokenizer.__get_tokens_with_heads`, so the expanded words no longer match the text. `__get_words_and_spaces` raises, and `_convert_doc` then **replaces the whole doc text** by `" ".join(tokens)` (the `"Due to multiword token expansion or an alignment issue, the original text has been replaced by space-separated expanded tokens."` warning).
3. The Stanza NER spans carry offsets of the original text. In the replaced doc they hit no token boundary, `char_span` returns `None`, and the entity is silently dropped → 200, empty result.
4. In indented text the replaced text is shorter than the original, so the token list ends before the text does. For a pattern recognizer hit past the last token, `LemmaContextAwareEnhancer._find_index_of_match_token` raises `ValueError("Did not find word ...")` → 500. (The loop condition `tokens_indices[i] == start or start < tokens_indices[i] + len(token)` accepts any start before the end of some token, so it only fires once the hit lies beyond the tokenizer's coverage. That is why the offending word in the message varies between runs.)

This is the presidio-side face of explosion/spacy-stanza#70 (open since 2021). Because Presidio inlines the tokenizer, the fix has to land here.

Pattern recognizers (EMAIL, IP, IBAN, ...) are not affected, they run on the original text. The stock spaCy `de_core_news_*` engine is not affected either (no MWT expansion).

**To Reproduce**

Image `mcr.microsoft.com/presidio-analyzer:2.2.362` (`presidio-analyzer` 2.2.362, `stanza` 1.14.0) with a Stanza NLP configuration:

```yaml
nlp_engine_name: stanza
models:
  - lang_code: de
    model_name: de
ner_model_configuration:
  labels_to_ignore: [O]
  model_to_presidio_entity_mapping:
    PER: PERSON
    LOC: LOCATION
    ORG: ORGANIZATION
    MISC: NRP
```

Minimal pair (`score_threshold` 0.4, `entities: ["PERSON"]`, `language: "de"`):

| text | result |
|---|---|
| `Wir treffen uns im Büro mit Thomas Bergmann.` | `[]` |
| `Wir treffen uns in dem Büro mit Thomas Bergmann.` | PERSON 32–47 |
| `Wir gehen zum Termin mit Thomas Bergmann.` / `... zu dem Termin ...` | `[]` / PERSON 28–43 |
| `Beim Kunden spricht Thomas Bergmann.` / `Bei dem Kunden ...` | `[]` / PERSON 23–38 |
| `Die Mail vom Kunden kam von Thomas Bergmann.` / `... von dem Kunden ...` | `[]` / PERSON 32–47 |

```bash
curl -s http://localhost:5001/analyze -H 'Content-Type: application/json' \
  -d '{"text":"Wir treffen uns im Büro mit Thomas Bergmann.","language":"de","score_threshold":0.4,"entities":["PERSON"]}'
# -> []
```

The 500 path, same container: one multi-word token plus enough whitespace that the space-joined replacement text ends before the IP starts (a run of 40 spaces here; real-world triggers are indented JSON, tables and lists):

```bash
curl -s -i http://localhost:5001/analyze -H 'Content-Type: application/json' \
  -d '{"text":"Wir treffen uns im Büro.\n                                        \nServer 192.168.10.20","language":"de","entities":["IP_ADDRESS"]}'
# -> HTTP 500 {"error":"Did not find word '192.168.10.20' in the list of tokens although it is expected to be found"}
```

The same text with `in dem` instead of `im` returns 200 with the IP at 77–90.

In-process, without HTTP:

```python
from presidio_analyzer.nlp_engine import StanzaNlpEngine
engine = StanzaNlpEngine(models=[{"lang_code": "de", "model_name": "de"}])
engine.load()
doc = engine.nlp["de"]("Wir treffen uns im Büro mit Thomas Bergmann.")
print(repr(doc.text))   # 'Wir treffen uns in dem Büro mit Thomas Bergmann . '  <- text replaced
print(list(doc.ents))   # []
doc = engine.nlp["de"]("Wir treffen uns in dem Büro mit Thomas Bergmann.")
print([(e.text, e.label_, e.start_char, e.end_char) for e in doc.ents])  # [('Thomas Bergmann', 'PER', 32, 47)]
```

**Expected behavior**

`doc.text == text`, NER entities with their original offsets (PERSON 28–43 for the first sentence), no 500 on indented text.

**Additional context**

A build-time patch we run in production keeps a multi-word token as **one** surface token in `__get_tokens_with_heads` (text and offsets of the token, the remaining annotations borrowed from its first word, lemma = surface form). With it the alignment holds, `doc.text == text`, entities keep their original offsets, and lemmas of ordinary tokens (hence context words) are unchanged. Verified on a 12-slice German document: before 11× 200 with 0 PERSON and 1× 500, after 12× 200 with 22 PERSON; the English leg identical.

```python
for token in sentence.tokens:
    if len(token.words) > 1:
        heads.append(0)
        tokens.append(_SurfaceToken(token))   # text/lemma = token.text, start/end_char, ner of the token,
        continue                              # every other attribute via __getattr__ from token.words[0]
    for word in token.words:
        ...
offset += sum(1 if len(token.words) > 1 else len(token.words) for token in sentence.tokens)
```

A `processors="tokenize,ner"` variant is not a way out: Stanza adds `mwt` for `de` regardless, and lemmas become empty.

The trade-off is that the morpho-syntactic annotation of the expanded words is lost for that token, which for a PII analyzer is the right side to be on: the silent empty result is the harmful outcome here, a 200 with no PERSON is indistinguishable from "nothing there" for every caller. Happy to turn this into a PR against `StanzaTokenizer` if that direction is acceptable.

ManoharPaturi:
Fixes #2249.

`StanzaTokenizer.__get_tokens_with_heads` flattened stanza multi-word-token expansions (`im` -> `in` + `dem`), so token texts no longer matched the source text. `_convert_doc` then rebuilt the doc text from those tokens, which corrupted the document text (see the issue repro), made Stanza NER spans miss every token boundary so entities were silently dropped, and could raise `ValueError` from the lemma enhancer on indented text.

Multi-word tokens now stay a single surface token (a small wrapper carrying surface text/lemma plus the first expanded word's annotations), with a word-index to token-index map that remaps dependency heads across the collapsed token. Single-word tokens keep the previous code path byte-for-byte.

Verified with the real German model: `doc.text` is preserved, `PER` and `IP_ADDRESS` entities are detected where pristine drops them. 11 new tests (pure-unit head remap + German model tests over all 8 contractions) fail on main and pass here; existing stanza suites pass unchanged (12 + 13 + 16), ruff clean vs main baseline.


ArjunPakhan:
Hi @ManoharPaturi 🚀

I pulled branch `pr-2253` locally and verified the analyzer changes in `stanza_nlp_engine.py`. 

Linting and formatting pass cleanly via `ruff`, and unit tests pass. Wrapping multi-word tokens as surface tokens while remapping dependency heads across collapsed tokens cleanly resolves the German MWT offset corruption (`im`, `am`, `zum`, `zur`, `beim`, `vom`, `ins`, `ans`) without introducing regressions to single-word token execution paths.

Great fix! +1 for maintainers to approve workflow execution and merge.

ManoharPaturi:
thanks a lot for pulling the branch and testing it yourself, glad it holds up across the contractions. appreciate the detailed writeup.

FIX-SUMMARY.md
```diff
@@ -0,0 +1,137 @@
+# Fix: StanzaNlpEngine German multi-word tokens replace the doc text and drop entities (#2249)
```
omri374: please remove this file

ManoharPaturi:
@omri374 removed, thanks.

on the copilot point about the head remap using an incorrect relative index: i checked it against the real German model rather than the synthetic fixture, and the remap comes out correct on every token. for 'Wir treffen uns im Büro mit Thomas Bergmann.' stanza gives words (in, dem) as ids 4,5 both governed by id 6 (Büro), which collapses to token 3 governed by token 4:

| token | stanza gov | expected rel | our output |
|---|---|---|---|
| Wir (w1) | w2 | +1 | +1 |
| treffen (w2) | root | 0 | 0 |
| uns (w3) | w2 | -1 | -1 |
| im (w4+w5) | w6 | +1 | +1 |
| Büro (w6) | w2 | -3 | -3 |
| mit (w7) | w8 | +1 | +1 |
| Thomas (w8) | w2 | -5 | -5 |
| Bergmann (w9) | w8 | -1 | -1 |
| . (w10) | w2 | -7 | -7 |

the offset + word_index_to_token_index term is the governor's absolute doc index and len(tokens) is the current token's absolute index at that point, so the difference is the relative head. happy to add this sentence as a regression test if useful.

ArjunPakhan:
Hey @ManoharPaturi! 🚀 

Glad the German model trace held up and disproved the Copilot bot finding. Since we worked through verifying the MWT offset logic and head remapping edge cases together, would you mind adding a co-author trailer to your final commit before merge? 

You can just add this to the end of the commit message:

Co-authored-by: Arjun Sanjay Pakhan <arjunpakhan@gmail.com>

Appreciate it, and great job getting this fix put together!

omri374 (approved):
Thanks!

presidio-analyzer/tests/test_stanza_nlp_engine.py
```diff
@@ -101,3 +122,172 @@ def test_spacy_stanza_english(stanza_pipeline):
...
+    def __init__(self, tokens):
+        self.tokens = tokens
+
+
+class _FakeStanzaDoc:
+    def __init__(self, sentences):
+        self.sentences = sentences
+
+
+def test_get_tokens_with_heads_collapses_mwt_and_remaps_heads():
```
omri374, line 164: Please add `@pytest.mark.skip_engine("stanza_de")`. We don't want to download a german stanza model on every unit test run

presidio-analyzer/tests/test_stanza_nlp_engine.py
```diff
@@ -101,3 +122,172 @@ def test_spacy_stanza_english(stanza_pipeline):
...
+    assert tokens[2].lemma == "im"
+    assert tokens[2].upos == "X"
+
+    # Relative heads: "Ich" -> "gehe", "gehe" is root, "im" -> "Schule",
+    # "Schule" -> "gehe", and the second sentence's words are unaffected
+    # by the words collapsed in the first one
+    assert heads == [1, 0, 1, -2, 1, 0]
+
+
+@pytest.mark.skip_engine("stanza_en")
```
omri374: should this be `@pytest.mark.skip_engine("stanza_de")`?

presidio-analyzer/tests/test_stanza_nlp_engine.py
```diff
@@ -101,3 +122,172 @@ def test_spacy_stanza_english(stanza_pipeline):
...
+    doc = stanza_pipeline_de(text)
+
+    assert doc.text == text
+    assert contraction in [t.text for t in doc]
+    # every token's character span maps back onto itself in the original text
+    for token in doc:
+        assert doc.text[token.idx : token.idx + len(token.text)] == token.text
+
+
+@pytest.mark.skip_engine("stanza_en")
```
omri374:
```suggestion
@pytest.mark.skip_engine("stanza_de")
```

presidio-analyzer/tests/test_stanza_nlp_engine.py
```diff
@@ -101,3 +122,172 @@ def test_spacy_stanza_english(stanza_pipeline):
...
+
+    # NER entities keep the character offsets of the original text
+    expected_start = text.index("Thomas Bergmann")
+    assert len(doc.ents) == 1
+    assert doc.ents[0].text == "Thomas Bergmann"
+    assert doc.ents[0].start_char == expected_start
+    assert doc.ents[0].end_char == expected_start + len("Thomas Bergmann")
+
+
+@pytest.mark.skip_engine("stanza_en")
```
omri374:
```suggestion
@pytest.mark.skip_engine("stanza_de")
```

omri374 (changes requested):
Thanks! The only thing missing is proper handling of downloading Stanza models during unit tests. You'd might need to update the conftest here:
https://github.com/ManoharPaturi/presidio/blob/b71a4b4aeaa889a0f51eb2c7d0612fc1702b1619/presidio-analyzer/tests/conftest.py#L31

To add stanza_de as a preloaded NLP Engine

ManoharPaturi:
done, trailer added to the final commit. thanks for the thorough verification pass, that kind of pull-the-branch testing is exactly what the fix needed.

ManoharPaturi:
@omri374 Updated! Added `stanza_de` to the `nlp_engines` fixture in `conftest.py` (along with a `stanza_de_nlp_engine` session fixture), updated `stanza_pipeline_de` and `stanza_pipeline` to use the preloaded engine when available, and updated the German tests to gate on `@pytest.mark.skip_engine("stanza_de")`. Pushed in b6834f6.

ManoharPaturi:
@omri374 following up on this one. the notes file is removed and i added the co-author trailer arjunpakhan asked for. his verification pass plus your main merge are on the thread. anything else needed before this can be merged?

omri374 (dismissed):

omri374:
Hi @ManoharPaturi, there's a CI failure, could you please take a look?

ManoharPaturi:
@omri374 found it and fixed. the preloaded-engine shortcut was returning the raw stanza pipeline to the english tests, but those run against the spacy-stanza wrapper (they assert EnglishDefaults and t.is_sent_start), so is_sent_start came back None after the first token and the assertion failed. the english fixture is back on load_pipeline, the german one keeps the preloaded engine with the skip_engine gate. pushed, both stanza suites pass locally (2+29 green).

ManoharPaturi:
@omri374 on the new copilot high-severity finding, i re-checked all three layers it touches and the claim doesn't hold on this head:

1. the convention contract. _convert_doc consumes heads with `[head + i for i, head in enumerate(heads)]`, so __get_tokens_with_heads must emit a RELATIVE head. our term is `offset + word_index_to_token_index.get(token.head - 1)` = the governor's absolute doc index, and `len(tokens)` = the current token's absolute doc index at that point (tokens accumulates across sentences and within the sentence, the current token is appended after). governor absolute minus current absolute is exactly the relative head the consumer expects. copilot's premise that len(tokens) never accounts for the current sentence is wrong, it grows within the sentence loop.

2. the claimed failing test. test_get_tokens_with_heads_collapses_mwt_and_remaps_heads passes on this head, locally and in CI. all 35 checks are green. the 'im gets 3 not 1' numbers reproduce only if len(tokens) were reset per sentence, which it is not.

3. end to end. the real german model through the full engine produces correct spaCy heads (the table upthread: im -> +1 to Büro, Büro -> -3 to treffen, etc), and that path includes the head + i conversion.

also noting copilot marked its own suggested changeset 'unable to commit, now outdated'. with checks green, could you re-approve and resolve the outdated threads? i believe this is otherwise ready.

presidio-analyzer/tests/test_stanza_nlp_engine.py
```diff
@@ -101,3 +128,172 @@ def test_spacy_stanza_english(stanza_pipeline):
...
+    assert tokens[2].lemma == "im"
+    assert tokens[2].upos == "X"
+
+    # Relative heads: "Ich" -> "gehe", "gehe" is root, "im" -> "Schule",
+    # "Schule" -> "gehe", and the second sentence's words are unaffected
+    # by the words collapsed in the first one
+    assert heads == [1, 0, 1, -2, 1, 0]
+
+
+@pytest.mark.skip_engine("stanza_de")
```
omri374, line 205: `skip_engine("stanza_de")` does not prevent the German model download. The marker only skips when `stanza_de` is missing from `nlp_engines`, and that only happens when the `stanza` package is not installed. The analyzer CI job installs with `--all-extras`, so all 5 Python-version jobs will register `stanza_de`, run the 10 German tests, and download the full German package via `DOWNLOAD_RESOURCES` on every run.

Please gate these tests behind an explicit opt-in (e.g. an env var checked when `nlp_engines` registers `stanza_de`), or drop the real-model German tests and rely on `test_get_tokens_with_heads_collapses_mwt_and_remaps_heads`, which already covers the collapse and head remap without a model. If you keep one real-model regression test, `test_spacy_stanza_german_multiword_tokens` is enough; the 8-way contraction parametrize exercises the same branch each time.

---

presidio-analyzer/tests/conftest.py
```diff
@@ -84,6 +87,22 @@ def spacy_nlp_engine(nlp_engines):
...
+        nlp_engine.load()
+    return nlp_engine
+
+
+@pytest.fixture(scope="session")
+def stanza_de_nlp_engine(nlp_engines):
+    nlp_engine = nlp_engines.get("stanza_de", None)
+    if nlp_engine:
+        nlp_engine.load()
+    return nlp_engine
```
omri374, line 107: Both of these session fixtures are dead code:

- `stanza_nlp_engine` is shadowed by module-scoped fixtures of the same name in `test_stanza_recognizer.py` and `test_stanza_batch_processing.py`, and `test_stanza_nlp_engine.py` does not use it, so it never runs.
- `stanza_de_nlp_engine` is not requested by any test; `stanza_pipeline_de` reaches into `nlp_engines` and calls `.load()` itself.

Either remove both, or make them the single load point: have `stanza_pipeline_de` depend on `stanza_de_nlp_engine`, and delete the module-level `stanza_nlp_engine` copies so the English model is loaded once per session instead of once per module (`StanzaNlpEngine.load()` rebuilds the pipeline on every call).

---

presidio-analyzer/tests/test_stanza_nlp_engine.py
```diff
@@ -17,11 +23,32 @@ def tags_equal(act, exp):
...
+        return stanza_de.nlp["de"]
+
+    import stanza
+
+    lang = "de"
+    stanza.download(lang)
+    # Same processors as StanzaNlpEngine.load uses, so that the mwt
+    # processor is part of the pipeline (Stanza adds it for German)
+    nlp = load_pipeline(lang, processors="tokenize,pos,lemma,ner")
+    return nlp
```
omri374: This fallback branch is unreachable. Every test using this fixture is marked `skip_engine("stanza_de")`, which skips before the fixture runs whenever `stanza_de` is absent, and the engine object is always truthy, so `if stanza_de:` never fails. Please remove the branch and take the engine from the `stanza_de_nlp_engine` fixture instead of loading it here.

---

omri374:
Thanks, agree on the copilot false positive, however I'd like the tests to properly use the stanza models to not create too much overhead on tests

ManoharPaturi:
@omri374 all three points addressed:

- german model is now opt-in: stanza_de only registers in nlp_engines when PRESIDIO_TEST_STANZA_DE is set, so the analyzer CI jobs skip the german tests and never download the package unless someone opts in. verified both modes locally: without the var 2 passed 2 skipped, with it the german tests run and pass.
- the session fixtures are now the single load point: the module-scoped stanza_nlp_engine copies in test_stanza_recognizer and test_stanza_batch_processing are removed so the conftest session fixture owns the english model, and stanza_pipeline_de now depends on stanza_de_nlp_engine directly.
- the unreachable fallback branch in stanza_pipeline_de is gone, and the 8-way contraction parametrize is dropped in favor of the single end-to-end regression it duplicated (the trailing-indent regression and the collapse/remap unit test stay).

presidio-analyzer/tests/test_stanza_nlp_engine.py
```diff
@@ -101,3 +123,151 @@ def test_spacy_stanza_english(stanza_pipeline):
...
+    # from the first expanded word
+    assert tokens[2].lemma == "im"
+    assert tokens[2].upos == "X"
+
+    # Relative heads: "Ich" -> "gehe", "gehe" is root, "im" -> "Schule",
+    # "Schule" -> "gehe", and the second sentence's words are unaffected
+    # by the words collapsed in the first one
+    assert heads == [1, 0, 1, -2, 1, 0]
+
+
```
omri374, line 190: Suggestion for testing parts of the logic when there is no model download. It runs whenever `stanza` is installed (the `stanza_en` marker only checks that), so CI covers the multi-word-token path on every leg.

```suggestion


@pytest.mark.skip_engine("stanza_en")
def test_convert_doc_keeps_mwt_surface_text_offsets_and_entities():
    """Test _convert_doc on a real Stanza Document with a multi-word token.

    The Document is built from CoNLL-U with the real Stanza Token, Word
    and Span classes instead of running the German model, so it runs on
    every test run without a model download. The annotations are what the
    German pipeline produces for the text ("im" expanded into "in" + "dem").
    """
    from types import SimpleNamespace

    from spacy.vocab import Vocab
    from stanza.models.common.doc import Span
    from stanza.utils.conll import CoNLL

    text = "Wir treffen uns im Büro mit Thomas Bergmann."
    conllu = "\n".join(
        [
            "1\tWir\twir\tPRON\tPPER\t_\t2\tnsubj\t_\tstart_char=0|end_char=3",
            "2\ttreffen\ttreffen\tVERB\tVVFIN\t_\t0\troot\t_\tstart_char=4|end_char=11",
            "3\tuns\tuns\tPRON\tPRF\t_\t2\tobj\t_\tstart_char=12|end_char=15",
            "4-5\tim\t_\t_\t_\t_\t_\t_\t_\tstart_char=16|end_char=18",
            "4\tin\tin\tADP\tAPPR\t_\t6\tcase\t_\t_",
            "5\tdem\tder\tDET\tART\t_\t6\tdet\t_\t_",
            "6\tBüro\tBüro\tNOUN\tNN\t_\t2\tobl\t_\tstart_char=19|end_char=23",
            "7\tmit\tmit\tADP\tAPPR\t_\t8\tcase\t_\tstart_char=24|end_char=27",
            "8\tThomas\tThomas\tPROPN\tNE\t_\t2\tobl\t_\tstart_char=28|end_char=34",
            "9\tBergmann\tBergmann\tPROPN\tNE\t_\t8\tflat\t_\tstart_char=35|end_char=43",
            "10\t.\t.\tPUNCT\t$.\t_\t2\tpunct\t_\tstart_char=43|end_char=44",
            "",
        ]
    )
    snlp_doc = CoNLL.conll2doc(input_str=conllu)
    snlp_doc.text = text
    sentence = snlp_doc.sentences[0]
    snlp_doc.entities = [
        Span(tokens=sentence.tokens[6:8], type="PER", doc=snlp_doc, sent=sentence)
    ]
    tokenizer = StanzaTokenizer(SimpleNamespace(processors={}), Vocab())

    with warnings.catch_warnings(record=True) as caught:
        warnings.simplefilter("always")
        doc = tokenizer._convert_doc(snlp_doc)

    assert doc.text == text
    assert not [w for w in caught if "multi-word token" in str(w.message)]
    assert [t.text for t in doc] == [
        "Wir",
        "treffen",
        "uns",
        "im",
        "Büro",
        "mit",
        "Thomas",
        "Bergmann",
        ".",
    ]
    for token in doc:
        assert doc.text[token.idx : token.idx + len(token.text)] == token.text
    # the collapsed token keeps its surface form as lemma and borrows
    # the annotations of the first expanded word
    assert doc[3].lemma_ == "im"
    assert doc[3].pos_ == "ADP"
    # dependency heads around the collapsed token point at the right tokens:
    # "im" -> "Büro", "Büro" -> "treffen", "Bergmann" -> "Thomas"
    assert [t.head.i for t in doc] == [1, 1, 1, 4, 1, 6, 1, 6, 1]
    # NER entities keep the character offsets of the original text
    assert [(e.text, e.start_char, e.end_char, e.label_) for e in doc.ents] == [
        ("Thomas Bergmann", 28, 43, "PER")
    ]

```

ManoharPaturi:
all tests are green

presidio-analyzer/tests/conftest.py
```diff
@@ -41,6 +42,12 @@ def nlp_engines(request, nlp_engine_provider) -> Dict[str, NlpEngine]:
             available_engines[f"{name}_en"] = engine_cls(
                 models=[{"lang_code": "en", "model_name": "en"}]
             )
+            # The German model is large; only register the engine (and run the
+            # German model tests) when explicitly opted in.
+            if os.environ.get("PRESIDIO_TEST_STANZA_DE"):
```
omri374: can we replace this with a pytest mark and document it? I'm concerned that this env var would never change to True and this test would be dead code. Another approach (which I added below) is not to download a model at all and test the logic without a full model.

ManoharPaturi:
@omri374 Updated per your recommendation. The German model download, engine registration, and opt-in env flag have been completely removed from conftest.py, and the live German tests were replaced with your CoNLL-U synthetic document unit test in test_stanza_nlp_engine.py. It exercises _convert_doc with real Stanza Token, Word, and Span objects across the multi-word token collapse, surface text offsets, dependency heads, and NER entity boundaries without any model download overhead. Ruff checks pass cleanly.

omri374 (approved):
Thanks!

PR #2254, opened 2026-09-14
fix(analyzer): omitted deny_list_score in YAML registry now defaults to 1.0

### LINKED ISSUE #2242: Omitted deny_list_score yields 0.0 through RecognizerRegistryProvider but 1.0 through PatternRecognizer and add_pattern_recognizer_from_dict

ppradyoth:
## Summary

Fixes #2242.

`CustomRecognizerConfig.deny_list_score` (`presidio_analyzer/input_validation/yaml_recognizer_models.py`) defaulted to `0.0` and was always present in `model_dump()` output, so `RecognizerRegistryProvider` (YAML / dict registry config) passed `deny_list_score=0.0` into `PatternRecognizer` even when the config said nothing about it. A deny-list match scored at `0.0` is filtered out by any positive `score_threshold`, so a YAML deny list without an explicit score effectively never fired through the provider path — while the same recognizer built via `PatternRecognizer()` directly, or `RecognizerRegistry.add_pattern_recognizer_from_dict()`, correctly defaulted to `1.0`.

## Fix

- Changed the field default to `None`.
- Added a `model_dump()` override on `CustomRecognizerConfig` that sets `exclude_none=True` by default, matching the existing pattern already used by `HuggingFaceRecognizerConfig` / `GLiNERRecognizerConfig` / `LangExtractRecognizerConfig` in the same file.

Excluding the key entirely (not just setting it to `None`) matters here specifically: `PatternRecognizer.from_dict()` forwards the dumped dict straight into the constructor via `cls(**dict)`, so passing an explicit `deny_list_score=None` would set the score to `None` outright rather than falling back to the constructor's own default of `1.0`.

## Behavior change

Deny-list matches loaded through `RecognizerRegistryProvider` (YAML or dict config) without an explicit `deny_list_score` now score `1.0` instead of `0.0`. Anyone relying on the old `0.0` default should set `deny_list_score` explicitly. Called out in `docs/analyzer/recognizer_registry_provider.md`.

## Testing

- Added unit tests in `test_yaml_recognizer_models.py` covering that an omitted `deny_list_score` stays `None` and is excluded from `model_dump()`, and that an explicit value is preserved.
- Added end-to-end tests in `test_recognizer_registry_provider.py` covering the full `RecognizerRegistryProvider` → `PatternRecognizer` path for both the omitted and explicit cases.
- Ran the exact repro script from #2242 against the fix — all three loading paths (`RecognizerRegistryProvider`, direct `PatternRecognizer()`, `add_pattern_recognizer_from_dict()`) now agree at `1.0`.
- `pytest tests/test_yaml_recognizer_models.py tests/test_recognizer_registry_provider.py`: 103 passed. The 2 pre-existing failures in that run (`transformers` not installed in this environment) are unrelated and fail identically on `main`.
- `ruff check` on the changed files: no new issues (2 pre-existing docstring lint findings elsewhere in `test_yaml_recognizer_models.py`, untouched by this change).

This fix was scoped, triaged, reviewed, and submitted by me, with implementation assistance from Claude Code.


presidio-analyzer/presidio_analyzer/input_validation/yaml_recognizer_models.py
```diff
@@ -365,7 +365,7 @@ class CustomRecognizerConfig(BaseRecognizerConfig):
         default=None, description="Words to deny/exclude"
     )
     deny_list_score: Optional[float] = Field(
-        default=0.0, ge=0.0, le=1.0, description="Deny list score"
+        default=None, ge=0.0, le=1.0, description="Deny list score"
```
omri374: why not default to 1.0 and avoid the `exclude_none` default?

ppradyoth: Good call, done. Just pushed a03886a6 that defaults the field to `1.0` directly and drops the `model_dump()` override entirely. Simpler and it was unnecessary: since the constructor's own default is also `1.0`, an explicit `1.0` and an omitted key now produce the same result, so there's no need to distinguish "not set" from "set to the default" anymore.

omri374 (approved):

Issue #2256, opened 2026-09-15
/analyze silently drops an unsupported entity when other requested entities are supported; IT_FISCAL_CODE unreachable in the default image


GiorgioDotcom:
Two related behaviors observed on presidio-analyzer 2.2.364 (ghcr.io/data-privacy-stack/presidio-analyzer:2.2.364, default configuration, language en).

1. Unsupported entity silently ignored

    POST /analyze
    {"text":"codice fiscale RSSMRA85M01H501Q, IBAN IT60X0542811101000000123456","language":"en","entities":["IT_FISCAL_CODE","IBAN_CODE"]}

    -> 200, [{"entity_type":"IBAN_CODE", ...}]

IT_FISCAL_CODE is not in the response and nothing signals that it could not be served. Compare:

    {"text":"...","language":"en","entities":["IT_FISCAL_CODE"]}

    -> 500, {"error":"No matching recognizers were found to serve the request."}

So the caller gets an error only when the unsupported entity is the sole one requested. In a gateway or DLP integration this is a footgun: an operator adds IT_FISCAL_CODE to the block list, the service keeps answering 200, and fiscal codes flow through unscanned. I measured exactly this on two LLM gateways that embed Presidio (LiteLLM and GoModel): 11 out of 11 positions leaked the fiscal code while credit card, email and IBAN were blocked. Reproduction and raw results: https://github.com/GiorgioDotcom/llm-gateway-privacy-bench (risultati/gomodel-0.1.91-presidio-cf.json and risultati/litellm-1.98.0-presidio-cf.json; the one-line registry override is in presidio-cf/recognizers-cf.yaml, results with it in risultati/*-riccf-*.json).

Suggestion: return 400 when any requested entity has no recognizer for the requested language, or at least include the list of unserved entities in the response (or a warning header), so integrations can fail closed.

2. IT_FISCAL_CODE unreachable by default

ItFiscalCodeRecognizer exists and works well (it handles omocodia and validates the check character), but conf/default_recognizers.yaml registers it with supported_languages: [it], while the published image loads only the en NLP model and supported_languages: [en]. Result: with the image as shipped there is no way to detect Italian fiscal codes without overriding the registry.

The one-line override works (RECOGNIZER_REGISTRY_CONF_FILE pointing to a copy of default_recognizers.yaml where ItFiscalCodeRecognizer has supported_languages: [en]); after that GoModel with Presidio caught the fiscal code in every position, including omocodic codes.

[...]

PR #2259, opened 2026-09-16
Warn about unsupported requested entities while preserving partial results

nikolas-sapa:
## Summary

Refs #2256. Stage 1 per your review: warning-first, no behavior change.

Mixed supported/unsupported entity requests still return partial results with HTTP 200. The registry warning now states ignoring unsupported entities is deprecated and will raise in a future version, names the entity and language, and points to `AnalyzerEngine.get_supported_entities(language)` or a matching ad-hoc recognizer.

## Design choices

- Warning text lives in `RecognizerRegistry.get_recognizers`, the single choke point, so engine + REST inherit it.
- Docstrings on `get_recognizers` and `AnalyzerEngine.analyze` document the deprecation and current contract.
- No REST status change, no config flag, no release-date claim.

## Compatibility

- Supported entities return identical results in mixed requests.
- Requests with zero matching recognizers keep the existing ValueError/message.
- Ad-hoc recognizers remain included in matching.

## Verification

- `tests/test_recognizer_registry.py`: 54 passed
- `tests/test_analyzer_engine.py`: 61 passed
- `tests/test_app.py` (new, service-layer preservation): 4 passed
- `ruff check` on touched source files + `git diff --check`: clean
- Not run: full analyzer suite, model-dependent tests, live-server E2E (the added E2E case asserts preserved 200 behavior but is not claimed as live-verified).

## Scope

No country-recognizer language config changes, no strictness option, no changelog edit. Single squashed commit, 6 files, +217/-2.


omri374:
As disccused on #2256, this is not backward compatible. It's best if this is done in stages, where the first stage is a warning + a notice that this would raise a warning in future versions. wdyt?

nikolas-sapa:
Updated per your review: this is now the warning-first stage only. The earlier hard-error commit is squashed away — mixed requests preserve partial results with HTTP 200, and the registry warning carries the deprecation notice with entity/language plus a pointer to `get_supported_entities`. Unit suites re-verified after the squash (registry 54, engine 61, app 4; ruff + diff-check clean). E2E case added for the preserved-200 behavior but not live-verified. Let me know if the warning wording needs adjusting.

presidio-analyzer/presidio_analyzer/analyzer_engine.py
```diff
@@ -207,6 +207,13 @@ def analyze(
         :param regex_flags: regex flags to be used for when allow_list_match is "regex"
         :param nlp_artifacts: precomputed NlpArtifacts
         :return: an array of the found entities in the text
+        :raises ValueError: if no recognizers match the requested entities and language.
```
omri374: Does it raise a ValueError?

nikolas-sapa:
Good catch on the `:raises:` — fixed in 82c6405: `analyze()` warns and preserves partial results, raising only when nothing matches at all. Removed the overclaimed line in `analyze()` and clarified the registry docstring to warn-first. Re-verified (engine 61, registry 54, ruff clean).

omri374 (approved):

Issue #739, opened 2021-07-02
Improve PatternRecognizer to respect multiple match groups


mprojlb:
PatternRecognizer treats complex match groups as one match group, this is not desirable behaviour. It would be preferable if all match groups were individually considered a match. 

Example:
```python
from presidio_analyzer import PatternRecognizer, Pattern

text = 'this is an example string with many words'

patterns = [
  # regex with non match group
  Pattern('Non match group', r'\b(is)(?: an)', 0.9),
  # regex with non match group
  Pattern('Multiple match groups', r'(string).*?(words)', 0.9),
]

pr_wrong = PatternRecognizer(supported_entity='X', patterns=patterns)
pr_right = MultiPatternRecognizer(supported_entity='X', patterns=patterns)

print('Wrong')
for m in pr_wrong.analyze(text, entities='X'):
  t = text[m.start: m.end]
  print(m, f'- "{t}"')
  
print('\nCorrect')
for m in pr_right.analyze(text, entities='X'):
  t = text[m.start: m.end]
  print(m, f'- "{t}"')
```
Output:
```
Wrong
type: X, start: 5, end: 10, score: 0.9 - "is an"
type: X, start: 19, end: 41, score: 0.9 - "string with many words"

Correct
type: X, start: 5, end: 7, score: 0.9 - "is"
type: X, start: 19, end: 25, score: 0.9 - "string"
type: X, start: 36, end: 41, score: 0.9 - "words"
```

A simple change to `PatternRecognizer.__analyze_patterns` something like this should work:
From:
```python
for match in matches:
    start, end = match.span()
    current_match = text[start:end]
```
To:
```python
for match in matches:
    # Loop through each match group if more than one
    offset = 0 if len(match.regs) == 1 else 1
    for start, end in match.regs[offset:]:
        current_match = text[start:end]
```

omri374:
Thanks @mprojlb, we'll look into this!

omri374:
Hey @mprojlb, thanks for this! Quick question: The first pattern should detect both "is" and " an". However on the proposed change the substring " an" is not detected. Same goes for the "with many" substring. Is this the expected behavior? Checking my understanding of the issue.

SharonHart:
@omri374 I think that 'an' is encapsulated in a non-capturing group (**?:** an)
In that case, the regex will match on 'an', but it will be excluded from the results

mprojlb:
As @SharonHart states groups that start `(?:` are non-capturing. 

[...]

PR #2260, opened 2026-09-16
feat(analyzer): report a regex capture group instead of the whole match

### LINKED ISSUE #1120: Use group from matched pattern (PatternRecognizer)
In the `PatternRecognizer`, it is not possible to use the span of a given group in a pattern, rather than using the entire span.
For example using regex `"Password: (\w)", the `PatternRecognizer` will anonymize "Password: 1234", but I would like to anonymize only "1234"

A way to specify a group to use, e.g., `start, end = match.span(pattern.use_group) if pattern.use_group else match.span()`.
Can do a PR if needed.

v0ropaev:
## Change Description

Adds an optional `capture_group` to `Pattern`, so a `PatternRecognizer` can report one capture group instead of the whole regex match. For example `password:\s*(?P<value>\S+)` with `capture_group="value"` detects `hunter2`, not `password: hunter2`.

**Root cause.** `PatternRecognizer.__analyze_patterns` always takes `match.span()`, so whatever a regex needs as an anchor ends up inside the entity. The only ways around it today are a variable-width lookbehind or a subclass with its own matching loop (which is what the reporter of #1120 ended up doing).

**What changes**

| Surface | Change |
|---|---|
| `Pattern` | new kwarg `capture_group: Optional[Union[int, str]] = None`, a group number or name. Checked at construction against the compiled regex: bool, negative, out-of-range and unknown names raise `ValueError`. Messages never include matched text |
| `Pattern.to_dict` | emits `capture_group` only when it is set, so serialised output of existing patterns stays byte-identical (checked for all 108 predefined pattern recognizers) |
| `PatternRecognizer` | uses `match.span(capture_group)`. Matches where the group doesn't participate are dropped by the existing empty-match check, and `validate_result` / `invalidate_result` receive the group text |
| `CustomRecognizerConfig` (YAML) | validates `capture_group` at parse time when it is present. YAML without it is validated exactly as before |
| ad-hoc REST recognizers | work through the existing `from_dict`, no loader change |
| docs | `api-docs.yml` Pattern schema, a "Detecting part of a regex match" section in `adding_recognizers.md`, `recognizer_registry_provider.md` |

**Behaviour changes**

- None unless `capture_group` is set: `match.span(0)` is `match.span()`, and an `/analyze` response with `return_decision_process` is identical to `main`.
- Configurations that set `capture_group` used to fail (`TypeError` in `Pattern.__init__`, HTTP 400 for ad-hoc recognizers) and now work.
- With a narrowed span, words the regex matched before the group become surrounding text, so a label like `password:` that is also a context word raises the score. Documented and covered by a test.
- The group is validated against the regex without flags. If the flags used at analysis time remove it (e.g. `re.VERBOSE` turning `#...` into a comment), the pattern is skipped with a warning (pattern name and group only), once per `analyze` call and regardless of the input text. A group *number* can also shift under such flags, so the docs recommend named groups there.
- `AnalysisExplanation` is unchanged, since the score formula is unchanged.

**Design decisions**

- **One group per pattern.** #739 (reporting every capture group) was closed as not planned, and reading groups implicitly would silently change results for the many regexes that wrap the whole pattern in `( ... )`, including the zip example in the docs. Several groups → several patterns on the same regex.
- **`capture_group` rather than `use_group`** from the issue: I think it reads better in YAML and JSON. Happy to rename.
- **Why not just lookbehind.** The `regex` module does support `(?<=password:\s*)\S+`, and a few predefined recognizers use exactly that. Still, a group keeps the regex portable and readable, is settable from YAML/REST without lookbehind tricks, fails early with a clear message, and hands the validation hooks only the value.

**Not in scope**

- Recognizers that replace the matching loop (`IbanRecognizer`) ignore `capture_group`. Documented.
- Flag-aware validation. The flags come from the recognizer, the registry config or the request, and `Pattern` doesn't see any of them at construction.
- The HTTP status for an invalid `capture_group` in an ad-hoc recognizer is 500, same as an invalid `score` today (`ValueError` → 500 in `app.py`).

### Verification

In `presidio-analyzer`, Python 3.12, `uv sync --locked --all-extras --group dev` plus `en_core_web_lg` / `en_core_web_sm`:

- `pytest tests/test_pattern.py tests/test_pattern_recognizer.py tests/test_yaml_recognizer_models.py tests/test_recognizer_registry_provider.py tests/test_analyzer_request.py tests/test_context_support.py`: **211 passed**. The same tests against `main`'s sources: 31 failed.
- Full suite the way CI runs it (`pytest --cov=presidio_analyzer`): **3498 passed, 13 skipped** (`main`: 3463 passed, 13 skipped; the difference is the 35 new tests).
- `diff-cover coverage.xml --compare-branch=origin/main --fail-under=90`: 41 changed lines, **100%**.
- `ruff check` with ruff 0.9.2 as in CI: clean.
- e2e `tests/test_api_analyzer.py -k ad_hoc` against a local analyzer started with the Dockerfile's conf: 5 passed (the new test gets a 400 on `main`). I didn't run the Docker e2e itself.
- `api-docs.yml` passes `openapi-spec-validator`.
- Not run locally: the 3.10 / 3.11 / 3.13 / 3.14 legs and the mkdocs build.

The new tests cover spans for a number, a name, `0` and unset; non-participating and empty groups; what the hooks receive; two patterns on one regex; the flags warning (logged once per call, no PII in the log); the `to_dict` / `from_dict` round trip; YAML through `RecognizerRegistryProvider`; ad-hoc request dicts; and the context-word interaction.

## Issue reference

Fixes #1120

## Checklist


v0ropaev:
Good catch from the Copilot review (it ended up in the suppressed comments, so replying here): the check for a group removed by the regex flags lived inside `for match in matches`, so a misconfigured pattern only warned when the text happened to contain a match of the whole regex. On an input without a candidate it stayed silent, which contradicts what the docs promise.

Fixed in ff16851:

- the group is now checked against the compiled regex (`groups` / `groupindex`) once per pattern, before `finditer`, and the pattern is skipped with the same warning. The match loop is back to a plain `match.span(group)` with no `try/except IndexError`;
- the check only runs when `capture_group` is set, so patterns without it never touch the compiled regex's groups (the existing regex-timeout tests patch `re.compile` with a `MagicMock` and keep passing);
- `test_when_regex_flags_remove_capture_group_then_pattern_is_skipped_with_warning` is parametrized over a numbered and a named group and now also asserts the warning on text with no candidate at all. That assertion fails on the previous commit.

Verification: the 6 touched test modules 211 passed; full `presidio-analyzer` suite with `--cov` 3498 passed, 13 skipped; `diff-cover --fail-under=90` 41 changed lines, 100%; `ruff check` (0.9.2) clean. PR description numbers updated.

Issue #1511, opened 2025-01-16
GitHub-Action with SARIF export


20001593:
**problematic**

We need presidio to have a github-action template to be able to launch the library directly from the CI/CD of a github repository on the contents of the repository in case of PR triggering on the main branch or protected branches

A workflow file with a SARIF export should me perfect.

omri374:
There's a 3rd party GH action for Presidio, have you looked into it? https://github.com/marketplace/actions/presidio-action

20001593:
Yes, it's a very good start.
But there's no SARIF export and i prefer a validated and complet template.
_(As Github is now Microsoft part, i hope about synergie between them)_

If we do it, we will share it here, but i prefer note our intention before start.

PR #2266, opened 2026-09-17
fix(cli): fix --no-warnings, exit code, github output and threshold config

### LINKED ISSUE #2265: presidio-cli: --no-warnings crashes, exit code is always 0, -f github creates no annotations, config threshold is not validated
Four bugs in presidio-cli on main (f251c513). Setup, following AGENTS.md:

    git clone https://github.com/data-privacy-stack/presidio
    cd presidio/presidio-cli
    uv sync --locked --all-extras --group dev
    uv run python -m spacy download en_core_web_lg
    CLI_PROJECT=$PWD

Run everything below from an empty directory. Inside presidio-cli/ the CLI loads presidio-cli/.presidiocli, which has no URL entity, so the two URL findings shown below would be missing.

    mkdir /tmp/presidio-cli-repro && cd /tmp/presidio-cli-repro
    presidio() { uv run --project "$CLI_PROJECT" presidio "$@"; }
    printf 'Hello Paulo Santos.\nThe latest statement for your credit card account 4111 1111 1111 1111\nwas mailed to 123 Any Street, Seattle, WA 98109.\nContact: john.smith@example.com\n' > pii.txt
    printf 'nothing to see here\n' > clean.txt

### 1. `--no-warnings` crashes when a file has a finding

    $ presidio --no-warnings -f standard pii.txt
    Traceback (most recent call last):
      ...
      File ".../presidio_cli/cli.py", line 135, in show_problems
        if no_warn and (problem.level != "error"):
                        ^^^^^^^^^^^^^
    AttributeError: 'PIIProblem' object has no attribute 'level'

`show_problems()` reads `problem.level`, but `PIIProblem` never sets it. The unit tests miss this because the `problems` fixture in `tests/test_cli.py` is a `mocker.Mock`, which creates `.level` on access.

### 2. The exit code is always 0

    $ presidio -f standard pii.txt; echo "exit=$?"
    pii.txt
      1:7       0.85     PERSON
      2:51      1.0      CREDIT_CARD
      3:31      0.85     LOCATION
      3:40      0.85     LOCATION
      4:10      1.0      EMAIL_ADDRESS
      4:10      0.5      URL
      4:21      0.5      URL

    exit=0
    $ presidio -f standard pii.txt clean.txt > /dev/null; echo "exit=$?"
    exit=0
    $ presidio -f standard - < pii.txt > /dev/null; echo "exit=$?"
    exit=0

`run()` ends with `if prob_num > 0: return_code = 1`, so exiting 1 on findings is intended. But `show_problems()` returns `max_level = 0`, which is never updated, and `run()` overwrites `prob_num` for every file instead of adding to it. The tests lock this in: `test_show_problems` expects `rc == 0` for two problems, and `test_run_current_dir` expects exit 0 for a workspace that contains a credit card number.

### 3. `-f github` does not create annotations

    $ presidio -f github pii.txt
    ::group::pii.txt
    ::0.85 file=pii.txt,line=1,col=7::1:7 [PERSON]
    ::1.0 file=pii.txt,line=2,col=51::2:51 [CREDIT_CARD]
    ::0.85 file=pii.txt,line=3,col=31::3:31 [LOCATION]
    ::0.85 file=pii.txt,line=3,col=40::3:40 [LOCATION]
    ::1.0 file=pii.txt,line=4,col=10::4:10 [EMAIL_ADDRESS]
    ::0.5 file=pii.txt,line=4,col=10::4:10 [URL]
    ::0.5 file=pii.txt,line=4,col=21::4:21 [URL]
    ::endgroup::

`::0.85` is not a workflow command. The Actions runner only handles registered commands such as `warning`, `error` and `notice` (`ActionCommand.TryParseV2` returns false for any other name), so these lines are printed as plain log text. The documented syntax is `::warning file={name},line={line},col={col}::{message}` (https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands). Property values and the message are not percent-escaped, so a file name containing a newline is printed as a separate log line, which the runner can read as a workflow command. With a directory argument (`presidio -f github .`) the file property is also `./pii.txt`, and the runner passes relative paths through unchanged.

### 4. `threshold` in the config file is not validated

    $ presidio -d "threshold: 5" -f standard pii.txt; echo "exit=$?"
    exit=0
    $ presidio -d "threshold: abc" -f standard pii.txt
    Traceback (most recent call last):
      ...
      File ".../presidio_cli/config.py", line 108, in parse
        self.threshold = float(conf["threshold"])
    ValueError: could not convert string to float: 'abc'

`PresidioCLIConfig.parse()` range-checks `self.threshold`, which still holds the default 0, and only afterwards assigns the configured value. So `threshold: 5` is accepted and hides every finding, and a non-number raises an uncaught `ValueError`. The `--threshold` flag already rejects 5 (`threshold must be between 0.0 and 1.0`, exit 2).

### Root cause

The CLI came from insightsengineering/presidio-cli, a yamllint port, and was imported in #918. The upstream initial commit already has `line += str(problem.score)  # TODO score to level` and `# max_level+=1` commented out, so level and exit-code handling was never finished. The config check has tested the wrong variable since the upstream "Threshold filtering" commit.

### Expected behavior

1. Each finding has a level: `error` for score 1.0 and `warning` below it, the same split the colored output already uses (red/yellow). `--no-warnings` prints only errors.
2. The exit code is 1 when at least one finding is printed (across all files, or stdin), and 0 otherwise.
3. `-f github` prints `::warning` / `::error` commands with escaped values and a plain relative file path, so GitHub Actions creates annotations.
4. An invalid `threshold` in the config fails with a clear `PresidioCLIConfigError` message and exit code 1.

v0ropaev:
## Change Description

Fixes four presidio-cli bugs described in #2265:

- **`--no-warnings` crash**: `PIIProblem` now has `level`: `"error"` when the score is 1.0, `"warning"` otherwise. This is the split `Format.standard_color` already used, and it now reads `problem.level`.
- **Exit code**: `show_problems()` returns the number of problems it printed (after `--no-warnings` filtering), and `run()` adds these up across files and stdin, so the existing `if prob_num > 0` check works.
- **`-f github`**: prints `::warning` / `::error` workflow commands, with property values and the message escaped the way `@actions/core` does it (`%`, CR, LF, plus `:` and `,` in properties). The `./` prefix that `presidio .` adds is removed from the `file` property, as `run()` already does before `analyze()`. Example: `::error file=pii.txt,line=2,col=51::2:51 [CREDIT_CARD] score=1.0`. Escaping also means a file name with a newline can no longer print its own workflow command into the log.
- **Config `threshold`**: the configured value is converted and range-checked before it is stored. Out-of-range values (`threshold: 5`), non-numbers (`abc`), YAML booleans (`true`, which `float()` would read as 1.0) and integers too large for a float now raise `PresidioCLIConfigError` instead of being accepted silently or crashing.

Tests: the Mock-based `problems` fixture, which hid the missing attribute, now uses real `PIIProblem` / `RecognizerResult` objects. Tests that asserted the old return value and exit code (`rc == 0`, exit 0 on a workspace with a card number) are updated. Each bug has new tests that fail on main and pass with this change; the color-split and valid-threshold tests are regression guards that pass on both. README: corrected github example, plus new "Warnings and errors" and "Exit codes" sections.

### Behavior changes

- `presidio` now **exits 1 when it reports at least one finding**, warnings included (it always exited 0 before). CI jobs or hooks that run it will fail on findings; `presidio . || true` keeps the old behavior. If you would rather follow yamllint (exit 1 only for errors, warnings exit 0 unless `--strict`), I can change it.
- `--no-warnings` prints only score-1.0 findings (before, it crashed on the first finding).
- `-f github` (and `auto` on GitHub Actions) output changes from `::<score> file=...::L:C [TYPE]` to `::warning|error file=...::L:C [TYPE] score=S` and now creates annotations. For stdin the file property is still `stdin`; that can be a follow-up.
- An invalid config-file `threshold` is now a config error (exit 1 with a message).
- `standard`, `colored` and `parsable` output text is unchanged.

Scoring 1.0 as "error" is a judgment call that follows the existing colored output. I can make it configurable if you prefer. The two commits are independent (the config threshold fix is the second one) and can be split into separate PRs if that is easier to review.

## Issue reference

Fixes #2265

## Checklist


omri374:
Thanks! I'll approve the PR, but we will be deprecating the presidio-cli component sometime soon. It's better to use the analyzer directly

omri374 (approved):

v0ropaev:
Thanks for the quick review.

Good to know about the deprecation — that changes what is worth doing next, so: I will leave #1511 (a `sarif` output format for the CLI) alone rather than sending it, since it would be new surface on a component you are retiring. If you would rather point people at the analyzer with a short snippet in the CLI README on the way out, I am happy to do that instead.

One thing worth a line in the deprecation note when it happens: this PR makes `presidio` exit 1 whenever it reports a finding, where it always exited 0 before. Anyone who has it in a pre-commit hook or CI job will start seeing failures, and `presidio . || true` is the one-line escape hatch.

My other open PR here, #2260 (`capture_group` on `Pattern`), is in the analyzer rather than the CLI, so the deprecation does not affect it.

omri374:
Thanks. While the presidio-cli code might change, the pypi package (which I would assume is used in the presidio action) wouldn't as it's a 3rd party package and not part of data privacy stack.

PR #2271, opened 2026-09-19
docs: enable country pattern recognizers on en image via RECOGNIZER_REGISTRY_CONF_FILE

zelihaguven:
## Summary

Documents how operators can enable country-specific pattern recognizers
(e.g. `ItFiscalCodeRecognizer` → `IT_FISCAL_CODE`) on the default English
`presidio-analyzer` image by overriding the registry with
`RECOGNIZER_REGISTRY_CONF_FILE`.

Defaults stay unchanged: native-language registration remains the default to
limit false positives (per discussion on #2256).

Also notes that without an override, requesting an unsupported entity together
with supported ones can return HTTP 200 while silently omitting the unsupported
entity — docs/operator guidance only; no analyzer behavior change in this PR.

Fixes part of #2256 (docs). Out of scope: #2259 warning/400 path.

## Changes

- Main how-to: `docs/analyzer/recognizer_registry_provider.md` (new section)
- Docker env vars: `docs/installation.md`
- Cross-links: `docs/supported_entities.md` (Italy), `docs/analyzer/languages.md`,
  `docs/analyzer/filtering_by_country.md`, `docs/analyzer/index.md`,
  `docs/analyzer/analyzer_engine_provider.md`

## Test plan

- [ ] Preview MkDocs: section anchors and internal links resolve
- [ ] Confirm Docker example matches `app.py` env var names
- [ ] Confirm snippet matches `default_recognizers.yaml` entry shape
- [ ] No code / registry default changes in this PR


GiorgioDotcom:
Thanks for writing this up. I opened #2256 and had offered to do the docs part, but I had not started a PR, so this covers it and there is no overlap.

I ran the documented steps against ghcr.io/data-privacy-stack/presidio-analyzer:2.2.364 rather than just reading them:

- The main path works exactly as written: copy of default_recognizers.yaml with `ItFiscalCodeRecognizer` set to `en`, mounted at `/app/recognizers.yaml` with `RECOGNIZER_REGISTRY_CONF_FILE`. `IT_FISCAL_CODE` comes back with score 1.0 for `language: en`, and it shows up in `GET /supportedentities?language=en`. The "replaces, not merges" note is correct and worth keeping prominent.
- The silent-drop warning matches what I measured for #2256 (200 with `IBAN_CODE` only, 500 when `IT_FISCAL_CODE` is requested alone).

One sentence does not hold on the default image: "You can list both `en` and `it` if the same deployment also serves Italian (`language: it`) requests."

- With `supported_languages: [en, it]` on the recognizer, `language: en` works but `language: it` still returns 500 "No matching recognizers were found to serve the request." The startup log explains it: "Recognizer not added to registry because language is not supported by registry - ... supported languages: it, registry supported languages: en".
- Adding `it` to the registry's top-level `supported_languages` as well does not fix it: the container exits at boot with "Misconfigured engine, supported languages have to be consistent registry.supported_languages: ['en', 'it'], analyzer_engine.supported_languages: ['en']".

So serving `language: it` needs the analyzer and NLP engine configured for Italian too (`ANALYZER_CONF_FILE` / `NLP_CONF_FILE` with an Italian model), which is a different setup from the one this section describes. I have not tested that last part, I am only reading it off the error message. Suggested wording:

> Listing `it` next to `en` on the recognizer is harmless but has no effect on the default image: requests with `language: it` are still rejected, because the registry and the NLP engine only support `en`. To serve Italian requests as well, configure the analyzer and NLP engine for `it` (see [languages](languages.md)); adding `it` to the registry alone makes the analyzer fail at startup with a "supported languages have to be consistent" error.

Two optional details from the same measurements, in case they are useful for the section: with the override the recognizer also detects omocodic codes (RSSMRA85M01H50MI, score 1.0), and a code with a wrong check character is still reported but with score 0.3 instead of 1.0, so integrations that apply a score threshold should know the two cases differ.

Files and exact results: https://github.com/GiorgioDotcom/llm-gateway-privacy-bench/tree/master/presidio-cf/verifica-pr2271

zelihaguven:
Thank you @GiorgioDotcom — really appreciate you verifying the steps on `presidio-analyzer:2.2.364` and catching the incorrect `en`/`it` sentence.

I pushed a follow-up commit that:
- replaces that wording with your suggested clarification (default image stays `en`-only; serving `language: it` needs analyzer + NLP config, not just the recognizer list)
- keeps the `country_code` / filtering note
- adds the optional omocodia (score `1.0`) vs wrong check character (score `~0.3`) notes from your measurements

Happy to adjust further if anything still looks off.

GiorgioDotcom:
Read a7adfea: the en/it paragraph now matches what the image actually does, and the omocodia / check-character note is accurate. Nothing else from my side. Thanks for the quick turnaround, and for the credit.

zelihaguven:
Closing this PR for now — may revisit later. Thanks for anyone who looked.

PR #2280, opened 2026-09-20
fix(analyzer): avoid quadratic duplicate result checks

### LINKED ISSUE #2279: EntityRecognizer.remove_duplicates scales poorly for large result sets
**Describe the bug**
`EntityRecognizer.remove_duplicates` becomes very slow when processing large result sets. Processing 20,000 non-overlapping results takes approximately 27 seconds on my machine.

The implementation compares each result against previously retained results, resulting in quadratic time complexity even when results do not overlap.

**To Reproduce**
Run the following code:
``` py
import time

from presidio_analyzer import EntityRecognizer, RecognizerResult

results = [
    RecognizerResult(
        entity_type="x",
        start=i * 10,
        end=i * 10 + 9,
        score=0.5,
    )
    for i in range(20_000)
]

start = time.perf_counter()
results = EntityRecognizer.remove_duplicates(results)
duration = time.perf_counter() - start

print(f"Processed {len(results)} results in {duration:.2f} seconds")
```

On my machine, this takes approximately 27 seconds.

**Expected behavior**
`EntityRecognizer.remove_duplicates` should efficiently process large result sets, particularly when most results do not overlap. Runtime should not grow quadratically with the number of results.

**Screenshots**
N/A

**Additional context**
Large documents or inputs producing many recognizer matches can trigger this behavior.

This method is used frequently in the analyzer flow:

 * At the end of each `PatternRecognizer.analyze` call.
 * By `AnalyzerEngine.analyze` when combining results from all recognizers.

Its performance can therefore significantly affect overall analysis time for result-dense inputs.

mbukowicz:
## Change Description

Optimizes `EntityRecognizer.remove_duplicates` by limiting containment checks to overlapping results of the same entity type. This improves performance for large, mostly non-overlapping result sets. Filtering semantics and primary ordering by score, start, and span length remain unchanged. Results tied on all three fields are now deterministically ordered by their stringified entity type instead of process-dependent set iteration order.

## Issue reference

Fixes #2279

## Checklist


mbukowicz:
A Fenwick tree-based implementation would be more efficient in the worst case, but it would also add complexity and make the code harder to reason about:
``` py
class _ContainingSpanIndex:
    """Index spans for efficient containment checks.

    Start positions are coordinate-compressed and stored in a Fenwick tree
    whose prefix maximum is the furthest end position seen at or before a
    candidate's start.
    """

    @staticmethod
    def _fenwick_stride(index: int) -> int:
        """Return the size of the Fenwick segment ending at index."""
        return index & -index

    @classmethod
    def _fenwick_parent_index(cls, index: int) -> int:
        """Return the parent index visited during a Fenwick prefix query."""
        return index - cls._fenwick_stride(index)

    @classmethod
    def _fenwick_next_index(cls, index: int) -> int:
        """Return the next index visited during a Fenwick point update."""
        return index + cls._fenwick_stride(index)

    def __init__(self, starts: List[int]) -> None:
        unique_starts = sorted(set(starts))
        self._start_indexes = {
            start: index + 1 for index, start in enumerate(unique_starts)
        }
        self._max_end_tree = [-1] * (len(unique_starts) + 1)

    def contains(self, start: int, end: int) -> bool:
        """Return whether an indexed span contains ``(start, end)``."""
        index = self._start_indexes[start]
        prefix_max_end = -1
        while index > 0:
            prefix_max_end = max(prefix_max_end, self._max_end_tree[index])
            index = self._fenwick_parent_index(index)
        return prefix_max_end >= end

    def add(self, start: int, end: int) -> None:
        """Add a span to the index."""
        index = self._start_indexes[start]
        while index < len(self._max_end_tree):
            self._max_end_tree[index] = max(self._max_end_tree[index], end)
            index = self._fenwick_next_index(index)


class EntityRecognizer:
    ...

    @staticmethod
    def remove_duplicates(results: List[RecognizerResult]) -> List[RecognizerResult]:
        """Remove duplicate and contained results.

        Results with a score of zero are removed. For each entity type, a result
        is removed when it is contained in a result with an equal or higher score.
        :param results: List[RecognizerResult]
        :return: List[RecognizerResult]
        """
        results = list(set(results))
        results = sorted(results, key=lambda x: (-x.score, x.start, -(x.end - x.start)))

        starts_by_entity_type: Dict[str, List[int]] = {}
        for result in results:
            if result.score == 0:
                continue
            starts_by_entity_type.setdefault(result.entity_type, []).append(
                result.start
            )

        span_index_by_entity_type: Dict[str, _ContainingSpanIndex] = {}
        for entity_type, starts in starts_by_entity_type.items():
            span_index_by_entity_type[entity_type] = _ContainingSpanIndex(starts)

        filtered_results: List[RecognizerResult] = []
        for result in results:
            if result.score == 0:
                continue

            # Results are processed by descending score, so the index contains
            # only results with an equal or higher score. A prefix maximum end
            # at or beyond this end means that a containing result exists.
            span_index = span_index_by_entity_type[result.entity_type]
            if span_index.contains(result.start, result.end):
                continue

            filtered_results.append(result)
            span_index.add(result.start, result.end)

        return filtered_results
```

I think this simpler approach is enough since overlaps are uncommon in practice.

presidio-analyzer/tests/test_entity_recognizer.py
```diff
@@ -124,11 +126,110 @@ def test_when_remove_duplicates_contained_shorter_length_results_removed():
...
+        )
+        for i in range(20_000)
+    ]
+
+    start_time = time.perf_counter()
+    results = EntityRecognizer.remove_duplicates(results)
+    duration = time.perf_counter() - start_time
+
+    assert len(results) == 20_000
+    assert duration < 5
```
mbukowicz: On my machine, this optimization reduced the duration from 27s to ~0.025s.

omri374: Cool, thanks!

presidio-analyzer/presidio_analyzer/entity_recognizer.py
```diff
@@ -283,28 +283,52 @@ def remove_duplicates(results: List[RecognizerResult]) -> List[RecognizerResult]
         :return: List[RecognizerResult]
         """
         results = list(set(results))
-        results = sorted(results, key=lambda x: (-x.score, x.start, -(x.end - x.start)))
+        results = sorted(
+            results,
+            key=lambda x: (
+                x.entity_type,
```
omri374: Sorting on `entity_type` can aise a `None` or non-string types.

The previous implementation only ever compared entity types with `==`. Putting `x.entity_type` first in the sort key means any result whose type is `None` (e.g. `RecognizerResult.from_json({"start": 0, "end": 5, "score": 0.5})`, where `entity_type` defaults to `None`) or a custom recognizer returning a non-string type now aborts the whole `AnalyzerEngine.analyze` call.

Reproduced on this head:

```python
rs = [RecognizerResult.from_json({"start": 0, "end": 5, "score": 0.5}),
      RecognizerResult("PERSON", 10, 15, 0.9)]
EntityRecognizer.remove_duplicates(rs)
# TypeError: '<' not supported between instances of 'NoneType' and 'str'
```

Suggested fix: key on `str(x.entity_type)`, or group results with a `dict` keyed by `entity_type` instead of sorting on it (which also removes the need for the `current_entity_type` sentinel below).

presidio-analyzer/presidio_analyzer/entity_recognizer.py
```diff
@@ -283,28 +283,52 @@ def remove_duplicates(results: List[RecognizerResult]) -> List[RecognizerResult]
...
+                ):
+                    to_keep = False
+                    break
 
             if to_keep:
                 filtered_results.append(result)
+                active_results.append(result)
 
-        return filtered_results
+        # Restore the existing score-first result order.
```
omri374, line 325: This re-sort keys on `(-score, start, -length)` only. Results that tie on all three but differ in `entity_type` previously came out in set-iteration order (effectively process-random because of string-hash seeding). Now they always come out in `entity_type`-alphabetical order, because the first sort above is stable and this one does not include the type.

That is a harmless determinism improvement, but `AGENTS.md` requires the PR description to state any change to returned entities, and the body currently claims "preserving existing behavior and result ordering". Please update the description so downstream consumers that iterate by position are warned.

presidio-analyzer/presidio_analyzer/entity_recognizer.py
```diff
@@ -283,28 +283,52 @@ def remove_duplicates(results: List[RecognizerResult]) -> List[RecognizerResult]
...
-                for filtered in filtered_results:
-                    # If result is contained in one of the other results
-                    if (
-                        result.contained_in(filtered)
-                        and result.entity_type == filtered.entity_type
-                    ):
-                        to_keep = False
-                        break
+            # Only overlapping results can contain the current result.
+            active_results = [
```
omri374: 🔴 **Worst case is still O(n²); the new test only covers the disjoint case.**

The pruning here only drops results that ended before the current one starts. Same-type results that nest with increasing score are all kept and all stay in `active_results`, so nothing is pruned and the inner loop runs exactly as before.

Reproduced on this head with n = 2000 results of one type, spans `[0, n - i]`, score `(i + 1) / n`:

| kept | `contained_in` calls | n(n-1)/2 |
|---|---|---|
| 2000 | 1,999,000 | 1,999,000 |

`test_when_number_of_results_is_large_then_remove_duplicates_finishes_in_reasonable_time` uses non-overlapping spans, so it would not catch this. Either state the real bound (linear in the number of overlapping kept results per type) in the docstring and PR, or add an overlapping-input test that backs the claim in the title.

---

presidio-analyzer/tests/test_entity_recognizer.py
omri374: `assert duration < 5` fails spuriously on a throttled or shared CI runner (or under coverage instrumentation), and a future O(n²) regression with a slightly cheaper inner loop could still pass. `.github/copilot-instructions.md` asks for exact expected values rather than ranges.

A deterministic alternative: wrap `RecognizerResult.contained_in` with a counting shim (e.g. via `monkeypatch`) and assert it is called 0 times for this non-overlapping input, or at most O(n) times. That also lets you drop `import time` on line 3.

presidio-analyzer/presidio_analyzer/entity_recognizer.py
```diff
@@ -283,28 +283,52 @@ def remove_duplicates(results: List[RecognizerResult]) -> List[RecognizerResult]
...
+                filtered for filtered in active_results if filtered.end >= result.start
+            ]
+
+            to_keep = True
+            for filtered in active_results:
+                # If result is contained in one of the other results
+                if (
+                    result.contained_in(filtered)
+                    and result.score <= filtered.score
+                    and result.entity_type == filtered.entity_type
```
omri374: 🟡 **This `entity_type` comparison is dead code.**

`active_results` is reset whenever the entity type changes (line 302), so every candidate in it already has the same type as `result`. Keeping the check suggests cross-type containment is possible here and misleads the next reader about the invariant.

The `current_entity_type` sentinel plus the reset branch is a hand-rolled `itertools.groupby(results, key=lambda r: r.entity_type)`; using it would remove the sentinel, the reset, and this comparison in one go.

---

presidio-analyzer/presidio_analyzer/entity_recognizer.py
```diff
@@ -283,28 +283,52 @@ def remove_duplicates(results: List[RecognizerResult]) -> List[RecognizerResult]
         :return: List[RecognizerResult]
         """
         results = list(set(results))
-        results = sorted(results, key=lambda x: (-x.score, x.start, -(x.end - x.start)))
+        results = sorted(
```
omri374, line 287: 🟡 **Docstring (lines 277-284, above) no longer matches the behavior.**

It still says duplicates are removed only for "identical start and ends and types", while the rewritten body explicitly drops any same-type result that is *contained in* a kept result with a score that is greater than or equal to it. Since the whole function was rewritten and a `score <=` condition was introduced, please update the reST docstring to describe the containment plus score rule (`AGENTS.md`: docs move with code).

---

omri374 (commented):
Thanks! Great addition. A few minor comments

presidio-analyzer/presidio_analyzer/entity_recognizer.py
mbukowicz: Good catch 👍.

I added a regression test for the missing `entity_type` case and updated the sort key to use `str(x.entity_type)`, preventing `None` or other non-string values from raising a `TypeError`.

presidio-analyzer/presidio_analyzer/entity_recognizer.py
mbukowicz: Good catch. I updated the PR description to explicitly call out the deterministic `entity_type` tie ordering and removed the claim that result ordering is entirely unchanged.

presidio-analyzer/tests/test_entity_recognizer.py
mbukowicz: Good point 👍 . I replaced the timing assertion with a deterministic test that fails if non-overlapping results trigger `RecognizerResult.contained_in` checks.

presidio-analyzer/presidio_analyzer/entity_recognizer.py
mbukowicz: Good suggestion 👍. I replaced the sentinel/reset logic with `itertools.groupby` and removed the redundant `entity_type` comparison.

presidio-analyzer/presidio_analyzer/entity_recognizer.py
mbukowicz: Updated the docstring to better reflect what the function does.

mbukowicz:
@omri374: All comments have been addressed.

Could you please take another look and see if everything looks good?

This is my first PR in this repo. Should I resolve comments once I’ve addressed them, or do you prefer to resolve them after verifying the changes?

omri374 (approved):

omri374:
> @omri374: All comments have been addressed.
> 
> Could you please take another look and see if everything looks good?
> 
> This is my first PR in this repo. Should I resolve comments once I’ve addressed them, or do you prefer to resolve them after verifying the changes?

Thanks @mbukowicz! I resolved all comments and approved the PR. Now waiting for the CI to successfully complete.

PR #2288, opened 2026-09-22
feat(anonymizer): opt-in deterministic encryption for the encrypt operator (#1033)

### LINKED ISSUE #1033: stable output for the encrypt operator
for analysis purposes on fields  that are  anonymized using the encrypt operator, it can be helpfull to have a stable output from the operator.   Right now when you run the anonymizer, you get a different result for each iteration. 

it would be grate if  a user can opt-out of the random iv in the cypher method. 

i tried to use hash but that can be computed back to the original value without the use of a secret.

v0ropaev:
### The gap

`presidio-anonymizer/presidio_anonymizer/operators/aes_cipher.py:24` draws a fresh random IV for every entity, unconditionally:

```python
iv = os.urandom(16)
```

and `operators/encrypt.py:12` declares `KEY = "key"` as the operator's only parameter. So the same value encrypts to something different every call:

```console
$ uv run python -c "from presidio_anonymizer.operators import AESCipher; k=b'1111111111111111'; print(AESCipher.encrypt(k,'John Doe')); print(AESCipher.encrypt(k,'John Doe'))"
hVVUWFk1ebYA2HH_wFxVj1FDJxz8DuY0o1Y9EpEdxsQ=
0reT2Nzv7e1nELqprSf5egFuV1g_Kj5sh6U3gixfaEA=
```

That rules out referential integrity across a dataset. `hash` already solved this for the non-reversible operator with `salt` (#1846, and `docs/anonymizer/index.md:243`), but the reporter explicitly rejected hashing because it is not reversible.

This follows the shape `AGENTS.md:104-108` prescribes for exactly this case:

> **Anonymizer operators**: non-reversible by default … Deterministic or format-preserving output is sometimes a hard requirement (e.g. referential integrity across a dataset); support it as an explicit, documented opt-in, never as the default behavior.

### What changed

A single optional boolean, `deterministic`, defaulting to `False`.

- `AESCipher.encrypt(key, text, deterministic=False)` — when the flag is set, the IV comes from `AESCipher._derive_iv` instead of `os.urandom`. The default branch is still `os.urandom(16)`, unchanged character for character.
- `AESCipher._derive_iv(key, encoded_text)` — a synthetic IV: `HMAC-SHA256(iv_key, plaintext)` truncated to 16 bytes, where `iv_key = HMAC-SHA256(key, "presidio-anonymizer/deterministic-iv")`. Two properties, both deliberate and both in the docstring. It is **keyed, not a bare digest of the plaintext**: an unkeyed digest is reproducible by anyone, so an attacker could confirm a guessed value straight from the IV without holding the key. And it is keyed with a **subkey rather than the encryption key itself**, so the key driving AES does not also produce a value that ships in the clear next to the ciphertext — the same separation RFC 5297 makes between SIV's two keys.
- `Encrypt` gains `DETERMINISTIC = "deterministic"`, forwards it from `params`, and type-checks it in `validate()` with the existing `validate_type` helper, so a non-boolean produces the repo's standard message: `Invalid parameter value for deterministic. Expecting 'boolean', but got 'string'.`

### The trade-off is explicit everywhere a reader looks

Deterministic encryption means identical plaintexts produce identical ciphertexts — that is the whole point, and it necessarily reveals which entities share a value and how often each value occurs. That is stated in the `encrypt` docstrings, in the `deterministic` row of the built-in operators table, in a new "Deterministic encryption for referential integrity" section in `docs/anonymizer/index.md` (with a `!!! warning "Deterministic encryption leaks equality"` admonition, placed beside the analogous hash-salt section), in the `Encrypt` schema in `docs/api-docs/api-docs.yml`, and in a new opt-in section at the end of `docs/tutorial/12_encryption.md`.

### Decrypt and the deanonymize path are byte-compatible — checked, not assumed

`decrypt.py` and `AESCipher.decrypt` are untouched (zero lines in the diff). The IV still travels prepended to the ciphertext inside the same urlsafe-base64 blob, so only the IV *source* changes and the wire format is frozen. Verified in both directions by loading `main`'s unmodified `aes_cipher.py` as a separate module alongside the patched one:

```
old fixture -> new decrypt: 'Chloë'     # main's own test fixture S184CMt9Drj7...JME0=
new default -> OLD decrypt: 'Chloë'     # new random-IV output, decrypted by unmodified main
new determin -> OLD decrypt: 'Chloë'    # new DETERMINISTIC output, decrypted by unmodified main
len(default)= 32  len(determin)= 32     # identical blob length
```

An existing deployment running the old `decrypt` restores deterministically-encrypted values with no change at all.

### Verification

Run in `presidio-anonymizer/` with uv, Python 3.14, ruff 0.15.21.

| | baseline (main @ 33586b8d) | with this change |
|---|---|---|
| `uv run pytest -q` | 324 passed, 7 skipped | **341 passed, 7 skipped** |
| `uv run ruff check .` | All checks passed! | **All checks passed!** |

- `uv run ruff check` from the repo root (what `ci.yml:39` runs) → All checks passed!
- `uv run ruff format --check` on both changed source files → 2 files already formatted
- `uv run diff-cover coverage.xml --compare-branch=origin/main --src-roots .` → **15 changed lines, 0 missing, 100%** (`aes_cipher.py` 100%, `encrypt.py` 100% on changed lines)
- Both documented examples were executed: the index.md example prints the same ciphertext for both `John Doe` occurrences (`same ciphertext: True`) while the same call without the flag prints `False`; the tutorial example round-trips back to `"My name is James Bond, and James Bond is my name"` through `DeanonymizeEngine`.
- `docs/api-docs/api-docs.yml` parses under `yaml.safe_load`; `Encrypt.properties` = `['type', 'key', 'deterministic']`.

Mutation-checked the new tests — 7 deliberate mutations, all caught, with the full suite green before and after:

| Mutation | Result | Caught by |
|---|---|---|
| Ignore the flag, always `os.urandom(16)` | 10 failed | the 4 determinism tests |
| Always derive (default becomes deterministic) | 1 failed | `..._no_deterministic_then_same_key_and_text_return_different_texts` |
| Unkeyed SHA-256 of the plaintext instead of HMAC | 1 failed | `..._iv_is_keyed_and_not_a_bare_hash_of_the_text` |
| HMAC key and message swapped | 1 failed | same test |
| Encryption key used directly instead of the derived subkey | 1 failed | same test |
| Truncate the derived IV to 12 bytes | 12 failed | all deterministic tests |
| Drop the `deterministic=` pass-through in `operate` | 4 failed | both mock tests + operator + integration |
| Drop the `validate_type` line | 1 failed | `..._non_boolean_deterministic_then_ipe_raised` |

The keyed-construction test decodes the IV back out of the ciphertext and compares it against stdlib `hmac.new(key, text, sha256).digest()[:16]`, asserting it is **not** `sha256(text).digest()[:16]` — an independent implementation, so it pins the design rather than restating it.

### What I left out

- The `{"key", "suite", "options"}` API sketched in the issue thread — a cipher-suite plugin surface, not this fix, and it breaks the existing `{"key": ...}` contract.
- AES-SIV / AEAD. `cryptography`'s `AESSIV` wants 256/384/512-bit keys, which collides with the 128/192/256-bit contract enforced in `Encrypt.validate` and would change the wire format, invalidating every existing ciphertext. This is SIV-*style* (a deterministic IV for CBC), the smallest change that satisfies the issue with the format frozen.
- AES-ECB (raised in the first comment): it leaks per-block structure, strictly worse than the whole-value equality leak here, and the reporter answered that himself in the thread with a CVE reference.
- `CHANGELOG.md`, per CONTRIBUTING.md and AGENTS.md.
- Repo-wide `ruff format` — it already fails on 7 untouched files on `main`, and a bulk reformat would bury a crypto diff. One incidental note: `main` has no blank line between `AESCipher.decrypt` and `is_valid_key_size`; inserting `_derive_iv` between them necessarily adds that separator, so `aes_cipher.py` now happens to pass `ruff format --check`.

## Issue reference

Fixes #1033

## Checklist


v0ropaev:
Both points were worth acting on. Addressed in `a18d4f1f`.

**Falsy non-boolean values.** Confirmed: `validate_type` guards on `if parameter_value and not isinstance(...)` (`services/validators.py:66`), so `"deterministic": 0`, `""` and `[]` skipped the check and were then read as `False` — a caller asking for the wrong type got the opposite of an error, silently. That is right for the optional string parameters the helper was written for, and wrong for a boolean.

Rather than special-case it inside `Encrypt`, I added `validate_type_strict` next to it: same message builder, `None` still means "not supplied", but a value that *is* supplied has to be of the type. It is the check any future boolean parameter will want. `validate_type` itself is untouched, so `replace`, `pii_entity` and everything else behave exactly as before.

Regression test is parametrized over `0`, `""`, `[]`, `"true"` and `1`. Reverting to `validate_type` fails the first three and leaves the last two passing, which is the split the old code had.

**CBC integrity.** Fair, though it is a property of the operator rather than of this option: `encrypt` has always been AES-CBC with no MAC, so ciphertexts are confidential but not authenticated, in both modes. I did not want to imply the option introduces it, so the warning block now says it plainly and says why it matters here specifically — deterministic ciphertexts are the ones likely to end up as dataset keys that other systems join on, and an attacker who can modify stored values can alter them undetectably either way.

`pytest`: 346 passed, 7 skipped (was 341). `ruff check`: clean.

v0ropaev:
Took the README point too, in `2eacb64a`. The package README lists each operator's parameters and had only `key` under encrypt, so `deterministic` was missing from the one place a reader of that package looks first — and it is the parameter that carries a trade-off worth stating before someone reaches for it. Wording matches the operators table and the guide section rather than inventing a third phrasing.

That leaves the four surfaces consistent: the package README, `docs/anonymizer/index.md` (table row plus the warning block), the `Encrypt` schema in `api-docs.yml`, and `docs/tutorial/12_encryption.md`.

